PR #1, opened 2020-05-11
Infinite Scroll example with table of agents

inouire:
Here is where I am so far, but the `revealed` trigger does not seem to work properly, it just works for the first element. When I fallback on a classical click trigger, it works

Another thing, I found an inconsistency on `afterend`  (in the code) and `afterEnd` (in the doc)
Which one is the good one?

Thanks


chg20:
Docs are wrong, should be afterend :/

Are you comfortable debugging the javascript?  I can look at it it not.

inouire:
Not really confortable to debug it :/ Thanks

PR #3, opened 2020-05-14
Example delete row

inouire:
I added a bit of code to support the DELETE requests, but it seem to fail working with the swap option:
the kutty-swapping class is added, but the element is never swapped with the content "" (empty, to delete the row)

Is it because it is empty?


inouire:
Ready for review

PR #7, opened 2020-05-15
Fix kt-params Link

breno-dsuite:


breno-dsuite:
Great fan of Intercooler, this is very neat!

chg20:
Thank you!

PR #12, opened 2020-05-16
Avoid shadowing the global prompt function

rschroll:
The prompt variable inside issueAjaxRequest shadowed the prompt function
that it was attempting to use, thereby breaking the use of kt-prompt.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -4,10 +4,6 @@
 * Support IE11 (stretch: IE10)
 * < 6k in .min.gz form
 
-## Launch TODOS
-
-* Move to development branch
-
 ## Post-Launch TODOS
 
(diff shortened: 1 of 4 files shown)


chg20:
Hard one to test, thanks for catching that!

Can you retarget the 'dev' branch?

rschroll:
Retargeted.

Yeah, I only caught this through trying things interactively.  I'm thinking adding it as an example would at least help manual testing.  Would you like that in this PR or in another one?

chg20:
An example is a good idea. Maybe an example with both the query and prompt in it?

Cheers,
Carson

chg20:
I ended up adding some manual tests for this.

Cheers,
Carson

PR #18, opened 2020-05-17
Allow request to have empty header values

rschroll:
This is a rather odd one, and a silly rabbit hole to get drawn into, but I noticed in #16 that submitting an empty string resulted in the "server" echoing `undefined`.  It looks like there were several assumptions that the header values, and specifically the response to the prompt, must be truthy.  Empty string is a valid response, so that's not the case.

This may be such a corner case that it's not worth fixing.  But it appears to be fixable without getting too involved.


chg20:
Seems reasonable to me, thanks!

PR #42, opened 2020-05-24
Switch afterRequest to standard event from error event if no error

rschroll:
I don't think this was causing any problems, but it could confuse anyone watching the event.


chg20:
Thanks, great catch!

Issue #23, opened 2020-05-18
Cannot use current page url for hx-get


mqudsi:
I would assume that `<input type="checkbox" hx-get name="foo" />` would perform a GET against the current page (like a form without an action),  and if not, then `<input type="checkbox" hx-get="." name="foo" />` would.

In reality, `hx-get` without a value does nothing and `hx-get="."` makes a request to `/` rather than the current page. The only way I could get the above to toggle between `/Registrations?foo=on` and `/Registrations` was to hard-code `hx-get="/Registrations"`, which is unfortunate.

chg20:
Thank you for mentioning this!  It turns out to my surprised that `<a href>foo</a>` is legal and does refer to the current page, so htmx should follow your suggested logic.

Are you interested in trying to implement it?

chg20:
fixed w/ https://github.com/bigskysoftware/htmx/commit/59444e8f44fd45f272f6180b3731b3ce02cb86fc

mqudsi:
Thank you very much. This really helps when composing behavior based on many separate htmx entities, each of which sets an individual URL parameter.

bencroker:
Am I correct in thinking that the following _should_ trigger a GET request to the current page when the value is changed? I'm not seeing that behaviour in 0.4.0.

```html
<input hx-get type="text">
```

chg20:
I don't like that I didn't submit a test w/ https://github.com/bigskysoftware/htmx/commit/59444e8f44fd45f272f6180b3731b3ce02cb86fc, so it might still be an open issue.  @bencroker what behavior are you seeing?

I need to think about how to test this.

bencroker:
I'm seeing no reaction. It looks like the issue might be that the `path` is empty here:
https://github.com/bigskysoftware/htmx/blob/f8dfe727b33d498e89f4d842e5c9341629a94897/src/htmx.js#L825-L829

chg20:
yep, that's it.  need to check if the attribute exists rather than test its value.

you want to take a crack at it?

bencroker:
The PR above fixes the issue, thanks for the nudge!

chg20:
OK, the tests are a little janky but they work and verify the behavior to a reasonable extent:

https://github.com/bigskysoftware/htmx/commit/d144742048e00a22724779365686d3bafab8c2a4

this will go out w/ 0.0.5

bencroker:
Thanks, I'll be sure to write a test next time, now that I see how you did it.

PR #43, opened 2020-05-26
Allows for empty verb attribute path

bencroker:
No test written, but should fix https://github.com/bigskysoftware/htmx/issues/23.


chg20:
Looks perfect.  How do we test it?

Actually having it hit a URL is a little rough because we would have to set it up such that we could make an ajax request to the "current" page...

Maybe pull out the htmx internal data (`htmx._("getInternalData")(elt)`) and verify that it has a path and a verb field on it?

bencroker:
I'm not quite yet familiar enough with the codebase to start pulling out the internals ;)

PR #46, opened 2020-05-28
Rename custom HTTP headers to no longer start with `X-`

acnebs:
This convention is deprecated, as per [this RFC](https://tools.ietf.org/html/rfc6648).

tl;dr – prepending custom headers with `X-` doesn't make them any clearer, and can present problems if your custom proprietary headers ever get standardized. [More from MDN.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)


chg20:
thanks!

PR #51, opened 2020-05-28
Allow rendering element template strings with Nunjucks

GavinRay97:
This change adds support for defining element blocks of text (like `<script>` tags similar to Mustache for example) with Nunjucks:

```html
<input
  type="text"
  name="query"
  hx-ext="json-enc, client-side-templates"
  hx-post="https://localhost:8080/v1/graphql"
  hx-trigger="keyup changed delay:500ms"
  nunjucks-template="employee-list"
  placeholder="query AllEmployees { employees { email } }"
  hx-swap="afterend"
/>

<script id="employee-list" type="nunjucks">
  {% for employee in data.employees %}
    <li>{{employee.email}}</li>
  {% endfor %}
</script>
```

The extension will try to use `nunjucks.render()` with the default/initial behavior (I'm actually not even certain how this works, it loads it from an URL? Never used a template rendering engine before today) first, if that fails and errors due to not being available, it will look for an element block matching the `nunjucks-template` ID and use it's `innerHTML` as the template string to `nunjucks.renderString()`, mirroring the Mustache functionality. This gives a nice flow where you can define "components" below where you intend to use them with Nunjucks (since Mustache is logicless it's not very featured for this) as `<script>` tags and then use `hx-swap="afterend"` or whatnot.

Also, minor change but moved the data parse function to top scope to not re-declare in each template handler block :+1:

nunjucks-template

### DIFF
--- a/src/ext/client-side-templates.js
+++ b/src/ext/client-side-templates.js
@@ -1,9 +1,9 @@
 htmx.defineExtension('client-side-templates', {
     transformResponse : function(text, xhr, elt) {
+        var data = JSON.parse(text);
 
         var mustacheTemplate = htmx.closest(elt, "[mustache-template]");
         if (mustacheTemplate) {
-            var data = JSON.parse(text);
             var templateId = mustacheTemplate.getAttribute('mustache-template');
             var template = htmx.find("#" + templateId);
             if (template) {
@@ -15,16 +15,19 @@ htmx.defineExtension('client-side-templates', {
 
         var handlebarsTemplate = htmx.closest(elt, "[handlebars-template]");
         if (handlebarsTemplate) {
-            var data = JSON.parse(text);
             var templateName = handlebarsTemplate.getAttribute('handlebars-template');
             return Handlebars.partials[templateName](data);
         }
 
         var nunjucksTemplate = htmx.closest(elt, "[nunjucks-template]");
         if (nunjucksTemplate) {
-            var data = JSON.parse(text);
             var templateName = nunjucksTemplate.getAttribute('nunjucks-template');
-            return nunjucks.render(templateName, data);
+            var template = htmx.find('#' + templateName);
+            try {
+              return nunjucks.render(templateName, data);
+            } catch (e) {
+              return nunjucks.renderString(template.innerHTML, data);
+            }
         }
 
         return text;


src/ext/client-side-templates.js
```diff
@@ -1,9 +1,9 @@
 htmx.defineExtension('client-side-templates', {
     transformResponse : function(text, xhr, elt) {
+        var data = JSON.parse(text);
```
chg20, line 3: I don't want to parse the JSON unless we know for sure there is a client side template associated with this request.  The hx-ext="client-side-templates" attribute can live on a parent tag (e.g. the body) so we could end up trying to parse non-JSON responses as JSON, throwing an exception

src/ext/client-side-templates.js
```diff
@@ -15,18 +15,21 @@ htmx.defineExtension('client-side-templates', {
...
             var templateName = handlebarsTemplate.getAttribute('handlebars-template');
             return Handlebars.partials[templateName](data);
         }
 
         var nunjucksTemplate = htmx.closest(elt, "[nunjucks-template]");
         if (nunjucksTemplate) {
-            var data = JSON.parse(text);
             var templateName = nunjucksTemplate.getAttribute('nunjucks-template');
-            return nunjucks.render(templateName, data);
+            var template = htmx.find('#' + templateName);
```
chg20, line 25: what do you think about having people use nunjucks-template="#foo" for the inline case? so we don't have to try/catch?

src/ext/client-side-templates.js
GavinRay97: AH that's fair, sounds good

chg20:
@GavinRay97 can you make the suggested changes and issue a new pull request?

GavinRay97:
> @GavinRay97 can you make the suggested changes and issue a new pull request?

Hey sorry, was caught up with life stuff. Yeah I can do that :+1:

src/ext/client-side-templates.js
GavinRay97: The reason the try/catch is there is because nunjucks throws if the template doesn't exist with `nunjucks.render()`, so in order to preserve the ability to render with both `.render()` and then fallback to `.renderString()` inline, I could only think of rescuing the initial `.render()` exception and trying `.renderString()` in the catch block. 

If that throws as well, then it doesn't matter because there's no more code to be executed in the block so it has no effect (besides not gracefully handling the error).

Am open to any suggestions here, `try/catch` is not clean but I looked through the nunjucks API for `.getTemplate()` and it seems that function only works if there's been some sort of filesystem environment mounted (but that could be my misunderstanding).

PR #53, opened 2020-05-28
Fix docs.md `trigger` attribute in polling section

webignition:
The polling section has example containing a `trigger` attribute. Updated this attribute to be `hx-trigger`.


chg20:
thank you!

webignition:
You're most welcome!

PR #65, opened 2020-05-29
Ensure submit type button values get captured

JoshAshby:
This PR adds support for including values on submit buttons.

`<button>` tags with a type of `submit` can have associated `name` and `value` attributes that'll be included in form submissions if that button was [clicked](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), however, line [1108](https://github.com/bigskysoftware/htmx/blob/c39e902cfd6ce078999fefae5019a7a23308bf82/src/htmx.js#L1108) seems to prevent htmx from respecting this, as it skips all buttons when serializing a forms inputs.

Here's a little demo page against httpbin. When the `Test default` button is clicked, httpbin reports that the forms data is

```
{
  "test": "some value"
}
```

However, when the htmx handled `Test htmx` button is clicked, httpbin reports empty form data.

```html
<!DOCTYPE html>
<head>
  <script src="https://unpkg.com/htmx.org@0.0.4"></script>
</head>

<p>Logs {}</p>
<form hx-post="https://httpbin.org/post">
  <button name="test" value="some value">Test htmx</button>
</form>

<p>Logs { "test": "some value" }</p>
<form method="post" action="https://httpbin.org/post">
  <button name="test" value="some value">Test default</button>
</form>
```

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1104,6 +1104,10 @@ return (function () {
             if(elt.name === "" || elt.name == null || elt.disabled) {
                 return false;
             }
+            // include buttons with values
+            if (elt.type === "submit" && elt.name) {
+              return true;
+            }
             // ignore "submitter" types (see jQuery src/serialize.js)
             if (elt.type === "button" || elt.type === "submit" || elt.tagName === "image" || elt.tagName === "reset" || elt.tagName === "file" ) {
                 return false;
--- a/test/core/regressions.js
+++ b/test/core/regressions.js
@@ -49,4 +49,18 @@ describe("Core htmx Regression Tests", function(){
         form.innerHTML.should.equal("variable=")
     });
 
+    it ('Handles button values properly', function() {
+        this.server.respondWith("POST", "/htmx.php", function (xhr) {
+            xhr.respond(200, {}, xhr.requestBody);
+        });
+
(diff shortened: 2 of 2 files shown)


chg20:
I think that button values are only included if they are clicked, right?

If I'm remembering right the hack around this is to send the "active" button value up with the request.  I didn't want to do that because it seemed too magical and there was some chatter that it didn't work 100% of the time, so instead I send up headers with the active information:

https://htmx.org/reference#headers

Maybe there's a better way to handle things.  In intercooler I tried to keep track of the last button clicked, but it was a mess.

JoshAshby:
Heh, I spaced on that while quickly hacking this and it worked out as the forms I was trying to get this working on only had one button with a value set (and even made that  mistake in the tests too); You're correct that it is only sent if that specific button was clicked.

While I can look into using the header, ideally I'd rather htmx emulate the known behavior of the forms and not have to special case the backend to handle htmx originating ones. If this ID is already tracked and is available at the time `shouldInclude` is called, however, it seems like it could be added into the check for if the trigger element is the button?

I haven't dug much into htmx though so I haven't come across where/when that header being set yet though so I guess I should go spelunking :)

chg20:
OK, I'm going to close this out pending a new pull request for this feature.

PR #78, opened 2020-06-01
cscortes

cscortes:
Here are my changes for the encodeParameters routine and issueAjax.  If you know of a better way, I would be happy to hear it.  Please let me know what you find.  

**The Problem**, setRequestHeader does not allow you to replace the encoding variable, it only allows you to append to it.  So I decided to run the "encodeParameter" function first, check to see if the encoding was set, and if not then set it to url.  
Since I only use the xhr variable to "set the header variable", I decided to send **headers** instead.  

_Bye the way, this is how I use sse-body extension:_

```
   <body id="thebody" hx-sse="connect /news_updates">

        <h1>Receiving Messages:</h1>
        <div id="sselist" hx-trigger="sse:new_news" hx-ext="sse-body" 
        hx-post="/news" hx-target="#nid" hx-swap="beforeend">
            <ul id="nid">
            </ul>
        </div>

    </body>
```
I was going to add it to my growing list of Flask/htmx examples.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "htmx.org",
-  "version": "0.0.1",
+  "version": "0.0.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


cscortes:
I guess it does.  Maybe I was trying to overthink it.  Never really saw the difference between the OnEvent Config and the encode parameters.  But if this is the approach you want to take, I'm good with it.  Also, I did submit a more generalized approach to the ajaxrequest routine, that would give you access to everything you needed in 1 call vs. 2. 

The pseudo code went beyond the json-enc extension to hopefully get the extension writer thinking about the normal case, and his/her deviations:

We did everything you need for the normal case, but if you really want to change things then here is your 1 chance to get it done. 

Alright --- got that out of my head space.  Back to this solution, if the solution is to use the details on the config event, probably have to add it to the documentation for extensions, if it is not already there otherwise the next extension writer will be frustrated.

Maybe we will need a "extensions tutorial" ??

cscortes:
added a pull request with your recommendation, I will tract this pull.

PR #82, opened 2020-06-05
Add an extension to include the commonly-used X-Requested-With header

36864:
This extension adds the X-Requested-With header, which is used by many javascript frameworks to identify ajax requests.

Specifically for my use case, laravel has helper functions to determine if a request was made through ajax that look at this header.


36864:
Done.

PR #85, opened 2020-06-08
Avoid throwing errors when parsing malformed JSON

rschroll:
Instead, treat it as if no value was present.

I ran into a problem where I had malformed JSON in the htmx-config meta tag.  HTMX stopped working, and the error message was laconic: It only identified one character from the offending string.  JSON is parsed in a couple of other places.  Some seem unlikely to get malformed JSON (the history cache, for example), but I thought safety would be best.

This solution powers through errors, printing out a warning to the console.  I like this approach since it keeps things working as much as possible.  But if we want to be more obvious to the developer that something has broken, we could:
1. Use `console.error` instead of `console.warn`.  I chose `warn` since execution is continuing, but maybe red text is better for catching people's eyes.
2. Throw the error after printing out a more verbose error.  This would make it more obvious to developers that something is broken, but it would also make it more likely that something breaks for users.


rschroll:
My two worries about handling this with an event:
1. It's not clear you want the same handler for all the places JSON parsing could fail.  I'd assume that a single handler would be the exception, not the rule.  I do think it would be reasonable to trigger different error events for these (and other) parsing errors.
2. As a developer, I don't want to have to set up logging just to discover an error.  I want it to jump out immediately.  (Taken to its logical conclusion, this would argue for re-raising the error once some context has been provided.)  I don't mind having an event triggered, nor do I mind an event handler being able to silence the error.  But I don't think event handling should be the only way to find out about errors.

The dev extension is an interesting idea.  If you want to open an issue and tag me on it, I'm happy to mull over ideas.

chg20:
agree on the first point: different errors for different situations seems like the right thing

on the second point, if you call

```javascript
triggerErrorEvent(elt, 'jsonError.htmx', e);
```

it will eventually call through to 

```javascript
  logError(detail.error);
```
which looks like this:

```javascript
        function logError(msg) {
            if(console.error) {
                console.error(msg);
            } else if (console.log) {
                console.log("ERROR: ", msg);
            }
        }
```
So error events should be logged to the console even if logging isn't set up.

chg20:
@rschroll any thoughts on my comments?  I like this change a lot and want to integrate it.

PR #90, opened 2020-06-11
fix hx-sse trigger example

ssendev:
The `sse:` prefix was missing


ssendev:
sorry i'll try again against master

PR #91, opened 2020-06-11
fix hx-sse trigger in example

ssendev:
The sse: prefix was missing


chg20:
thank you, and thank you for catching the throttle issue to

turns out the test for trigger spec parsing wasn't running properly due to javascripts insane loop capture semantics!

PR #92, opened 2020-06-11
Fixup json enc

cscortes:
Included recommend json-enc fix, passed all tests and I tested it against FLASK test, also worked.  I will close my last pull request.


chg20:
Looks great!  Thank you!

PR #93, opened 2020-06-11
Use Element.prototype.closest if available

twidi:
It is used in the `closest` function, which is now also used, via
another new function `getClosestEltWithAttribute` that uses `closest`,
for:
- `getClosestAttributeValue`
- `getTarget`

**Tests**: no new tests are needed, and all existing tests pass
**Doc**: no doc is needed

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -60,12 +60,13 @@ return (function () {
             }
         }
 
+        function getClosestEltWithAttribute(elt, attributeName) {
+            return closest(elt, "[" + attributeName + "],[data-" + attributeName + "]");
+        }
+
         function getClosestAttributeValue(elt, attributeName) {
-            var closestAttr = null;
-            getClosestMatch(elt, function (e) {
-                return closestAttr = getAttributeValue(e, attributeName);
-            });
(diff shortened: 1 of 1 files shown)


twidi:
**Note to the author of this project**:

 It is my first PR on this project, but you can expect many to come. 

I love the idea and started to play with it in a POC I'm currently doing and I saw a few things where I think I may be able to help.

Thanks for `htmx`!

chg20:
thank you @twidi I really appreciate the help!  :)

Is the main benefit of this change performance?

I need to review your other change (which looks great!) but I might not have time until next week.  I'm shooting for a 0.1 release next friday and want to make sure that your change gets in.

thank you again

twidi:
Yes it's for performance. I didn't run any benchmark but I assume that native closest is faster that looping over parents in javascript

chg20:
@twidi given that the hierarchy is going to have an O(log(n)) depth, I'm not too concerned about the perf here, so I'm going to reject this change for now, pending a compelling perf reason to special case it.  I appreciate the change though.

I hope to review (and accept!) you other excellent changes later this week.

PR #94, opened 2020-06-11
Handle events and content after `handleSwap` in extensions

twidi:
When swap was handled by an extension via its `handleSwap` method, a few expected things didn't work:
- `afterSwap.htmx` and `afterSettle.htmx` events where not propagated to `onLoad` methods of extensions
- `load.htmx` was not called for newly added elements
- new content in the target was not handled by `htmx`, so for example if an element with ` hx-trigger=load` was included, nothing was done about it.

This is fixed in this PR, the work being split in two commits that touch two independent things.

See the description of these commits to know for each what the problem was, what was expected and how it is handled.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -403,7 +403,7 @@ return (function () {
             while(fragment.childNodes.length > 0){
                 var child = fragment.firstChild;
                 parentNode.insertBefore(child, insertBefore);
-                if (child.nodeType !== Node.TEXT_NODE) {
+                if (child.nodeType !== Node.TEXT_NODE && child.nodeType !== Node.COMMENT_NODE) {
                     settleInfo.tasks.push(makeAjaxLoadTask(child));
                 }
             }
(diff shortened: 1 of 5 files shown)


src/htmx.js
```diff
@@ -403,7 +403,7 @@ return (function () {
             while(fragment.childNodes.length > 0){
                 var child = fragment.firstChild;
                 parentNode.insertBefore(child, insertBefore);
-                if (child.nodeType !== Node.TEXT_NODE) {
+                if (child.nodeType !== Node.TEXT_NODE && child.nodeType !== Node.COMMENT_NODE) {
```
chg20, line 406: great catch!

twidi: I guess it's not a problem because later calls to queryselector won't find anything but with this we can avoid some function calls :)

twidi:
If #98 is adopted, there will be some code to add the the second commit of this PR (or to #98 if this one is adopted first) to handle the new behavior of `afterSwap` and `afterSettle` when content is swapped by `handleSwap` (ie settings `settleInfo.elts` to the elements returned by `handleSwap`), something like:

```diff
 if (newElements) {
     if (typeof newElements.length !== 'undefined') {
         // if handleSwap returns an array (like) of elements, we handle them
+        settleInfo.elts = [];
         for (var j = 0; j < newElements.length; j++) {
             var child = newElements[j];
             if (child.nodeType !== Node.TEXT_NODE && child.nodeType !== Node.COMMENT_NODE) {
                 settleInfo.tasks.push(makeAjaxLoadTask(child));
+                settleInfo.elts.push(child);
             }
         }
     }
```

chg20:
@twidi I reallly appreciate all this work you are putting in.  My mental energy is being taken up on another project I'm struggling with but I hope to be able to review these changes soon and comment on them.

src/htmx.js
chg20: i wonder if we should push the nodeType check inside of makeAjaxLoadTask() so we don't have to repeat ourselves...

chg20:
OK, I merged this change, but manually backed out the first change made because I don't quite understand it.  It seems like if the target is nuked via an outerHTML swap, not calling settle, etc. on a (formerly) parent element is reasonable to me.

twidi:
@chg20 `if the target is nuked via an outerHTML swap` => it's not about the removed target in a `outerHTML` swap, but for the newly added element in a swap handled via an extension.

This is explained in the message of this first commit and the included test show the case.

twidi:
In fact the test added in the first commit is wrong, because afterSwap is still called once... but for the target, and not for the children.

I think it may be solved by the pr #98

PR #95, opened 2020-06-13
capture sse data and treat as rawparameters to ajax call

cscortes:
This adds the feature of passing on the sse event data to the ajax call as raw parameters.  All tests are still passing.


1cg:
Unfortunately SSE implementation has changed dramatically.  This will need to be looked at again.  Sorry about that! :/

PR #96, opened 2020-06-13
Changes to use actual requested verb, instead of POST

cscortes:
Created a extension to use the actual verb instead of POST with override.  Saw a feature request for it.  Just wanted to make it available is someone needs it.  I know it would have eliminated a hack I used for flask to had the POST + overrride.  

example:

`        <div hx-include="#checked-contacts" hx-target="#tbody">
            <a class="btn" hx-put="/activate" hx-ext="actual-verb">Activate</a>
            <a class="btn" hx-put="/deactivate" hx-ext="actual-verb">Deactivate</a>
        </div>
` 

* Might be nice if I could apply to all hx-put, patch, delete, etc calls globally.

### DIFF
--- a/src/ext/actual-verb.js
+++ b/src/ext/actual-verb.js
@@ -1,0 +1,7 @@
+htmx.defineExtension('actual-verb', {
+    onEvent: function (name, evt) {
+        if (name === "configRequest.htmx") {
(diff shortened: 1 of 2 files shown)


chg20:
:) already implemented in `dev`

Issue #97, opened 2020-06-13
Inconsistency in event elements


twidi:
In events we have, in `detail`, `elt` and `target`, and `elt` is forced to the element passed when calling `triggerEvent`, so `elt` is for some events the element triggering the htmx action (the one with the `hx-` attributes) and sometimes its the target. 

I think `elt` should always be the element triggering the htmx action.

We have the "event target" in the `target` attribute that is the element passed when calling `triggerEvent` (so the triggering element or the target depending on the event)

What we should have is, imho:

- the original element triggering the htmx action
- the original target before (potential) replacement
- the final target (that can be the same as the original target if the `hx-swap` is `innerHTML`)

With this we may be able to do more powerful things with extensions.

Note: this wasn't only the case in the dev branch since you changed the way events are triggered by iterating on `settleInfo.elts`

twidi:
A possibility to always have the triggering element in the event detail is to add `trigger: elt` when defining `eventDetail`.

If you like the idea I can make a PR for that.

twidi:
About final target, in fact `after` events are triggered for each new "children", so it's good (that is `detail.elt` in these events). So the only thing missing is the `trigger` option as I proposed above.

twidi:
I did a proposal in #98

bencroker:
Closing due to inactivity, see https://github.com/bigskysoftware/htmx/pull/98#issuecomment-685223658.

PR #98, opened 2020-06-14
Add `trigger` in event `detail` 

twidi:
(this is a proposal for #97)

`detail.elt` can sometimes be the element that triggered the request, or
sometimes the target, so we cannot rely on it easily.

And as we always have the `target` key correctly filled, this commit
also adds the `trigger` key.

This also updates the `afterSwap.htmx` and `afterSettle.htmx` events to
be triggered for all newly added elements when `hx-swap` is
`beforebegin` and `beforeend`, for consistency with other modes.

**Tests**: many are added in `test/core/events.js` to check many cases.
**Doc**: `www/events.md` is updated accordingly.

**Note**: this PR also add 2 commits to fix some code/docs about existing events and one to enhance the `load.htmx` test

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -417,13 +417,16 @@ return (function () {
 
         function insertNodesBefore(parentNode, insertBefore, fragment, settleInfo) {
             handleAttributes(parentNode, fragment, settleInfo);
+            var newElts = [];
             while(fragment.childNodes.length > 0){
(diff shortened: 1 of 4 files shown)


1cg:
Hi @twidi very sorry but this patch has gotten stale.
If you are still interested in pursuing it, can you jump on the htmx discord and we can discuss?

PR #113, opened 2020-06-23
Local vars extension -- working code

cscortes:
Still need to add some test cases.  This code was based off of include-values extension and previous IC-local-vars code.  On configuration, values set as parameters.  If they need to get "returned" then they are sent back on a Header HX-Set-Local-Vars.  I was struggling between having them both on send and return use the header -- as to make it obvious what is going on.  Comments?  I will also generate some test cases after some feedback.

This is how I envision it work -- from my own Flask tests (not included). 

```html
    <body>
        <p>
            <a class="anchor" hx-get="/countertime" hx-ext="local-vars" include-local-vars="mycounter myservertime" hx-target="#updatedata"> Click to Count and Time </a>
        </p>
    
        <p>
            <a hx-get="/counter" hx-ext="local-vars" include-local-vars="mycounter" hx-target="#updatedata"> Click to Count </a>
        </p>

        <p>
            <a hx-get="/systime" hx-ext="local-vars" include-local-vars="myservertime" hx-target="#updatedata"> Click to Time </a>
        </p>

        <div id="updatedata" style="border: blue; border-width: 2px; border-style: solid; width: 300px; padding-left: 5px;">
            <!-- counter displays here  -->
            <p>No Updates</p>
        </div>
    </body>```


cscortes:
After trying to write a simple Flask example with this code, It really felt like I was hacking stuff on the server side.   I will withdraw this pull request and change things to be symetric, HX-Local-Vars header on both the request and response.  The reason is that there are also session variables (which I can just copy paste, but I can see where they could start to conflict).  So we could have another HX-Session-Vars and maybe even HX-Cookie-Vars without being worried that there will be a name clash.

PR #117, opened 2020-06-25
Local vars extension

cscortes:
Added a few features to my new extension:

1.  Prepending HX- to my include-local-vars.  It makes it consistent with htmx naming convention and it is easier to point out the difference when you are making a video about htmx. 
2.  Added warning, because you might not have localstorage enabled or might mistype the hx-include-local-storage.
3.  accept only hx-ext="local-vars" and hx-include-local-storage="x y z" on same html tag.
4.  Symetric HX-Local-Vars on response and request headers.  To make the backend programming obvious.


src/ext/local-vars.js
```diff
@@ -0,0 +1,36 @@
+htmx.defineExtension('local-vars', {
+    onEvent: function (name, evt) {
+        if (!localStorage) {
+            console.log("Warning: browser does not support local storage.")
+        }
+        elt = evt.detail.elt;
+        if (name === "configRequest.htmx") {
```
chg20, line 7: can you update to the new event naming standard: "htmx:configRequest"

src/ext/local-vars.js
```diff
@@ -0,0 +1,36 @@
...
+            if (elt.hasAttribute("hx-include-local-vars") || elt.hasAttribute("data-hx-include-local-vars") ) {
```
chg20, line 8: I would prefer a non-hx prefix, something like `local-vars` and `data-local-vars` so these aren't confused with a standard attribute

src/ext/local-vars.js
```diff
@@ -0,0 +1,36 @@
...
+                evt.detail.headers["HX-Local-Vars"] = JSON.stringify(data);           
+            }
+            else {
+                console.log("Warning: hx-local-vars without hx-include-local-vars.")
+            }
+        }
+    },
+    transformResponse: function(text, xhr, elt) {
+        if (elt.hasAttribute("hx-include-local-vars") || elt.hasAttribute("data-hx-include-local-vars")) {
+            var localVars = xhr.getResponseHeader("HX-Local-Vars");
```
chg20, line 26: similarly, would prefer to drop the "HX-" prefix

cscortes:
Okay, using new style events.  Remove hx- prefix.

src/ext/local-vars.js
cscortes: ok

src/ext/local-vars.js
cscortes: ok

src/ext/local-vars.js
cscortes: ok

cscortes:
Added to the documentation.  Did run the tests, but not after the event name changes since I am not sure how to change from master to dev in my sandbox.

cscortes:
Also, added a new feature to my extension, with addition of a command, sessionStorage:1 you can use session vars instead of local vars.  include-local-vars="sessionStorage:1 myvar1 myvar2".

src/ext/local-vars.js
cscortes: Are you okay with include-local-vars and  data-include-local-vars?  If so, I think we are done :-).

cscortes:
Fixed the typos and added a little clarification.  Thanks for catching those grammer errors for me!

1cg:
Sorry, I let this one get too out of sync w/ the base.

Issue #121, opened 2020-06-26
Make `hx-push-url` more flexible


bencroker:
Currently, [hx-push-url](https://htmx.org/attributes/hx-push-url/) accepts `true` or `false` values only. Would it be possible to make `hx-push-url` accept a custom string as well?

For example, send a request to `/account` and set the URL to `/account/home`.

```html
<div hx-get="/account" hx-push-url="/account/home">
  Go to My Account
</div>
```

chg20:
Great idea, pretty easy to implement.  You want to give it a crack?

bencroker:
Sure thing!

PR #122, opened 2020-06-26
Allowed `hx-push-url` to accept a string

bencroker:
Re: https://github.com/bigskysoftware/htmx/issues/121

Tried my hand at writing a test this time ;)

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1123,8 +1123,14 @@ return (function () {
         }
 
         function shouldPush(elt) {
-            return getClosestAttributeValue(elt, "hx-push-url") === "true" ||
+            var pushUrl = getClosestAttributeValue(elt, "hx-push-url");
+            return (pushUrl && pushUrl !== "false") ||
                 (elt.tagName === "A" && getInternalData(elt).boosted);
+        }
+
+        function getPushUrl(elt) {
+            var pushUrl = getClosestAttributeValue(elt, "hx-push-url");
+            return (pushUrl === "true" || pushUrl === "false") ? "" : pushUrl;
         }
 
         function addRequestIndicatorClasses(elt) {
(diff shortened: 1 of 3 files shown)


www/attributes/hx-push-url.md
```diff
@@ -7,7 +7,7 @@ title: </> htmx - hx-push-url
 
 The `hx-push-url` attribute allows you to "push" a new entry into the browser location bar, which creates
 a new history entry, allowing back-button and general history navigation.  The possible values of this
-attribute are `true` and `false`.
+attribute are `true`, `false` and a custom string.
```
chg20, line 10: Could you change this to "or a custom string", I think the original sentence was incorrect. :/

src/htmx.js
```diff
@@ -1123,10 +1123,16 @@ return (function () {
...
         function shouldPush(elt) {
-            return getClosestAttributeValue(elt, "hx-push-url") === "true" ||
+            var pushUrl = getClosestAttributeValue(elt, "hx-push-url");
+            return (pushUrl && pushUrl !== "false") ||
                 (elt.tagName === "A" && getInternalData(elt).boosted);
         }
 
+        function getPushUrl(elt) {
+            var pushUrl = getClosestAttributeValue(elt, "hx-push-url");
+            return (pushUrl === "true" || pushUrl === "false") ? "" : pushUrl;
```
chg20, line 1133: could you return null rather than an empty string here?  I don't like taking advantage of the empty string coercion, null feels more correct to me.

www/attributes/hx-push-url.md
bencroker: Agreed and done.

src/htmx.js
bencroker: Done.

PR #127, opened 2020-06-30
Make indicator and request classes customisable

bencroker:
This makes the `htmx-indicator` and `htmx-request` classes customisable, both in meta tags _as well as_ directly in JavaScript. To accomplish this, the indicator styles are included on page load instead of immediately. Please let me know if this is unsafe to do for any reason @chg20.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1522,7 +1522,7 @@ return (function () {
 
                             var swapSpec = getSwapSpecification(elt);
 
-                            target.classList.add("htmx-swapping");
+                            target.classList.add(htmx.config.swappingClass);
                             var doSwap = function () {
                                 try {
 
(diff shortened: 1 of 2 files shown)


bencroker:
After posting this I realised that the `htmx-settling` and `htmx-swapping` classes should also be made configurable for completeness, however I'm wondering whether all of this is worth the extra code and documentation requirements. Happy for this PR to be scrapped if you find it unconstructive @chg20.

chg20:
I think it's a reasonable change to make them all configurable, but I don't think we should change the styles we insert to match them.  

If you are picky enough to want your own indicator class names, you should be picky enough to write your own CSS for them.  :)

bencroker:
Good point! I'll tweak the PR accordingly.

bencroker:
Added the config variables, but it feels a bit strange to hard-code the classes back into the indicator styles block, so I left them as they are.

https://github.com/bigskysoftware/htmx/blob/3fd502af73d2cdeab8f43e6844ac07afd7558c56/src/htmx.js#L1671-L1680

PR #143, opened 2020-07-17
detect multiple attribute and handle getting values

tapvt:
I found an issue where `<select name='foos[]' multiple>` type inputs were not being handled properly. They would only POST a single option value, even when multiple option values were selected.

This PR resolves that issue and should be compatible back to ~IE8.

The change looks for the `multiple` attribute on the modified element, then re-assigns `value` via map as an array of the selected options.

Thanks for this great library!


tapvt:
I lied. `Array.from()` is not IE compatible. Deal breaker?

1cg:
Hi Ben,

Thank you for this contribution.  I am currently in the middle of a pretty big move and won't be able to review it likely for a few weeks.  A quick glance says it looks good, although I would like to make sure it is at least IE11 compatible.

I think the internal function toArray() should do the same thing:

https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L148

Thank you again, I will review as soon as things settle down

tapvt:
Awesome, thanks @1cg. I'll shoot for IE11 compat and update the PR. No rush as I'm running of my fork for now. Hope the move goes smoothly!

tapvt:
@1cg - Updated to use your internal `toArray()` + `Array.map()`. MDN indicates that `Array.map()` is supported back to IE9.

1cg:
Great change, thank you for writing a test!

Can you retarget to the `dev` branch?

tapvt:
@1cg I can get that done tomorrow morning

1cg:
Hey Ben, just merged this change.  I had to revert the stuff in dist.  In the future, just update the stuff in src.  Not a big deal, but I couldn't accept the pull request via Github due to conflicts.

tapvt:
Got it. Makes sense, thank you!

PR #145, opened 2020-07-20
Fix issue with unquoted id attribute in query selector

rros:
When having a dot in the id-attribute of the response, the handleAttributes function throws an Exception:

> DOMException: Element.querySelector: 'DIV[id=ViewModel.Test]' is not a valid selector

The issue is the attribute-value in the selector isn't quoted.
```javascript
var oldNode = parentNode.querySelector(newNode.tagName + "[id=" + newNode.id + "]");
```

 This causes the exception when there is a dot in the value.

This PR fixes this issue and adds a testcase to the regressions test suite to demonstrate the issue.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -395,7 +395,7 @@ return (function () {
         function handleAttributes(parentNode, fragment, settleInfo) {
             forEach(fragment.querySelectorAll("[id]"), function (newNode) {
                 if (newNode.id && newNode.id.length > 0) {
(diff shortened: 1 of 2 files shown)


tapvt:
Whoa. I'm surprised to see how permissive the [spec for HTML5 ids is](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute).

rros:
Haha, it is! It's pretty common to have dots in ids when using ASP.NET Core with model binding.

1cg:
Thank you!

Issue #148, opened 2020-07-24
Reconnect WebSockets


benpate:
WebSockets don't automatically reconnect, for instance, in the case of a server restart or failover.  I have skimmed through the HTMX source code, but I don't see it attempting to reconnect automatically, and I don't see anywhere that I (as the site developer) can change this behavior.

It seems like this would be a useful addition to the HTMX Websocket handling, and adding it should not break any existing developer expectations.

Here's an article on StackOverflow that outlines the basic idea -- essentially listening for an `onclose` event, waiting 5 seconds (or some other configurable duration), then trying to start a new WebSocket connection.

https://stackoverflow.com/questions/3780511/reconnection-of-client-when-server-reboots-in-websocket

Would you like me to make a pull request?

1cg:
Sure, happy to look at one!

vlcinsky:
Docs at https://htmx.org/api/#config already show `wsReconnectDelay:full-jitter` configuration option what means PR #381 is merged.

Could @benpate or someone else involved in this ticket review and possibly close this issue?

benpate:
:)

PR #150, opened 2020-07-25
Adding websocket retries

benpate:
This patch addresses #148 by adding retries to WebSockets when a connection fails or is closed.  This combines many of the ideas in the [linked Stack Overflow article](https://stackoverflow.com/questions/3780511/reconnection-of-client-when-server-reboots-in-websocket).  

It uses a randomized Exponential Backoff algorithm to try to mitigate large numbers of clients from flooding servers at once -- and allows only a single timer to run, in case the browser triggers multiple "onclose" events in quick succession.

I've tried to match your coding style and naming conventions but this is a new project for me, so feel free to make it fit the rest of the source code.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -750,11 +750,39 @@ return (function () {
         }
 
         function processWebSocketSource(elt, wssSource) {
-            var socket = htmx.createWebSocket(wssSource);
+            
+            var socket
+
+            try {
+                socket = htmx.createWebSocket(wssSource);
+            } catch(err) {
+                maybeRetryWebSocketSource(elt, wssSource)
+                return
+            }
+
+            socket.onopen = function(e) {
+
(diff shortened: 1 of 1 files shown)


1cg:
Looks reasonable to me (I'm not the web socket expert) is this still worth pulling in?  Has it been used in real-world web socket work?

benpate:
No, I have not used this in real-world web socket work.  I did get the algorithm from a blog post that claimed this was "the right way to do it."  For my work (which is also not live yet either) it made more sense to use SSE, which already has retries built-in.

I do still think this is a good step forward for Web Sockets, but it does not make them complete or battle-tested.

1cg:
Closing in favor of https://github.com/bigskysoftware/htmx/pull/381 since that's being actively worked on.  Ben, would you mind reviewing that code?

PR #159, opened 2020-08-14
Close connections on swapInnerHTML

benpate:
Currently, HTMX does not close all WebSocket and SSE connections when a node's innerHTML is swapped.  This pull request adds `closeConnections` calls into the `swapInnerHTML` function, so that connections are not leaked after content is swapped.


1cg:
retarget to `dev` please, thank you!

benpate:
Done.  Thanks for the correction.  I'll update all of my pull requests.

Issue #154, opened 2020-08-03
best way to trigger JS to focus an input text field that was newly swapped in


ptrthomas:
this seems to be a common use case. 

so I have a section of the page swapped with multiple elements that can include multiple `<input>` fields

but I want to focus only one of these fields, i.e. the cursor should be within a specified text field, and positioned after any text value that may be present

I'm finding it hard to figure out the best way to achieve this, I know `afterSettle` may be involved: https://htmx.org/events/#htmx:afterSettle

do suggest what you think would be the best / most elegant way to handle this

ptrthomas:
I figured out one possible approach which may be a massive hack, but posting here in case it helps others and for feedback. I also feel a built-in way to "post process" some JS after a swap would be a nice addition to htmx

so in the HTML chunk returned, my server-side templating adds a `<script>` tag like this. this just needs to be added at the end of all the HTML, and it won't appear on the page

```html
<script id="afterSettle">document.querySelector('#todo2').focus()</script>
```

and all that remains is to do an `eval()` on it. this event handler set up at the root page level does the trick:

```js
      htmx.on("htmx:afterSettle", function (evt) {
        var el = document.getElementById('afterSettle');
        if (el) {
          Function('"use strict";return (' + el.innerHTML + ')')();
        }
      });
```

do you think a built-in way to `eval()` pre-JS and post-JS sent similar to an out-of-band HTML snippet would be a feature request worth considering ?

benpate:
I'd vote for `eval()`-ing any `script` tags in the swapped content.  If performance is a consideration, then perhaps it could be put behind another tag, for example, `hx-script="true"`

I think this would be simpler than adding before/after logic, and less overhead than having to make another magic script id to be baked in to the library.

clarkevans:
Lacking any experience with htmx, doing an `eval` of script tags seems like a simple solution.  It has the nice advantage of keeping the code local to the widget that constructs the corresponding HTML.  Would the <script> tag remain in the DOM, or would it be removed?  Is there a way for the script to get the `id` of the element that was swapped, or must user code keep track of this?  Is there a need for `afterSettle` (simpler is better), are there any reasons why the script would need to run before the swap? What sort of nasty edge cases can we think of?

ptrthomas:
@clarkevans I have been using this in a sample app.

> nice advantage of keeping the code local to the widget that constructs the corresponding HTML

totally agree

> Would the <script> tag remain in the DOM, or would it be removed?

[...]

PR #160, opened 2020-08-14
FEATURE: Support "autofocus" attribute when swapping content

benpate:
This adds an additional check to `makeAjaxLoadTask` that focuses on the first element with the HTML standard "autofocus" attribute set.  This addresses one of the original requests in #154

I've put it right next to the makeAjaxLoadTask function for clarity, but it should probably be organized into a different place, next to other corresponding functions.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -411,8 +411,16 @@ return (function () {
             return function () {
                 processNode(child, true);
                 processScripts(child);
+                processFocus(child)
                 triggerEvent(child, 'htmx:load', {});
             };
+        }
+
+        function processFocus(child) {
+            var el = child.querySelector("[autofocus]")
+            if (el != null) {
+                el.focus()
+            }
         }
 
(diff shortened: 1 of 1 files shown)


1cg:
Looks good!  Can you retarget to the `dev` branch?

benpate:
Done.  Thanks for the correction.

1cg:
One more request: is this testable?  If so, would you mind adding a test?

benpate:
That is a fair request.  I'm not too familiar with the automated tests you're using.  I made sure I didn't BREAK anything, but that's probably not enough.  I'll need to dig in a little deeper to figure out how to make a solid test.  Stay tuned!

1cg:
The tests are pretty straight forward, let me know if you have questions.

1cg:
Added tests and fixed a small issue:

https://github.com/bigskysoftware/htmx/commit/278b5f119645ef1d7a49c0db6338b3be43c513c5

PR #162, opened 2020-08-14
FEATURE: find selector option of the hx-target attribute

nyash:
This adds the `find` selector option to the `hx-target` attribute which allows targetting child descendant nodes.


1cg:
I like it.  Can I ask you to add a few tests for it and retarget to the `dev` branch?

PR #163, opened 2020-08-14
FIXED CHROME issue - "refuse to get unsafe header" warning.

akwatra:
FIXED CHROME issue - refuse to get unsafe header warning.
otherwise you will have to send explicitly header from server.
see LINE: 1500

header-err

I have`t updated dist/min.* Require to compile those, after merge.


1cg:
Thanks!

PR #166, opened 2020-08-17
Added event tests and fixed names

bencroker:


1cg:
Hey Ben, can you retarget to dev?

bencroker:
Yes, done.

bencroker:
Anything else you need from me to merge this @1cg? I'm trying to understand when exactly events are triggered and seeing some unusual results, which is on of the motivations for having these new tests in place.

1cg:
What are you seeing?  I had problems in intercooler where I was accidentally triggering events on the replaced content rather than the replacement content.  Maybe that's still an issue?

bencroker:
After some further investigation, it appears that `htmx:afterSettle` and `htmx:afterSwap` are triggered on the target (or each of the replacement elements if `outerHTML` was specified), whereas `htmx:afterOnLoad` and `htmx:afterRequest` are triggered on the initiating element.

I've created a little demo at https://putyourlightson.dev/htmx.html. Open the console tab of dev tools to see the various events being fired.

Screenshot 2020-08-31 at 10 21 57

bencroker:
This also makes me wonder, when using hyperscript, on which elements the `_=""` code should be added. Since events bubble up, could it be added to any element (or parent element of) the element that triggers an event?

Issue #130, opened 2020-07-03
Limit number of elements | Pagination Support


sirinath:
Can an option to be added to how many elements that can be added when using:
- `beforebegin`
- `afterbegin`
- `beforeend`
- `afterend`

with the option to prepend, append and ignore newer times. When prepend and append the oldest item will be removed when the limit is hit.

Also add the ability to:
- specify an ID range or list of item to display based on a state or variable - if returned IDs are outside the list they will not be displayed
- request multiple times - range or the list of items can be specified as variables
- update specific items at particular indices - IDs can be returned in the response header. IDs not in already present can be ignored, prepended or appended.
- insert specific items at particular indices - IDs of adjacent items can be returned in the response header. If items are not adjacent. new items will not be inserted to prevent double insertion.
- delete specific items at particular indices - IDs can be returned in the response header

chg20:
```html
  <div hx-swap="beforeend limit:10 trim:end" ...>...</div>
```

sirinath:
We need to be able to specify what end new items are added or which end items are trimmed. No need to specify both as one implies the other.

chg20:
Good point:

```html
  <div hx-swap="beforeend limit:10" ...>...</div>
```

sirinath:
We need:
- `<div hx-swap="beforeend limit:10 trim:end" ...>...</div>` - add at the front and trim at the end or 
- `<div hx-swap="beforeend limit:10 add:end" ...>...</div>` - add at the end and trip at the front

sirinath:
All this might be a overkill so just `<div hx-swap="beforeend limit:10 trim:end" ...>...</div>` and `<div hx-swap="beforeend limit:10 trim:start" ...>...</div>` maybe just what is needed.

benpate:
Hey @sirinath -- from this discussion, it looks like you're still just discussing this feature, and that it didn't get built, yet.  Is that right?  Let me know if you're still waiting on it and I might take a crack at writing a PR.

sirinath:
I am waiting for these features with:
- pagination
- infinite scrolling
support.

benpate:
Hi @sirinath,

I've spent a little time on this, and have made some progress.  But I'm not sure that this is complete.  First, check out my PR #169, and tell me what you think.  It should support the following:

```html
<div hx-get="/myResource?page=1" hx-swap="afterend limit:10" hx-trigger="revealed"></div>
[...]

PR #169, opened 2020-08-19
FEATURE: Limit number of child elements in hx-swap

benpate:
This pull request partially addresses #130 by adding an optional `limit:` argument to the `hx-swap` attribute.  It looks at where new nodes are being added into the DOM, and removed nodes from the opposite end.  So, since `hx-swap="afterbegin"` adds new elements to the top of the node's list of children, `hx-swap="afterbegin limit:10"` will remove elements from the bottom of the node's list of children until there are only ten.

I think there may be more work to do here, and will continue that discussion in the original issue #130.

Along the way, this code also addresses a minor bug in the `maybeReveal` function that can crash if the user scrolls too quickly.  We address this by double checking that nodes have been initialized before allowing the "reveal" trigger to continue.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -730,7 +730,7 @@ return (function () {
 
         function maybeReveal(elt) {
             var nodeData = getInternalData(elt);
-            if (!nodeData.revealed && isScrolledIntoView(elt)) {
+            if (isScrolledIntoView(elt) && (nodeData.revealed != true) && (nodeData.verb != undefined)) {
                 nodeData.revealed = true;
                 issueAjaxRequest(elt, nodeData.verb, nodeData.path);
             }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -31,6 +31,18 @@ return (function () {
             }
         }
 
+        function parseInteger(str) {
+            try {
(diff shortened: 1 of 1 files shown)


1cg:
Finally getting a chance to look at this.  Code looks good, but I notice some code related to scrolling in here.  Did two branches cross pollute?

benpate:
Honestly, this was an interesting idea, but I think I'd scrap it, now.  Though it works well enough on its own, it would really need more work to be actually useful.  For instance, I started working on a previous/next mechanism so that when items were added to the top/bottom of a container we could still scroll back to the original position.  I think most of the discussion happened in issue #130, if you want to refresh your memory.

It's a pretty cool enhancement to infinite scrolling, and I'm happy to take another shot at it, but it feels incomplete to me as it is, and we should get some architectural guidance from @1cg before we put too much more time into it.

1cg:
Roger, I'll close it for now and we can revisit after you are world famous with the preload extension...

benpate:
Not sure I want to be world famous, but if it does happen, I promise to give you a guest spot on my podcast:  "The @benpate experience"

PR #172, opened 2020-08-24
Fixes htmx not defined when script is loaded deferred 

### LINKED ISSUE #170: deferred script htmx not defined
`<script defer src="https://unpkg.com/htmx.org@0.0.8"></script>`  you get a htmx not defined error

rros:
There are some global references to htmx in the source. When loaded deferred this causes an htmx is not defined error. To fix this, I created the variable `htmx` with the object which gets return at the end of the module.

Fixes #170

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -11,6 +11,47 @@
 return (function () {
         'use strict';
 
+        // Public API
+        var htmx = {
+            onLoad: onLoadHelper,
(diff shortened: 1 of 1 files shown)


bencroker:
Nice fix!! Can confirm that this indeed does fix the issue.

1cg:
thanks!

Issue #168, opened 2020-08-19
Scroll Event Throttling


benpate:
I'm running into a situation where `hx-trigger="revealed"` is breaking if I scroll through elements too quickly.  I think it's in part because multiple onscroll events are overlapping each other.  I'll do more digging before I say I have a good handle on the issue, though.

Here's an article that presents [an alternate way of handling scroll events](https://www.benmarshall.me/attaching-javascript-handlers-to-scroll-events/).  Basically setting a flag to guarantee that only one event handler is running at any given time.  There may be better ways to handle this, but it's a good start.

benpate:
I included [code to address this issue ](https://github.com/bigskysoftware/htmx/pull/169/commits/2f0854e65b180c726fb4b31fe03674976c4e7eca) in #169.  It could easily be turned in to its own PR, if we decide not to merge the original PR.

1cg:
Ah, that explains that code.

Yes, please pull that out to a separate commit.

benpate:
Done.

PR #182, opened 2020-09-02
FIX: Scroll Event Throttling

benpate:
[As requested](https://github.com/bigskysoftware/htmx/issues/168#issuecomment-685244091), I'm moving this into its own PR.  

My apologies for the lack of tests.  I'm still trying to get my head around the testing tools that HTMX uses.  This did pass all of my manual tests.  In case it's helpful, I've posted my test page here: http://sseplaceholder.openfollow.org/scroll.html

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -57,6 +57,8 @@ return (function () {
             return "[hx-" + verb + "], [data-hx-" + verb + "]"
         }).join(", ");
 
+        var windowIsScrolling = false // used by initScrollHandler
+
         //====================================================================
         // Utilities
         //====================================================================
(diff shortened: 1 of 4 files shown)


benpate:
In case it's helpful, I've posted a demo here -- http://sseplaceholder.openfollow.org/scroll.html

1cg:
This doesn't look super easily testable to me.  I do have manual tests that I (sometimes) remember to run before releases.  Take a look at `/tests/manual`

Maybe include your existing test in there?

Anyway, the change looks good.  I am going to merge it in the next day or so.

benpate:
Thanks for merging this into dev.  I've just updated this pull request with a new manual test page that demonstrates the new scrolling behavior.  So, you may want to re-merge, so that the new test/demo is included. :)

1cg:
@benpate moved the polling interval inside the conditional, so we aren't setting it up every time `initScrollHandler()` is called:

https://github.com/bigskysoftware/htmx/commit/5f300f1bbca867cb7455e4ee802cf98dea430d25

benpate:
Good catch. Thank you!

o-o-o-o-o:
Instead of scroll throttling, would it make more sense to use [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)?

benpate:
This is a great suggestion. I did not know about this API, and on first glance, it would make sense to use it. 

From the article you linked, it looks like it is not supported by IE, which is a current browser target for HTMX. 

We might consider a parallel implementation for newer browsers, but that might cause maintainability headaches down the road. 

I'll defer to @1cg on browser support issues. But, let me know if (and how) you'd like to implement this API, and I'm happy to take a crack at it.

1cg:
We are targeting IE11 for now.  I don't see a lot of perf complaints around the revealed stuff, but if I do I'll consider a separate IE11 code path.

PR #188, opened 2020-09-07
Pullrequest 304 not modified

benpate:
This PR demonstrates updates one test case to demonstrate the behavior that should occur when HTMX receives a 304 "Not Modified" response from a server.  As I mentioned in [this comment](https://github.com/bigskysoftware/htmx/commit/d5f6ef29005d8d5bf664f0e0ba276cb4576fbcff#commitcomment-42102045) HTMX should still be able to swap content into the DOM when the browser uses content from the cache.


1cg:
Hey Ben,

Yeah, I was too quick to throw that in, you are right.  I have reverted it in dev.

This pull request has a ton of changes in it, so I'm not going to accept it, but if you want to redo it w/ just the test I can pull it over.

benpate:
Sorry. I just tried to add the one test, but must have included other stuff. I'll resubmit.

PR #192, opened 2020-09-13
FEATURE: Swap HTML on Server Sent Events (take 3)

benpate:
This PR implements the [new SSE syntax](https://gist.github.com/1cg/5dd26530aeec92a9389b367747bbb5ee) proposed by @1cg.  I'll briefly summarize his spec here:

### Swapping on Named Events

This PR adds a new `swap on` syntax to the hx-sse attribue that can be used to swap content into the DOM for any named SSE event. 

```html
  <div hx-sse="connect /event_stream">
    <div hx-sse="swap on chatter">
      ...
    </div>
  </div>
```
The `swap on` indicates that this element wants to listen for `chatter` messages and swap the body when they occur.  

The `hx-swap` attribute can then be used to determine exactly how the swap is done, as with normal swaps.

### Data-only Messages

In the case of a data-only message, the message type is `message` according to the SSE spec.  Thus, the form `swap on message` should work the same as the named event implementation.

```html
  <div hx-sse="connect /event_stream">
    <div hx-sse="swap on message">
      ...
    </div>
  </div>
```

### Declaring and Using An SSE Endpoint On The Same Element

This PR supports both connecting and listening for an event (or events) in the same `hx-sse` declaration, by
allowing comma separated values:

```html
  <div hx-sse="connect /event_stream, swap on message">
     ...
  </div>
```

### Leave Triggers In Place
This PR does not change the way that SSE can be used to trigger additional events in HTMX.  So the following, original code still works:

```html
  <div hx-sse="connect /event_stream">
    <div hx-get="/chatroom" hx-trigger="sse:chatter">
      ...
    </div>
  </div>
```

## Links

1. MDN SSE Article - <https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events>
1. SSE Spec - <https://html.spec.whatwg.org/multipage/server-sent-events.html>
2. Initial SSE improvement PR - <https://github.com/bigskysoftware/htmx/pull/155>
3. Second SSE improvement PR - <https://github.com/bigskysoftware/htmx/pull/185>
4. Updated SSE syntax proposal - <https://gist.github.com/1cg/5dd26530aeec92a9389b367747bbb5ee>


1cg:
Do you know why this PR includes so many other commits?

https://github.com/bigskysoftware/htmx/pull/192/commits

:confused:

benpate:
Shoot.  I still seem to be messing up the base branch.  Resubmitting it now...

PR #193, opened 2020-09-13
FEATURE: Swap HTML on Server Sent Events (take 4)

benpate:
This PR implements the [new SSE syntax](https://gist.github.com/1cg/5dd26530aeec92a9389b367747bbb5ee) proposed by @1cg, and fixes [my previous PR](https://github.com/bigskysoftware/htmx/pull/192) that was built upon the wrong base branch  I'll briefly summarize his spec here:

### Swapping on Named Events

This PR adds a new `swap on` syntax to the hx-sse attribue that can be used to swap content into the DOM for any named SSE event. 

```html
  <div hx-sse="connect /event_stream">
    <div hx-sse="swap on chatter">
      ...
    </div>
  </div>
```
The `swap on` indicates that this element wants to listen for `chatter` messages and swap the body when they occur.  

The `hx-swap` attribute can then be used to determine exactly how the swap is done, as with normal swaps.

### Data-only Messages

In the case of a data-only message, the message type is `message` according to the SSE spec.  Thus, the form `swap on message` should work the same as the named event implementation.

```html
  <div hx-sse="connect /event_stream">
    <div hx-sse="swap on message">
      ...
    </div>
  </div>
```

### Declaring and Using An SSE Endpoint On The Same Element

This PR supports both connecting and listening for an event (or events) in the same `hx-sse` declaration, by
allowing comma separated values:

```html
  <div hx-sse="connect /event_stream, swap on message">
     ...
  </div>
```

### Leave Triggers In Place
This PR does not change the way that SSE can be used to trigger additional events in HTMX.  So the following, original code still works:

```html
  <div hx-sse="connect /event_stream">
    <div hx-get="/chatroom" hx-trigger="sse:chatter">
      ...
    </div>
  </div>
```

## Links

1. MDN SSE Article - <https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events>
1. SSE Spec - <https://html.spec.whatwg.org/multipage/server-sent-events.html>
2. Initial SSE improvement PR - <https://github.com/bigskysoftware/htmx/pull/155>
3. Second SSE improvement PR - <https://github.com/bigskysoftware/htmx/pull/185>
4. Updated SSE syntax proposal - <https://gist.github.com/1cg/5dd26530aeec92a9389b367747bbb5ee>


1cg:
:) can you retarget to `dev`?

Almost there!

benpate:
Yes, of course.  This is so frustrating.  I think I'm going to need some remedial GitHub training once this is done.  I switched my base branch to htmx:dev and everything.  I'll give it a shot later tonight :(

1cg:
No worries git is hard for me too.  If push comes to shove we can just make a patch, and pretend it's '95 again.

clarkevans:
Hi.  I recommend not using a comma as a delimiter (https://gist.github.com/1cg/5dd26530aeec92a9389b367747bbb5ee#gistcomment-3452723).

benpate:
Closing in favor of #194

PR #196, opened 2020-09-15
DOCUMENTATION: Server Sent Events

benpate:
This documents the current state of the SSE PR #194.  I know we're still discussing updates to the syntax, and this will be updated if that changes.

Please let me know if this is too much or too little for the descriptions.  I tried to mirror the existing documentation style, but (obviously) feel free to make changes so that everything matches.


1cg:
thank you!

PR #202, opened 2020-09-24
Add required fields for path-deps (fixes #201)

### LINKED ISSUE #201: Missing `xhr.url` and `xhr.method` for path-deps to work
Hello,

first of all, thanks again for developing HTMX. I was already a big fan of Intercooler, and HTMX is a very nice reboot in its spirit.

Here is one thing I noticed when changing part of a code base from Intercooler to HTMX: It seems the _deps_ mechanism is not working. Perhaps I'm missing something, but according to the website, I do the following:

* Load `htmx.min.js`
* Load `ext/path-deps.js`

Add `hx-ext` to an element/to a parent element, and then e.g.:

	<p hx-get="/test" hx-trigger="path-deps" path-deps="/test/post">Awaiting submit</p>
	<p hx-post="/test/post">Submit</p>

So that `GET /test` should be called when `POST /test/post` is being sent. However, it doesn't work like that, and it seems because something is missing in either `path-deps.js` or `htmx.js`:

In `path-deps.js`, `xhr.method` and `xhr.url` are being accessed in these lines, even though they don't seem to be standard fields of an XMLHttpRequest (https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/XMLHttpRequest):

https://github.com/bigskysoftware/htmx/blob/bd94f7a41fd87b59671aa706d6d12fbb3895eda9/src/ext/path-deps.js#L23
https://github.com/bigskysoftware/htmx/blob/bd94f7a41fd87b59671aa706d6d12fbb3895eda9/src/ext/path-deps.js#L27

Do you perhaps add these fields manually to the `xhr` object in `htmx.js`?

In any case, I can get `path-deps` to work by adding these lines to `htmx.js`, e.g. after this line:

https://github.com/bigskysoftware/htmx/blob/bf5e1804acf2431dee7a213f48ad37c1267bc6b4/src/htmx.js#L1627

	...
	// copy out in case the object was overwritten
	path = requestConfig.path;
	verb = requestConfig.verb;
	headers = requestConfig.headers;
	filteredParameters = requestConfig.parameters;
	xhr.url = path;
	xhr.method = verb.toUpperCase();

Let me know if I misunderstood something.

Eroica:
In path-ext.js, the fields `xhr.url` and `xhr.method` are required.
The missing fields can be added after creating the `xhr` object.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1625,6 +1625,10 @@ return (function () {
             verb = requestConfig.verb;
             headers = requestConfig.headers;
             filteredParameters = requestConfig.parameters;
+            
+            // Required fields for path-deps.js
+            xhr.url = path;
+            xhr.method = verb.toUpperCase();
 
             var splitPath = path.split("#");
             var pathNoAnchor = splitPath[0];


1cg:
As @bencroker points out, I don't think these should be added to the xhr object.  That's a testing framework implementation detail bleeding into the implementation.

The right thing is to pass this information in to the context for the event.

PR #214, opened 2020-10-21
Fixed defaultSwapStyle not used for extension’s handleSwap fallback

jreviews:


1cg:
Thank you!  One thing that worries me here is the possibility for an infinite loop if a bad swap value is passed in.

Can you post the source from the extension so I can see how the swap is being used?  We might want to add another parameter to swap here to indicate if innerHTML should be used or not.

jreviews:
Sure, this is what I have:

```js
handleSwap: function (swapStyle, target, fragment) {
	if (typeof morphdom === 'undefined') {
		return false
	}
	
	if (swapStyle === 'morphdom') {
		morphdom(target, fragment.outerHTML, {
			onBeforeElUpdated: (from, to) => {
				// From Livewire - deal with Alpine component updates
				if (from.__x) {
					window.Alpine.clone(from.__x, to)
				}
			},
		})

		return [target] // let htmx handle the new content
	}
},
```

jreviews:
@1cg Would it work if the extension returns the swap method string it wants to use when handleSwap is aborted?

So if you return `innerHTML`, `outerHTML`, etc. then that's the swap method used? Otherwise it defaults to innerHTML? Then it's up to the extension to return the global swap if that's what it wants.

1cg:
Is this issue still outstanding?  I'd like to swap the context back into my head so I can figure out a good solution...  Sorry it's been open so long :/

jreviews:
I think it is still an issue. In the code example above, say you register the "morphdom" extension, but morphdom is not actually loaded on the page, so you want to skip the processing of the swap. htmx will default to using innerHTML instead of the configured global swap method which could be outerHTML or something else.

1cg:
I finally understand.  It only took... *squints* a year and change.

I will merge and write a test.

PR #218, opened 2020-10-26
 Website: demo modal dialogs using Bootstrap

benpate:
This PR adds an online demo for showing modal dialogs using Twitter Bootstrap.  

I can't get the SASS plugin working with eleventy, so I'm guessing about the global page CSS. It looks like the UIKit stylesheet may mess up the global look of the page. If this is a dealbreaker (or can't be fixed) then we may not be able to use this commit.


1cg:
thank you!

Issue #224, opened 2020-11-06
htmx:beforeRequest Event Stops Triggering if Request is Aborted


ChexWarrior:
Hi everyone,

I've noticed that if I listen for a [htmx:beforeRequest](https://htmx.org/events/#htmx:beforeRequest) event and then cancel the request in that event, subsequent triggers never fire and the event is never called again.

Here is a minimal example that shows the issue:
```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>Document</title>
</head>
<body>
  <label for="block">Block Request?</label>
  <input id="block" type="checkbox">
  <br>
  <button hx-get="https://httpbin.org/html"
          hx-target="#target">
        Click Me
  </button>
  <div id="target"></div>
  <script src="https://unpkg.com/htmx.org@0.3.0"></script>
  <script>
    const btn = document.querySelector('button');
    const checkbox = document.querySelector('input');

    btn.addEventListener('htmx:beforeRequest', function(e) {
      console.log('before request...');
      if (checkbox.checked) {
        e.detail.xhr.abort();
      }
    });
  </script>
</body>
</html>
```
This example just grabs some html from [httpbin](https://httpbin.org/#/Response_formats/get_html) and swaps it with the div.target element when the button is clicked, however if you check the box above the button and then click the request is canceled.  This all works as I would expect, but after the request is canceled in this way the event is never triggered again and the button no longer issues a request to httpbin, I would expect the event to still be triggered every time I click the button.  Am I misunderstanding something or is this a bug?

Thank you in advance

ChexWarrior:
Okay, still not 100% clear on this yet, but it looks like the issue lies in the issueAjaxRequest function within htmx.js.  The code checks if the element sending the request already has a request in progress, if it does it queues the request and exits the function, otherwise it sets the requestInFlight flag to true and continues.  When I run my example above with the checkbox enabled it aborts the XmlHttpRequest and never sets the requestInFlight flag back to false.  When I subsequently attempt to trigger the ajax call again it sees that a request is still in progress and returns without doing anything.

PR #225, opened 2020-11-06
End request lock when xmlHttpRequest is aborted

ChexWarrior:
This is a fix for the issue I raised in #224.  

I'm new to this code base so this is very much WIP, but I've added what I think addresses the main problem and welcome any feedback.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1998,6 +1998,10 @@ return (function () {
                 triggerErrorEvent(elt, 'htmx:sendError', eventDetail);
                 endRequestLock();
             }
+            xhr.onabort = function() {
+                removeRequestIndicatorClasses(elt);
+                endRequestLock();
+            }
             if(!triggerEvent(elt, 'htmx:beforeRequest', eventDetail)) return endRequestLock();
             addRequestIndicatorClasses(elt);
 
--- a/test/index.html
+++ b/test/index.html
@@ -51,6 +51,9 @@
     <li>
         <a href="manual/sse-settle.html">SSE - Settling</a>
     </li>
+    <li>
+      <a href="manual/aborting-request.html">Aborting Request Test</a>
+    </li>
 </ul>
 
 <h2>Mocha Test Suite</h2>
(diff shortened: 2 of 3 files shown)


1cg:
looks good to me

is there a way to add a test?

ChexWarrior:
Sure thing, I will look into creating a test!

ChexWarrior:
@1cg Running into a slight issue with the tests. It seems Sinon doesn't like it when you abort an XMLHttpRequest before it sends.  Here is test I'm working on (in test/core/events.js):
```js
    it.only("htmx:beforeRequest continues to fire after aborting a request", function () {
        var counter = 0;
        var handler = htmx.on("htmx:beforeRequest", function (evt) {
            counter += 1;
            evt.detail.xhr.abort();
        });
        try {
            var div = make("<form hx-trigger='click' hx-post='/test'><input name='param' value='foo'></form>");
            div.click();
            div.click();
            should.equal(counter, 2);
        } finally {
            htmx.off("htmx:beforeRequest", handler);
        }
    });
```

The event handler is called with no issue on the first div.click(), however when the xhr goes to send within issueAjaxRequest Sinon throws an error in its verifyState method because the xhr object's readyState property is 0 instead of 1.  It seems that Sinon doesn't expect you to abort a request berfore sending it however this seems allowed on the actual browser implementation of XMLHttpRequest.  Does this seem like a bug in Sinon or should we add logic to prevent xhr.send() if the xhr has already been aborted?

1cg:
OK, might not be easily testable w/ sinon.

Would you mind looking at how hard it would be to add a manual test to the /test/manual directory for it and use the normal xhr support?

I run through those as well before releases manually to verify things are working.

ChexWarrior:
I've added a test that is essentially the same as the minimal example I used in #224.  I've setup a beforeRequest event that adds a message to the target area's innerHTML, this message should continue to be appended even if a previous beforeRequest event canceled the current XMLHttpRequest.

PR #232, opened 2020-11-11
Allow expression vars to override raw parameters

bencroker:
This PR allow expression vars to override raw parameters.

Before:
```html
<form hx-get="">
    <inpup type="hidden" name="temperature" value="0">
    <button type="submit" hx-vars="temperature: -23">Go sub-zero</button>
</form>

<!-- Submits {temperature: 0} -->
```

After:
```html
<form hx-get="">
    <inpup type="hidden" name="temperature" value="0">
    <button type="submit" hx-vars="temperature: -23">Go sub-zero</button>
</form>

<!-- Submits {temperature: -23} -->
```

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -993,9 +993,11 @@ return (function () {
                 elt.addEventListener(getTriggerSpecs(elt)[0].trigger, function (evt) {
                     var headers = getHeaders(elt, webSocketSourceElt, null, elt);
                     var results = getInputValues(elt, 'post');
-                    var rawParameters = results.values;
                     var errors = results.errors;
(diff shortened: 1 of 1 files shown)


bencroker:
@1cg I just found a test called `hx-vars do not override inputs`, so this is by design?
https://github.com/bigskysoftware/htmx/blob/dev/test/attributes/hx-vars.js#L62-L73

1cg:
I don't see a compelling reason that it was done this way.  Seems like the hx-vars is more local to the call and so it should have precedence.

I'll pull and modify the test.

bencroker:
Thanks!

Issue #238, opened 2020-11-18
HTMX forms only support multi-selects if the multiple attribute is set and not just listed


sarahboyce:
Been using htmx on a django project and here multi-selects are indicated when the `multiple` attribute is listed in the select element
In HTMX the attribute has to be set:
https://github.com/bigskysoftware/htmx/blob/1f028b14da0e8d9a39221f426ef7cd266b92df02/src/htmx.js#L1479

Could this be updated so that multiselects are whenever the attribute is listed so that this is compatible with django forms?

1cg:
Hi Sarah.

Can you give me some example html?  The best approach is to build out a test so I can guarantee I fix the issue.

sarahboyce:
> Hi Sarah.
> 
> Can you give me some example html? The best approach is to build out a test so I can guarantee I fix the issue.
Sorry it took me so long to respond @1cg ^

This is the html I have generated:
````
<select name="question_list" data-minimum-input-length="0" data-allow-clear="true" data-placeholder="" data-maximum-selection-length="3" class="form-control django-select2 select2-hidden-accessible" title="" id="id_question_list" multiple="" tabindex="-1" aria-hidden="true">
  <option value="1">Question 1?</option>
  <option value="2">Question 2?</option>
  <option value="3">Question 3?</option>
  <option value="4">Question 4?</option>
  <option value="5">Question 5?</option>
</select>
```

1cg:
Ah, so the problem is that the attribute has an empty value, correct?  But it still behaves as a mutli-select in the browser...

tapvt:
@1cg - I've got a test written and a PR inbound shortly

bencroker:
@1cg I can confirm that htmx only submits the first selected value in a multi-select field.

1cg:
Hi @sarahboyce, this issue is fixed in `dev` and should be out with the next release (sometime this week)

https://github.com/bigskysoftware/htmx/commit/fda4957d653ebf683d6017c07a6ef34ed2d72983

sarahboyce:
You stars thank you all 🌟

PR #243, opened 2020-11-25
handle multi-select when "multiple" attr is present w/o value

tapvt:
Attempted resolution to #238


1cg:
I accepted this pull and added the `false` check as well.

Thank you!

PR #246, opened 2020-11-30
Fix handling of 'multiple' attribute

ojensen5115:
Typically a `multiple` attribute element does not have a value, but is instead included in the same way as `disabled`, `selected`, `checked`, etc., i.e. `<select name="my_selector" multiple>`. The current implementation fails to recognize the `multiple` attribute in this form because it attempts to read the attribute's value, determines that it is the empty string, then double-negates this to end up with `false`.

According to caniuse.com, the `multiple` attribute can be read directly from an [input](https://caniuse.com/mdn-api_htmlinputelement_multiple) or [select](https://caniuse.com/mdn-api_htmlselectelement_multiple) element via `element.multiple` in every major browser.


1cg:
Thank you, this is fixed in `dev` but I like your fix better.

Can you retarget to the dev branch and I'll pull it in?

ojensen5115:
Hm, my git-fu isn't as good as it should be. I've switched the branch from `master` to `dev`, but the PR was based off of `master`'s HEAD. I'm not sure how to fix this besides closing the PR and opening a new one. Should I do so?

PR #253, opened 2020-12-04
Add hx-oob-target

dkniffin:


dkniffin:
As agreed upon in Discord, it would be better to change this behavior a bit:

- [ ] Instead of adding a new attribute `hx-oob-target` for this, we should use the value of `hx-swap-oob` for the OOB target selector
- [ ] Currently, that value is used to specify the swap behavior of OOB swaps (eg `innerHtml`). We should change it to use the value in `hx-swap` to be consistent and re-use that attribute
- [ ] To make this a non-breaking change, since the set of values for `hx-swap` is distinct from the selectors we'd be using here, we can add a bit of logic to check for the `hx-swap` values in `hx-swap-oob` and handle them accordingly
- [ ] Since we don't want to support having `hx-swap-oob` support two types of values indefinitely, we can deprecate the `hx-swap` attributes, and in 2.0, we'll drop support for it.

I'll try to take care of this next week, if I find time.

1cg:
Thinking on this more, I see a reason not to use hx-swap:  the new content might legitimately want to use it for a future triggered request:

```html
<div id="div1" ic-get="/whatever" ic-swap="innerHTML" oob-swap="true">...</div>
```
So with that, I would like to propose that we extend the oob-swap syntax like so:

```
oobSwapSpec => swapType | cssSelector | swapTypeAndSelector
swapType => "innerHTML", "outerHTML", ...
cssSelector => <any valid CSS selector>
swapTypeAndSelector => swapType ':' cssSelector
```
You can have a swap type  or a css selector or a swap type followed by a colon, followed by a css selector.  This allows for maximum flexibility and is easy to implement (just split on the first colon) and is backwards compatible.  

Examples:

```html
<div oob-swap="innerHTML">...</div>
<div oob-swap="[someProp]">...</div>
<div oob-swap="outerHTML:[someProp]">...</div>
```

1cg:
I decided to do a poor mans version of this feature.  If you want to use a CSS selector, you need to explicitly pick a swap strategy, so something like this:

```html
<div oob-swap="outerHTML:[someProp]">...</div>
```
That way we can know that that first colon is a delimiter and not part of the CSS selector.  Just makes it a lot simpler for a feature that I don't expect to get a ton of use.

dkniffin:
@1cg Sounds good. Sorry I haven't had time to look at this. I actually ended up using IDs for the feature I was working on. I may come back to this at some point, but I'm not sure.

PR #263, opened 2020-12-12
SSE listen

wtpayne:
I made a little and slightly hacky change to support my SSE use-case.
I'm not suggesting that this PR be merged... it's really just here for discussion.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1076,8 +1076,12 @@ return (function () {
                     processSSESource(elt, value[1]);
                 }
 
+                if ((value[0] === "listen")) {
+                    processSSEClient(elt, value[1], false)
+                }
+
                 if ((value[0] === "swap")) {
-                    processSSESwap(elt, value[1])
+                    processSSEClient(elt, value[1], true)
                 }
             }
         }
@@ -1091,7 +1095,7 @@ return (function () {
             getInternalData(elt).sseEventSource = source;
         }
 
-        function processSSESwap(elt, sseEventName) {
+        function processSSEClient(elt, sseEventName, doSwap) {
             var sseSourceElt = getClosestMatch(elt, hasEventSource);
             if (sseSourceElt) {
                 var sseEventSource = getInternalData(sseSourceElt).sseEventSource;
@@ -1101,20 +1105,21 @@ return (function () {
                         return;
                     }
 
-                    ///////////////////////////
-                    // TODO: merge this code with AJAX and WebSockets code in the future.
+                    if (doSwap) {
+                        ///////////////////////////
+                        // TODO: merge this code with AJAX and WebSockets code in the future.
+                        var response = event.data;
+                        withExtensions(elt, function(extension){
+                            response = extension.transformResponse(response, null, elt);
+                        });
+                        var swapSpec = getSwapSpecification(elt)
+                        var target = getTarget(elt)
+                        var settleInfo = makeSettleInfo(elt);
+                        selectAndSwap(swapSpec.swapStyle, elt, target, response, settleInfo)
+                        settleImmediately(settleInfo.tasks);
+                    }
 
-                    var response = event.data;
-                    withExtensions(elt, function(extension){
-                        response = extension.transformResponse(response, null, elt);
-                    });
-
-                    var swapSpec = getSwapSpecification(elt)
-                    var target = getTarget(elt)
-                    var settleInfo = makeSettleInfo(elt);
-
-                    selectAndSwap(swapSpec.swapStyle, elt, target, response, settleInfo)
-                    triggerEvent(elt, "htmx:sseMessage", event)
+                    triggerEvent(elt, "htmx:sse:" + sseEventName, event)
                 };
 
                 getInternalData(elt).sseListener = sseListener;
--- a/test/attributes/hx-sse.js
+++ b/test/attributes/hx-sse.js
@@ -133,5 +133,32 @@ describe("hx-sse attribute", function() {
         byId("d2").innerText.should.equal("Event 2")
     })
 
+    it('does not swap content on SSE listen', function () {
+        var div = make('<div hx-sse="connect:/event_stream">\n' +
+            '  <div id="d1" hx-sse="listen:e1"></div>\n' +
+            '  <div id="d2" hx-sse="listen:e2"></div>\n' +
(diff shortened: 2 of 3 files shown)


src/htmx.js
```diff
@@ -1066,6 +1066,10 @@ return (function () {
                     processSSESource(elt, value[1]);
                 }
 
+                if ((value[0] === "forward")) {
```
1cg, line 1069: I like "listen" rather than "forward" here: 

```html 
<div hx-sse="listen"> ... </div> 
```

wtpayne:
Record of the conversation from Discord:-
[3:19 PM] wtpayne: Let me give the rationale.
[3:19 PM] wtpayne: So ... I've got a dashboard builder.
[3:19 PM] wtpayne: The dashboard components themselves can be dynamically added, removed and modified.
[3:19 PM] wtpayne: For this, sse:swap is perfect.
[3:20 PM] wtpayne: But I also want the dashboard components to display oscilloscope-like graphs...
[3:20 PM] wtpayne: And for the data for the graph to come over as JSON, so I can use D3.js to do the plotting.
[3:21 PM] wtpayne: So I have some situations where I want to do things the HTMX way ... and have a bit of HTML swapping in.
[3:21 PM] wtpayne: And other situations where I want to do things differently, and have JSON coming in over SSE and being used to update a chart or other visualization with D3.js...
[3:22 PM] wtpayne: So I need to be able to say 'do something with the content of this SSE event ... but don't swap it into the DOM'

src/htmx.js
```diff
@@ -1081,6 +1085,24 @@ return (function () {
             getInternalData(elt).sseEventSource = source;
         }
 
+        function processSSEForward(elt, sseEventName) {
```
1cg, line 1088: Rather than creating a new method, what about renaming this to something like `processSSEClient` and passing in a swap boolean?  That would reduce some of the code duplication and is about as clear IMO.

src/htmx.js
```diff
@@ -1081,6 +1085,24 @@ return (function () {
...
+            var sseSourceElt = getClosestMatch(elt, hasEventSource);
+            if (sseSourceElt) {
+                var sseEventSource = getInternalData(sseSourceElt).sseEventSource;
+                var sseListener = function (event) {
+                    if (maybeCloseSSESource(sseSourceElt)) {
+                        sseEventSource.removeEventListener(sseEventName, sseListener);
+                        return;
+                    }
+                    triggerEvent(elt, sseEventName, event)
```
1cg, line 1097: Could we reuse the same code in processSSESwap for triggering the event?

```js
                    triggerEvent(elt, "htmx:sseMessage", event)
```

You would need to look into the `event.detail` object to see the event name, which I guess is a little annoying.

Alternatively, we could adopt this approach (using the SSE event name for the DOM event) in place of the current `htmx:sseMessage` approach in `processSSESwap()`

I don't have strong opinions on this.

1cg (changes requested):
please retarget to `dev` branch as well, thank you for working on this!

src/htmx.js
wtpayne: Yup. I agree with that as well.

src/htmx.js
wtpayne: I think using htmx:sseMessage minimises the chances of a name collision.

src/htmx.js
wtpayne: I don't have a strong opinion on the name, so I defer to you on this. I've made the change as requested.

src/htmx.js
wtpayne: I've made the change.

src/htmx.js
wtpayne: I've made the change.

wtpayne:
> please retarget to `dev` branch as well, thank you for working on this!

Retargeted to 'dev' branch as requested.

wtpayne:
Added tests, and checked that tests pass.

src/htmx.js
```diff
@@ -1091,19 +1095,19 @@ return (function () {
...
+                            response = extension.transformResponse(response, null, elt);
+                        });
+                        var swapSpec = getSwapSpecification(elt)
+                        var target = getTarget(elt)
+                        var settleInfo = makeSettleInfo(elt);
+                        selectAndSwap(swapSpec.swapStyle, elt, target, response, settleInfo)
+                    }
 
-                    selectAndSwap(swapSpec.swapStyle, elt, target, response, settleInfo)
                     triggerEvent(elt, "htmx:sseMessage", event)
```
wtpayne, line 1111: I actually think we need a different name for each event, so I'm changing my mind on this. Going to try some experiments first.

1cg:
```js
                    triggerEvent(elt, "htmx:sse:" + event.name, event)
```
?

wtpayne: I'm trying something like that ... but I need to do some more investigation first.

benpate: I'm jumping in late here, but wouldn't this make it more difficult to listen to different kinds of sse messages?  If there's just one event name, then I'd only need one function, which could just switch on the message name to differentiate.  Maybe it's an issue of architecture.

1cg:
What's the status of this pull request?  Is it ready for review?

wtpayne:
I think so. I've been using it like this without too many issue so far.

wtpayne:
However... I did make a fix for what appeared to me to be a bug ... Settle
wasn't being triggered, so I added it in. It seems to work, but I'm not
sure 100% if it's the right thing to do.

wtpayne:
@benpate - I have a separate callback for each type of SSE event, and it would be exceedingly difficult to merge them together. (The callbacks are themselves transmitted over SSE) so my very strong preference would be to leave the SSE event name in the DOM event name if at all possible.

wtpayne (commented):
I think the call to settleImmediately is the right thing to do ... but I'm not sure.

1cg:
resurrecting this thread: @benpate just pulled sse out to a separate extension and I'd like to revist this change in that context

alexpetros:
It looks like this is stalled out. It in the time since, Server-Side Events have been pulled out to an extension. If the current SSE extension still has the issue, feel free to either re-open this PR or make a new one and we'll get it looked at ASAP.

Issue #254, opened 2020-12-06
Extension point: transformRequest


tomberek:
A nice mechanism to transform a request before it is sent is the dual to the existing `transformResponse`. (suggestion: `transformRequest`)  This would allow extensions that primarily implement these two extension points to create drivers that would allow non-htmx friendly backends to be used transparently.

Real-world example: I wanted to add a "Range" header to requests, but ended up using a global listener on "htmx:configRequest" and some conditionals on the element in question. This paired up with a specific `raw-to-json` -style extension particular to the backend I was using. Being able to bundle the two into a single concept would simplify implementations.

Example: I can imagine writing a driver for various HTTP backends (think postgrest) that need transformation of both the request and response in order to function with HTMX.

benpate:
I have come across another use case where this would be very helpful.  I'm currently using cookies to sign users in to my web application.  But, I'd really like to use localStorage instead.  But, this would require some extension that could find the users session token (JWT in this case) and insert it into each request before it is sent to the server.  This is not currently possible with htmx, but it would be very easy to accomplish once we have this extension point.

benpate:
Hey @tomberek -- after I wrote this PR and dug in to the code a little more, I discovered there's already a way to do this using events.  The [`htmx:beforeRequest`](https://htmx.org/events/#htmx:beforeRequest) event allows you to manipulate the request configuration before it is sent to the server.

Does this work for you?  Are you comfortable closing this issue?

1cg:
I am going to close it, the `htmx:beforeRequest` event now handles this pretty completely and gracefully

tomberek:
The global events can handle the need(I used configRequest in my use case), but I don’t believe it scales. The goal would be to be able to write an extension that can do this. This would also round out the API for extensions. 

Or did something change or get added that I’m not aware of?

PR #265, opened 2020-12-13
Extension Point: transformRequest

benpate:
This PR addresses #254 by adding a `transformRequest` function to extensions.  

I'm also including a simple extension that demonstrates this feature by listening for an "Authorization" header in any XHR responses and storing the value in sessionStorage.  If we have an "Authorization" token in sessionStorage, then this value is included in the headers of any subsequent XHR requests.

### DIFF
--- a/src/ext/authentication-header.js
+++ b/src/ext/authentication-header.js
@@ -1,0 +1,25 @@
+htmx.defineExtension('authentication-header', {
+
+	transformRequest: function(xhr) {
+		
+		if (window.sessionStorage) {
+			let auth = window.sessionStorage.getItem("Authentication")
+			if (auth != "") {
+				xhr.setRequestHeader("Authentication", auth)
+			}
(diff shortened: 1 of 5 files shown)


1cg (commented):
Looks reasonable.  Needs docs for the extension and the new extension method, as well as tests

benpate:
This may overlap with the htmx:beforeRequest event.  Is it good to allow duplicates?  Or should there be one true way?

1cg:
Yeah, `htmx:beforeRequest` is the standard way to modify requests.  I'd like to stick with that for now, although I'm willing to listen to reason...

benpate:
Yes, I agree.  A better, standard way exists already.  "Listening to reason" means keeping it the way it is.

BTW.  Thanks for all of your work on these issues and PRs.  I'll sift through all of your notes this weekend.

Issue #259, opened 2020-12-07
Extension Point: init function


benpate:
Would it be valuable to have an `init()` function for extensions?  This might help in cases where there's some global setup that needs to occur.  It makes sense to me to pass in a few arguments, such as the element where the extension is defined.

**Use Case**
To create a "modal dialog" extension, I'd like to be able to create a DOM node at the bottom of the document where modal content is loaded.  This might also include a modal backdrop DIV, which is common for most modal dialog designs.  

```html
<button hx-ext="uikit-modal" hx-get="modal.html">Open Modal</button>
```

benpate:
With `htmx:beforeProcessNode` I think this request may be moot.  At the very least, it should be on the back burner until after 1.1 and we can see if it's even needed at all.

1cg:
closing until needed by a real world example...

PR #267, opened 2020-12-13
Extension Point: init()

benpate:
This PR addresses #259 by adding an `init()`function to extensions.  This function is called once for every node that lists the extension in `hx-ext=""`.

I know this will require examples and documentation to merge.  I'm happy to add that once the basic idea of this extension point has been blessed.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1216,7 +1216,7 @@ return (function () {
             if (elt.querySelectorAll) {
                 var boostedElts = isBoosted() ? ", a, form" : "";
                 var results = elt.querySelectorAll(VERB_SELECTOR + boostedElts + ", [hx-sse], [data-hx-sse], [hx-ws]," +
-                    " [data-hx-ws]");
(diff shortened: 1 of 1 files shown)


1cg:
I need to think about this one more.  If an extension is on a parent element of the newly initialized content, should it be called with the new content?  I tend to think so...

Also, as always, needs tests and docs...

benpate:
I've given some thought to your comment.  I still don't feel understand the events system thoroughly enough, but here's my take:

If an child element is updated, wouldn't that fire the `htmx:processedNode` event?  If so, why do we need an init() function at all?  I went back and forth on this, and I still believe there is still a place for this.  Alternatively, we could trigger htmx:init event the first time that something is processed by htmx.  The reason here would be to know the first time something is called.

1cg:
Hey ben, I'm going to close this out for now.  We can revisit if a strong use case comes up at some point.

Issue #244, opened 2020-11-27
Inheritance


tomberek:
It sometimes makes it hard to apply an extension or certain attributes when they are automatically inherited by children and when those attributes are not wanted at that lower level. I'm not sure there is a clean design to opt-out of hx-attribute inheritance.

1cg:
Hey Tom,

Can you give me some examples? 

Maybe hx-ext needs to support a `-` syntax too?

```html
<div hx-ext='-some-ext' ...>...</div>
```

tomberek:
My example involved using `client-side-templates` and a custom extension for a component that transforms the raw text to the right kind of JSON (newline delimited info into an array of lines).

Inside that component I wanted a `hx-get` to bypass this processing and just display the raw result. My workaround for the situation was to add a conditional into my custom extension that detected if the element had the right hx-ext attribute.

bencroker:
@tomberek does setting `hx-ext=""` not oveerride the inherited behaviour?

benpate:
@bencroker -- It doesn't look like setting `hx-ext=""` would override the inherited behavior. The `getExtensions` function just [appends extensions to the list from its parents](https://github.com/bigskysoftware/htmx/blob/c0f1e06eab0c7d510e1d8ee95f84ff443a7ede70/src/htmx.js#L2142).  An "override" should be easy to add in, but it's not there now.

bencroker:
Thanks for clarifying @benpate!

1cg:
I would like to introduce a syntax for disabling an extension at a given point.  I don't think just having a blank `hx-ext` attribute is clear enough.  Some proposals:

```html
  <div hx-ext="ignore my-ext">...</div>
  <div hx-ext="remove my-ext">...</div>
  <div hx-ext="exclude my-ext">...</div>
  <div hx-ext="disable my-ext">...</div>
```

I like `ignore` the best, the other ones seem to imply a more global disabling of the extension to me.  What do you all think?

bencroker:
`disable` would probably be my first choice, `ignore` second.

benpate:
I'll just add that we use the `command:argument` pattern that we use in other places, and it would be nice to stay consistent with that syntax.  So I'd recommend `hx-ext="disable:extension-name"` or `hx-ext="ignore:extension-name"`.

I'm not to stuck on any particular command.  But fewer letters (and syllables) are usually less cognitive overhead, so I'd vote for "ignore" or "remove" (if we are, indeed, voting)

[...]

PR #268, opened 2020-12-14
Add "ignore:" command for extensions

benpate:
This addresses issue #244.  I don't know why GitHub says it can't automatically merge :(


1cg:
Will try to review this change soon, thank you!

1cg:
Hey Ben, thank you again for working on this.  I think I'd prefer to just update the signature of `getExtensions` to include an ignore set.  Something like this:

```js
        function getExtensions(elt, extensionsToReturn, extensionsToIgnore) {
            if (elt == null) {
                return extensionsToReturn;
            }
            if (extensionsToReturn == null) {
                extensionsToReturn = [];
            }
            var extensionsForElement = getAttributeValue(elt, "hx-ext");
            if (extensionsForElement) {
                forEach(extensionsForElement.split(","), function(extensionName){
                    extensionName = extensionName.replace(/ /g, '');
                    // if starts with ignore: add to extensionsToIgnore
                    // if not, check if name is not in extensionsToIgnore
                    var extension = extensions[extensionName];
                    if (extension && extensionsToReturn.indexOf(extension) < 0) {
                        extensionsToReturn.push(extension);
                    }
                });
            }
            return getExtensions(parentElt(elt), extensionsToReturn);
        }
```
It's basically the same logic, but does away with the map and filter.  It feels less violent and more along the flavor of this code base to me.  Does that work?

benpate:
Ok.  This has been updated with a version of the code you provided.  All tests are passing.

Issue #256, opened 2020-12-06
Extension Point: Publish xhr() function


benpate:
This is a proposal to publish htmx's xhr feature for external functions to use. The htmx library currently does a lot of work to xhr requests to access enhanced htmx properties in the HTML markup.  It would be useful for external libraries (mostly extensions) to be able to use this, too.  Example use cases:

* A "preload" extension that is able to preload page fragments before they are needed.

benpate:
The new `htmx.ajax()` function is awesome.  Thank you!

PR #273, opened 2020-12-17
Extension: Preload Links

benpate:
This extension preloads selected links into the browser's cache.  It works with all child elements that contain an `href` or `hx-get` attribute.  

As has become my style, I promise I'll make documentation and test cases if the general idea of this PR is approved :) but please let me know if I should change this habit.

Example code:

```html
<body hx-ext="preload">

  <h1>These will be preloaded<h1>
  <a href="yes.html" preload>I will be pre-loaded</a>
  <button hx-get="yes.html" preload>I will also be pre-loaded</button>

  <h1>These will NOT be preloaded</h1>
  <a href="no.html">I will NOT be pre-loaded</a>
  <form hx-post="/bad-idea" preload></form>

</body>
```


benpate:
On further research, I've identified one potential problem with this extension.  Since it does not use the htmx XHR function, it does not include htmx headers in the request.  So, if the web server is depending on these, for instance, to differentiate between a full page load and an htmx-style partial page load, it will not work.  In this case, all `hx-get` requests will look like full page loads, and will potentially include unwanted HTML in the response.

@1cg - This could be corrected if we publish the xhr function used by htmx (per #256)

1cg:
`htmx.ajax` is now available in the `dev` branch, would love to get this merged soon!

benpate:
Awesome.  Thank you.  I'll take another pass at this.

benpate:
Thank you @1cg for your quick work pushing `htmx.ajax` into the dev branch.  Unfortunately, I am running into some issues using it so we'll need to discuss implementation details to figure out how to proceed.

For this preload extension specifically, I want to load content without putting it into the DOM.  The trouble is that `htmx.ajax` also appears to swap content into the DOM.  If the current DOM node is passed in as the `context` argument, the new HTML fragment appears directly in that node.  If the `context` argument is left empty, the fragment replaces the entire document body.  

In either case, this makes it impossible to use for preloading resources, because they show up immediately in the document, and sometimes in very wrong places.  While this extension is the only example I have right now, I think that this could be a problem for other extensions in the future, too.  

There are many ways this could be addressed.  One fix would be to separate the swapping logic from the`htmx.ajax` function -- which makes the most sense to me, and would probably have benefits in other areas.  I think this could be a big win, but I don't know how difficult this would be architecturally, or what other implications this would have for htmx as a whole.

Thanks again for helping to make this extension the best it can be.  I'll jump in and help anywhere you think best, just point me in the right direction and say "go."

1cg:
lol, sorry, I should have documented it better.  like... at all.

You can use this form:

```js
  htmx.ajax("GET", "/whatever", {handler:function() { /* NO OP for cache */ });
```
here is the raw code:

https://github.com/bigskysoftware/htmx/blob/20436f243c46df4e3965cb4a90c8a1941e70b25a/src/htmx.js#L1836

benpate:
Ok.  Thanks!  I swear I looked at the code, but it's.. intense.  I'll see what I can do with this, and am also working on some way to add a `wait:` argument to delay the trigger on mouseovers, too.

benpate:
I've made a number of updates (detailed in the check-in above.  @1cg, could you please review this design before I finalize tests and documentation?  Here's the syntax as it stands right now:

```html
<body hx-ext="preload"> ... Initializes the extension

    <a href="/server">Link</a> ... WILL NOT get preloaded because it needs an explicit attribute
    <button hx-post="/server preload>Button</button> ... WILL NOT preload hx-post, hx-put, or hx-delete transactions

    <a href="/server" preload>Link</a> ... WILL get preloaded.  Default is onMouseDown, 0ms delay
    <button hx-get="/server" preload>Button</button> ... Preloaded using htmx.ajax (so request headers are correct)

    <div hx-get="/server" preload="on:mouseover">Load More</div> ... changes trigger to mouseover, default wait = 100ms
    <div hx-get="/server" preload="on:mouseover wait:200ms">Load More</div> ... set a custom wait period
 
   <div hx-get="/server" preload="on:load">Load More</div> // Preload content immediately with on:load
   <div hx-get="/server" preload="on:load wait:2s">Load More</div> // Or load them 2 seconds after the node is processed

    <div preload="on:mouseover wait:200ms"> ... Default values are inherited by child nodes
        <a href="/server" preload>Link</a> ... but links STILL need "preload" to be processed.
        <a href="/server" preload="wait:50ms">Link</a> ... and can override values (if you REALLY need to)
    </div>

</body>
```

I'm pretty happy with the way this is working right now.  But, this still needs (at least) one more pair of eyes on it before I continue any further.  

Separate from tests and documentation, one future enhancement we might consider would be the ability to start a preload when a node has scrolled into view.  That might mean publishing events from htmx's scroll indicator, which would require a separate amount of work before we begin.

1cg:
OK, I have accepted this into `dev`, thank you!  Now it needs docs.  I guess the testing has to be manual though, eh?

PR #275, opened 2020-12-18
Optimise processInputValue() function

bencroker:
This PR avoids having to keep track of every single processed element, by instead only processing an input value if its `name` does not already exist. This should improve performance but more importantly prevent multiple parameters with the same name from being submitted in a single request. Also, in a non-GET request, the parameters inside the current form will automatically win out over parameters included later down the line.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1472,16 +1472,6 @@ return (function () {
         // Input Value Processing
         //====================================================================
 
-        function haveSeenNode(processed, elt) {
-            for (var i = 0; i < processed.length; i++) {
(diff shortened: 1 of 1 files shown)


1cg:
I think in some cases you want multiple values for a given input name.  Checkboxes, for example, are often used to submit a bunch of values for the same name, allowing you to select a bunch of rows and have them all come in under the same parameter name.

Am I misunderstanding?

bencroker:
Ah of course, that was an oversight on my part. I'll rethink this and resubmit, thanks for looking at it!

PR #277, opened 2020-12-18
Change how getInputValues() works

bencroker:
This PR changes how `getInputValues()` works.

### Before

1. If a non-GET request, collect all input values from the enclosing form.
2. Concatenate the current element's value.
3. Concatenate all remaining values from explicit includes.

### After

1. If a non-GET request, collect all input values from the enclosing form.
2. Collect the current element's value.
3. Collect all remaining values from explicit includes.
4. Merge values from step 3 with values from step 2 with values from step 1.

The key difference is that in a non-GET request, if an element is inside a form then that form's input field values will _override_ input fields that exist outside the form.

### Example

```html
<div hx-include="*">
  <input hx-get="/" type="text" name="secret" value="abc">
  <form hx-post="/" id="form-1">
    <input type="text" name="id" value="1">
    <input type="submit" value="Submit">
  </form>
  <form hx-post="/" id="form-2">
    <input type="text" name="id" value="2">
    <input type="submit" value="Submit">
  </form>
</div>
```

### Before

Changing `secret` results in the following request params:
```
secret=abc
id=1
id=2
```

Submitting `form-1` results in the following request params:
```
secret=abc
id=1
id=2
```
*** **The `id` above will be evaluated to `2` by most servers since it comes last.** ***

Submitting `form-2` results in the following request params:
```
secret=abc
id=1
id=2
```

### After

Changing `secret` results in the following request params:
```
secret=abc
id=1
id=2
```

Submitting `form-1` results in the following request params:
```
secret=abc
id=1
```

Submitting `form-2` results in the following request params:
```
secret=abc
id=2
```

Let me know what you think and whether I should proceed to writing this up as a test.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1569,7 +1569,7 @@ return (function () {
             }
 
             // include the element itself
-            processInputValue(processed, values, errors, elt);
+            processInputValue(processed, values.element, errors, elt);
 
             // include any explicit includes
             var includes = getClosestAttributeValue(elt, "hx-include");
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1556,12 +1556,16 @@ return (function () {
 
         function getInputValues(elt, verb) {
             var processed = [];
-            var values = {};
+            var values = {
+                form: {},
+                element: {},
+                includes: {},
+            };
             var errors = [];
 
             // for a non-GET include the closest form
             if (verb !== 'get') {
-                processInputValue(processed, values, errors, closest(elt, 'form'));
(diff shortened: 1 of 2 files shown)


bencroker:
Added a test that is a bit long-winded but is passing;
https://github.com/bigskysoftware/htmx/blob/bencroker-patch-2/test/attributes/hx-include.js#L37-L51

1cg:
Thank you ben!

bencroker:
Excellent!

PR #285, opened 2020-12-24
Handle <form novalidate> correctly

whamtet:
Thanks for the great library!  When we have

```
<form novalidate...>
```

The browser does not show validation errors, however htmx still silently validates which is confusing.  This PR will prevent htmx from validating to be consistent with the browser.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1565,7 +1565,9 @@ return (function () {
 
             // for a non-GET include the closest form
             if (verb !== 'get') {
-                processInputValue(processed, values, errors, closest(elt, 'form'), validate);
+                var parentForm = closest(elt, 'form');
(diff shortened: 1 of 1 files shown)


src/htmx.js
```diff
@@ -1559,20 +1561,22 @@ return (function () {
             var values = {};
             var errors = [];
 
+            var validate = matches(elt, 'form') && elt.noValidate !== true;
```
whamtet, line 1564: This is the validation condition.  We only wish to validate when the form itself is explicitly submitted and noValidate is not set.

1cg:
Thank you for this change!  Two changes:

* Can you retarget it to `dev`?
* Can you add some tests?

Thanks!

whamtet:
See https://github.com/bigskysoftware/htmx/pull/288

Issue #286, opened 2020-12-26
Extension: path-deps extended to support explicit path refreshing


jvosloo:
In intercooler.js we have an api method to explicitly trigger a refresh on an element or path: `Intercooler.refresh(eltOrPath)`
I would be useful to extend the htmx `path-deps` extension to support a similar explicit call to refresh dependent paths.

jvosloo:
I will have a go at a pull request for this, using a custom event-based mechanism

1cg:
This has been accepted into `dev` thank you!

PR #287, opened 2020-12-26
Extend path-deps extension to handle explicit path refreshes

jvosloo:
This is to implement #286.

**TLDR**: Went with `PathDeps.refresh("/path/to/refresh");` in the end.
Keeping the rest here for historical purposes.

-----

By catering for a new custom event called `htmx:refreshPath`, we should be able to trigger the refresh of a specific path, e.g. 

- `htmx.trigger(document, "htmx:refreshPath", {path:"path/to/refresh"});`
- `htmx.trigger(document, "htmx:refreshPath", {path:"path/to/refresh/*"});`

I have not added any tests or documentation yet, happy to do so if this idea is deemed to be ok in principle.

**Note**: Since we're interested in triggering updates on all elements with the given path (or dependent paths), the element to trigger on is irrelevant, I just went with "document" in the example. Not sure if there's a better way

### DIFF
--- a/src/ext/path-deps.js
+++ b/src/ext/path-deps.js
@@ -35,6 +35,9 @@
                 }
             } else if (name === "htmx:refreshPath") {
                 refreshPath(evt.path);
+            } else if (name === "htmx:refreshElement") {
+                var elt = evt.detail.elt;
+                htmx.trigger(elt, "path-deps");
             }
         }
     });
--- a/src/ext/path-deps.js
+++ b/src/ext/path-deps.js
@@ -35,7 +35,7 @@
                 }
             } else if (name === "htmx:refreshPath") {
                 refreshPath(evt.path);
-            } else if (name === "htmx:refreshElement") {
+            } else if (name === "htmx:refresh") {
                 var elt = evt.detail.elt;
                 htmx.trigger(elt, "path-deps");
             }
--- a/src/ext/path-deps.js
+++ b/src/ext/path-deps.js
@@ -35,10 +35,7 @@
                 }
             } else if (name === "htmx:refreshPath") {
                 refreshPath(evt.path);
-            } else if (name === "htmx:refresh") {
-                var elt = evt.detail.elt;
-                htmx.trigger(elt, "path-deps");
-            }
+            } 
         }
     });
 })();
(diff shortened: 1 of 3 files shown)


jvosloo:
I had a blond moment and added and then promptly removed another event called `htmx:refreshElement`, but that's a little daft, because if you already know the element you want to refresh, you can just trigger the `path-deps` event on it directly, e.g. `htmx.trigger(htmx.find("#myEl"), "path-deps")`

1cg (commented):
Thank you for addressing this shortcoming compared with intercooler.

The logic looks good.  As far as the API goes, what about surfacing a global symbol, `PathDeps`, with a single method on it:

```js
  PathDeps.refresh("/some/path")
```

That way you don't have to go through htmx to get to the functionality?

jvosloo:
> Thank you for addressing this shortcoming compared with intercooler.
> 
> The logic looks good. As far as the API goes, what about surfacing a global symbol, `PathDeps`, with a single method on it:
> 
> ```js
>   PathDeps.refresh("/some/path")
> ```
> 
> That way you don't have to go through htmx to get to the functionality?

Yours is a far better idea than what I had 🤦‍♂️.  I added the PathDeps api with the refresh method. Does it look ok?

1cg:
Looks reasonable to me!  Can you add tests for it?

1cg:
Pinging again: all this needs are some tests in `test/ext/path-deps.js` and I'm happy to pull

jvosloo:
Sure, no problem. I'm away for a few days with the fam so I will get on it first thing next week 👍

jvosloo:
@1cg I added a couple of tests and updated the documentation to reflect the addition of the API and the `refresh` method.
I couldn't really think of more useful tests to add, since the refresh method reuses the same logic that's already being extensively tested by the existing tests?

1cg:
Looks good, thank you for adding tests.  I don't understand this code though:

```js
    _root.PathDeps = {

        PathDeps.refresh: function(path) {
            refreshPath(path);
        }
    };
```

Why is the property named `PathDeps.refresh` rather than just `refresh`?

1cg:
Ah, looks like it was an accident.  I fixed it and all the tests pass...

jvosloo:
> Ah, looks like it was an accident.  I fixed it and all the tests pass...

Yup, oversight. Don't know how I managed to leave that in there 🤦🏻‍♂️ Thanks for fixing :)

PR #288, opened 2020-12-27
skip validation when novalidate or form submitted indirectly

whamtet:
Hi @1cg here is the previous pull request but pulled into dev with tests as you requested.  Shout out if you have any more queries.


1cg:
thank you!  great change!

whamtet:
Thanks for accepting it!  Looking forward to using HTMX on my next project.

PR #293, opened 2020-12-28
Link to hx-swap for timing (docs only).

guettli:
It took me a wile to find the matching part in the docs how to increase the duration of the settle and swap phase.

I hope this helps other newcomers.


1cg:
Thank you!

Issue #297, opened 2020-12-30
Documentation for htmx:processedNode event


benpate:
There's an undocumented `htmx:processedNode` event.  Is this undocumented for a reason?  If this event is finalized, then let me know and I'll submit a PR with documentation.

benpate:
Just to close the loop from another comment elsewhere.  I'm actually hoping that this event/name is NOT finalized, and it could be updated to `htmx:afterProcessNode` so that it will aligh with the PR listed above. :)

1cg:
The event is used in a few extensions, but I'd be OK with renaming it since it wasn't documented.

PR #298, opened 2020-12-30
Add event htmx:beforeProcessNode

benpate:
This PR triggers a new event before a node is processed by htmx.  It allows extensions to modify nodes in the DOM tree before htmx sees them, enabling higher-order operations.

It is loosely related to #297, and should probably be considered in conjunction with renaming `htmx:processedNode` -> `htmx:beforeProcessNode`.


benpate:
As a demo of what this event enables, here's `hx-boost` rewritten as an extension.  This isn't thoroughly tested, so don't try this with real code, but it works really nicely as a proof-of-concept.  Allowing extensions to modify HTML code *before* it's processed by htmx opens up all kinds of new possibilities.

```javascript
htmx.defineExtension("boost", {
    onEvent: function(name, event) {
        
        // Only look for htmx:beforeProcessNode events
        if (name !== "htmx:beforeProcessNode") {
            return;
        }

        if (event.target == undefined) {
            return;
        }

        var getAttribute = function(node, attr, defaultValue) {
            return node.getAttribute(attr) || node.getAttribute("data-"+attr) || defaultValue;
        }

        var getClosestAttribute = function(node, attr, defaultValue) {
            if (node == undefined) {
                return defaultValue;
            }
            return getAttribute(node, attr, defaultValue) || getClosestAttribute(node.parentElement, attr, defaultValue);
        }

        event.target.querySelectorAll("a").forEach(function(node) {

            // Do not boost links to external sites
            if (document.location.hostname != node.hostname) {
                return;
            }

            // Use href for hx-get
            if (getAttribute(node, "hx-get") == undefined) {
                node.setAttribute("hx-get", node.getAttribute("href"));
            }

            // Allow configuration values for hx-target.  Defaults to <body>
            if (getAttribute(node, "hx-target") == undefined) {
                node.setAttribute("hx-target", getClosestAttribute(node, "boost-target", "body"));
            }

            // Allow configuration values for trigger.  Defaults to primary mouse button click.
            node.setAttribute("hx-trigger", getClosestAttribute(node, "boost-trigger", "click[!button]"));
            node.setAttribute("hx-push-url", "true");
        })
    }
})
```

1cg:
Looks reasonable, please add a test for it

benpate:
I've added tests and documentation.  

Also, I moved the `htmx:beforeProcessNode` event trigger after the code that checks if a node has already been initialized.  This should prevent duplicate calls on events.

Finally, per our discussion in #297, I've renamed `htmx:processedNode` to `htmx:afterProcessNode` in all `/src` and `/src/ext` files.  It seems like these two should go together in a single "transaction."  I didn't touch `/www` or `/src` because it seems like those are handled by some other process.  

Please let me know if this makes sense and works for you, or if there are any other changes/improvements you'd like me to make.

PR #300, opened 2021-01-02
Extension: Preload Updates

benpate:
This (should) complete the preload extension.  I've changed the configuration options so that they fit HTML standards a little better, and have updated tests and documentation.  

If I have missed anything or made any errors, please let me know and I'll fix it ASAP.


1cg:
Awesome, thanks Ben.

One potential idea (maybe a future improvement?): would it be possible to put `preload` on a parent element and get that behavior for all anchros/hx-get's within it?   Is this a good idea or a bad one?  It might imply some logic like we have in `hx-boost` where we examine the URL to ensure it is AJAX safe...

I'm thinking I'd like to use this on the htmx.org website, and it would be nice to simply drop it on the body and be done...

benpate:
I'm updating/simplifying this PR so that preload can be inherited. I like making this explicit, but it's an easy change to accommodate different styles. 

I don't know that we need to worry about internal/external links. Since were only preloading a page (and not executing it) the worst that would happen is that the next page from the third-party website would load faster. 

I think we should make the behavior clear in the documentation, then let people use it as they need. 

I'll try to post this next round of updates shortly.

1cg:
> I don't know that we need to worry about internal/external links.

Well, you can't make ajax calls against external URLs typically.  So you'd need to load them in something like a hidden iframe in that case...

Thank you for continuing to work on this.  I think it's going to be potentially *the* signature extension!

benpate:
I've updated this extension to allow inheritance, and cleaned up the API a bit.  The new documentation page should describe this in enough detail.  Please let me know that this works, or if anyone finds errors/omissions in the code or documentation.

PR #304, opened 2021-01-04
Website: Add preload elements to website navigation

benpate:
- for now, this only affects the top navigation.  More elements can be preloaded in the future if we like how this works.


1cg:
:fire:

PR #330, opened 2021-01-13
Preload Resources

benpate:
Updates to the preload extension that preload linked resources (images only, for now) included in the preloaded document (or HTML fragment).  Scripts and Stylesheets are not affected.  Tested on current versions of Firefox, Chrome, and Safari.


1cg:
do you think it's worth making this an option rather than the default behavior?

like `preload='aggressive'` or something?

benpate:
I've added `preload-images="true"` as an option that is disabled by default, and updated manual tests and documentation to reflect this.

PR #334, opened 2021-01-15
hx-include: Add `closest` and `find`, include descendants of non-form elements

DenizAksimsek:
- The `closest` and `find` features of `hx-target` have been factored out, and added to `hx-include`.
- When using `hx-include`, if we include a non-form element (`shouldInclude` returns false e.g. div) then we look through its descendants for elements which should be included.

_This PR is the same as #333, but with the changes applied to the dev branch._


1cg:
Can you revert the changes to `/www` and then I'll accept?  Thanks!

DenizAksimsek:
> Can you revert the changes to `/www` and then I'll accept? Thanks!

@1cg done!

tomberek (approved):

PR #350, opened 2021-02-05
Handle preserved element that might not be existing

andyhorng:
In some cases that preserved element might not be existing in the document. It might be deleted by JavaScript or the element from server is first time to load into document.


1cg:
Thank you, and thank you for the test!

Issue #323, opened 2021-01-10
path-deps not triggering when element removed


akashchopra:
The path-deps extension is very useful, but I've come across a situation where it doesn't work as expected:

```
<div hx-ext="path-deps" hx-get="/new/content" hx-trigger="path-deps" path-deps="/foo/bar">Some content</div>
<p hx-delete="/foo/bar" hx-swap="outerHTML">Click to delete</p>
```

The `/foo/bar` endpoint returns empty content, so the paragraph is removed when I click on it. However, the `div` above is not updated, despite the path-dep. 

I can make it work by removing the `hx-swap` attribute (so that only the inner HTML is removed), but this workaround is not suitable in my real code.

Is this a bug, or have I misunderstood something about how path-deps or removing elements should work?

jwmullally:
I think I ran into the same issue.

The path-deps trigger works OK for me here:
```html
<div hx-ext="path-deps">
  <div id="outercontainer">
    <button hx-post="/reflect" hx-target="this">Press</button>
  </div>  
  <div hx-get="/someurl" hx-trigger="path-deps" path-deps="/reflect" hx-target="this"></div>
</div>
```

But not here when the outer container is replaced during the POST request
```html
<div hx-ext="path-deps">
  <div id="outercontainer">
    <button hx-post="/reflect" hx-target="#outercontainer">Press</button>
  </div>  
  <div hx-get="/someurl" hx-trigger="path-deps" path-deps="/reflect" hx-target="this"></div>
</div>
```

Legendary software btw, thanks :)

1cg:
Hmm, interesting.  Would one of you be willing to create a test for this case in the `path-deps` test file?

https://github.com/bigskysoftware/htmx/blob/b46e8ab10f1e267bb635ee7afa4fc016229537d7/test/ext/path-deps.js#L160-L159

1cg:
Fixed with https://github.com/bigskysoftware/htmx/commit/506a4474b5cc03248079781bf010aa8e9968748c

Thank you for the test!

PR #352, opened 2021-02-05
Add test for path-deps outerHTML case

jwmullally:
This test reproduces issue #323

Test output:
```
       path-deps replacing containing element fires event:

      AssertionError: expected 'FOO' to equal 'Deps fired!'
      + expected - actual

      -FOO
      +Deps fired!
      
      at Context.<anonymous> (ext/path-deps.js:171:31)

npm ERR! Test failed.  See above for more details.
```

---

For some reason, with this line the test passes, even though this is the similar issue I am seeing. Maybe I am mistaken...

```
var div1 = make('<div id="outer"><button id="buttonSubmit" hx-post="/test" hx-target="#outer" hx-ext="path-deps" >Click Me!</button></div>')
```

### DIFF
--- a/test/ext/path-deps.js
+++ b/test/ext/path-deps.js
@@ -158,4 +158,16 @@ describe("path-deps extension", function() {
         div2.innerHTML.should.equal("Path deps 2 fired!");
     });
 
+    it('path-deps replacing containing element fires event', function () {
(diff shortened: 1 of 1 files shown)


1cg:
merged manually w/ fix in https://github.com/bigskysoftware/htmx/commit/506a4474b5cc03248079781bf010aa8e9968748c

PR #357, opened 2021-02-07
Run tests on GitHub Actions

adamchainz:
For some reason the CircleCI results aren't public - clicking the link from GitHub requires a login. I thought it might be easier and would make test results public if the tests ran on GitHub Actions. But no pressure if you prefer Circle CI!

The latest Chrome is pre-installed in the GHA environment, so there's no need to specially install it, plus tests now run on PR's.


adamchainz:
There's one failure - maybe that's also the reason Circle CI is failing?
```
  1 failing

  1) Core htmx perf Tests
       DOM processing should be fast:
     AssertionError: Should take less than 100ms on most platforms, took: 527.0800000000122ms
      at Context.<anonymous> (core/perf.js:49:14)
```

PR #359, opened 2021-02-07
Improve HX-Trigger documentation

adamchainz:
* Fix URL for move from `X-HX-` prefix to `HX-` - not sure if this would need a redirect.
* Cover all three `HX-Trigger-*` headers in the page.
* Remove some redundant redundancy in the list of response headers


1cg:
thanks!

Issue #363, opened 2021-02-09
hx-swap-oob requires target having an ID


BoPeng:
I have potentially multiple avatar in a page so the avatar is defined as
```
<div class="avatar">
</div>
```
without an `id`.

After the avatar is updated from a `hx-post="update_profile"` somewhere, the returned piece has

```
<div class="avatar" hx-swap-oob="true:.avatar">
</div>
```
which, according to the [doc on `hx-swap-oob`](https://htmx.org/attributes/hx-swap/), would update the first `avatar`, which would be the one on the header and I am ok with it. However, the code triggers

```
DOMException: Failed to execute 'querySelector' on 'Document': '#' is not a valid selector.
```

from what I believe by line 453 of this piece of code

https://github.com/bigskysoftware/htmx/blob/c26e3c10bbcee836ea0ba7b091525737ac6cfd00/src/htmx.js#L441-L454

despite the fact that the `oobElement` is found by class name so no id is required.

https://github.com/bigskysoftware/htmx/blob/c26e3c10bbcee836ea0ba7b091525737ac6cfd00/src/htmx.js#L470-L474

The code also appears to suggest that all matching elements will be replaced (which is even better for me), instead of "If a selector is given, the first element matching that selector will be swapped.".

Is this a bug?

benpate:
[The documentation](https://htmx.org/attributes/hx-swap-oob/) says that only the first matching selector should be swapped.    So, it seems like this is not behaving the way it's supposed to.

However, in practice, I think you're best off having an id for any element that you want to `oob-swap`.  That will make your code much more explicit.

BoPeng:
The document does not mention the target must have an ID, and it does not appear to be necessary as long as the target can be located. I did not study the source code in detail but it does not make much sense to find the elements by provided selector, then again by id.

Whether or not allowing the replacement of multiple targets would be a design decision. It would work fine for my conversation app with multiple avatars but maybe not such a good idea for other applications.

benpate:
Yeah, there's "good, abstract design principles" and then there's "my specific requirements" -- they never seem to line up, do they? :)

Personally, I wish htmx used more IDs than selectors -- the same way that HTML usually uses IDs to target other nodes.  But since we're using selectors, it seems logical to me to follow through all the way and update ALL of the items that match.

@1cg -- How do you feel about allowing `oob-swap` to update multiple elements?  If you had a pull request that implemented it (with tests and documentation, of course) would you consider merging it in?

[...]

PR #367, opened 2021-02-12
Stop re-locating hx-swap-oob oobElement by ID which might not exist

BoPeng:
#363 

This allows `hx-swap-oob` for element without an ID, such as

```
<div class="someclass"  hx-swap-oob="true:.someclass">
</div>
```
The code is also slightly more performant because the original code calls `querySelect(oobElement.id)` to re-locate the `target` in most cases.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -446,7 +446,7 @@ return (function () {
             } else if (oobValue.indexOf(":") > 0) {
                 swapStyle = oobValue.substr(0, oobValue.indexOf(":"));
                 selector  = oobValue.substr(oobValue.indexOf(":") + 1, oobValue.length);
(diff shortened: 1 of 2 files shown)


1cg:
Hello,

Can I ask you to write a few tests around this functionality?

See `test/attributes/hx-swap-oob.js`

Thank you for helping improve htmx!

BoPeng:
Sorry @1cp, I can recognize that the tests are written for mocha but I have never used mocha before . I tried to add two tests but do not know how to run `mocha` against them... Unless you can point me to a step by step guide on how to test htmx, I am afraid that I will have to turn over this PR to you.

BoPeng:
I think I got the logics wrong here. The PR proves again that I need to learn JS more.

PR #371, opened 2021-02-15
Updated to note that change in format and add info about optional prefix

OmenApps:
When connecting to websockets, the previous format using optional prefix was:

`hx-ws="connect ws:/chatroom"`

If has been changed to:

`hx-ws="connect:ws:/chatroom"` or `hx-ws="connect:wss:/chatroom"`

Made this fix in the docs.


1cg:
Thank you!

PR #381, opened 2021-02-22
WebSocket Reconnect

srkunze:
I noticed that htmx does not reconnect after the connection has been dropped for various reasons.

I added this functionality because during development/deployments, server restarts are quite common and page reloads add unnecessary delays to the feedback loop.

This patch also adds the possibility to configure the reconnect interval since different use-cases and stages requires different latencies, e.g. dev: 3sec, prod: 20-30sec (in my case). If you think of a chat-like web app, you might want even have 5-10 secs, whereas message boards/issue trackers could live with up to 60 secs or more.

So, I leave this here open for discussion.

EDIT:

As an addition, one could randomize the reconnect period (reduced server startup load). We then might need an additional config variable such as (wsReconnectMinInterval and wsReconnectMaxInterval). In the wild, a constant should work pretty well for smaller projects. If you think, that the scope of htmx is also for larger projects, we could add it easily if you are okay with the additional config var.

EDIT2:

Just to make you aware of it: https://github.com/joewalnes/reconnecting-websocket (if you prefer pulling a third-party). I would rather go with a simpler solution but especially this field has a lot of opinions on what resembles the perfect solution.

EDIT3:

OK, my fault. I didn't check on existing issues nor pull requests. So to link things up: #148 #150

EDIT4:
A well-described analysis can be found from AWS: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -44,7 +44,8 @@ return (function () {
                 settlingClass:'htmx-settling',
                 swappingClass:'htmx-swapping',
                 allowEval:true,
-                attributesToSettle:["class", "style", "width", "height"]
+                attributesToSettle:["class", "style", "width", "height"],
+                wsReconnectInterval: 10000
             },
             parseInterval:parseInterval,
             _:internalEval,
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1044,7 +1044,7 @@ return (function () {
             for (var i = 0; i < values.length; i++) {
                 var value = values[i].split(/:(.+)/);
                 if (value[0] === "connect") {
-                    processWebSocketSource(elt, value[1], 0);
+                    ensureWebSocket(elt, value[1], 0);
                 }
                 if (value[0] === "send") {
                     processWebSocketSend(elt);
@@ -1052,7 +1052,11 @@ return (function () {
             }
         }
 
-        function processWebSocketSource(elt, wssSource, retryCount) {
+        function ensureWebSocket(elt, wssSource, retryCount) {
+            if (!bodyContains(elt)) {
+                return;  // stop ensuring websocket connection when socket bearing element ceases to exist
+            }
+
             if (wssSource.indexOf("/") == 0) {  // complete absolute paths only
                 var base_part = location.hostname + (location.port ? ':'+location.port: '');
                 if (location.protocol == 'https:') {
@@ -1071,7 +1075,7 @@ return (function () {
                 if ([1006, 1012, 1013].includes(e.code)) {  // Abnormal Closure/Service Restart/Try Again Later
                     var delay = getWebSocketReconnectDelay(retryCount);
                     setTimeout(function() {
-                        processWebSocketSource(elt, wssSource, retryCount+1);  // creates a websocket with a new timeout
+                        ensureWebSocket(elt, wssSource, retryCount+1);  // creates a websocket with a new timeout
                     }, delay);
                 }
             };
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1061,6 +1061,15 @@ return (function () {
                 triggerErrorEvent(elt, "htmx:wsError", {error:e, socket:socket});
                 maybeCloseWebSocketSource(elt);
             };
+
+            socket.onclose = function (e) {
+                if ([1006, 1012, 1013].includes(e.code)) {  // Abnormal Closure/Service Restart/Try Again Later
+                    setTimeout(function() {
+                        processWebSocketSource(elt, wssSource);  // creates a websocket with a new timeout
+                    }, htmx.config.wsReconnectInterval);
+                }
+            };
+
             getInternalData(elt).webSocket = socket;
             socket.addEventListener('message', function (event) {
                 if (maybeCloseWebSocketSource(elt)) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -45,7 +45,7 @@ return (function () {
                 swappingClass:'htmx-swapping',
                 allowEval:true,
                 attributesToSettle:["class", "style", "width", "height"],
-                wsReconnectInterval: 10000
+                wsReconnectDelay: 'full-jitter'
             },
             parseInterval:parseInterval,
             _:internalEval,
@@ -1044,7 +1044,7 @@ return (function () {
             for (var i = 0; i < values.length; i++) {
                 var value = values[i].split(/:(.+)/);
                 if (value[0] === "connect") {
-                    processWebSocketSource(elt, value[1]);
+                    processWebSocketSource(elt, value[1], 0);
                 }
                 if (value[0] === "send") {
                     processWebSocketSend(elt);
@@ -1052,7 +1052,7 @@ return (function () {
             }
         }
 
-        function processWebSocketSource(elt, wssSource) {
+        function processWebSocketSource(elt, wssSource, retryCount) {
             if (wssSource.indexOf("ws:") !== 0 && wssSource.indexOf("wss:") !== 0) {
                 wssSource = "wss:" + wssSource;
             }
@@ -1064,11 +1064,15 @@ return (function () {
 
             socket.onclose = function (e) {
                 if ([1006, 1012, 1013].includes(e.code)) {  // Abnormal Closure/Service Restart/Try Again Later
+                    var delay = getWebSocketReconnectDelay(retryCount);
                     setTimeout(function() {
-                        processWebSocketSource(elt, wssSource);  // creates a websocket with a new timeout
-                    }, htmx.config.wsReconnectInterval);
+                        processWebSocketSource(elt, wssSource, retryCount+1);  // creates a websocket with a new timeout
+                    }, delay);
                 }
             };
+            socket.onopen = function (e) {
+                retryCount = 0;
+            }
 
             getInternalData(elt).webSocket = socket;
             socket.addEventListener('message', function (event) {
@@ -1127,6 +1131,19 @@ return (function () {
             } else {
                 triggerErrorEvent(elt, "htmx:noWebSocketSourceError");
             }
+        }
+
+        function getWebSocketReconnectDelay(retryCount) {
+            var delay = htmx.config.wsReconnectDelay;
+            if (typeof delay === 'function') {
+                return delay(retryCount);
+            }
+            if (delay === 'full-jitter') {
+                var exp = Math.min(retryCount, 6);
+                var maxDelay = 1000 * Math.pow(2, exp);
+                return maxDelay * Math.random();
+            }
+            logError('htmx.config.wsReconnectDelay must either be a function or the string "full-jitter"');
         }
 
         //====================================================================
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1118,8 +1118,8 @@ return (function () {
                 return getInternalData(parent).webSocket != null;
             });
             if (webSocketSourceElt) {
-                var webSocket = getInternalData(webSocketSourceElt).webSocket;
                 elt.addEventListener(getTriggerSpecs(elt)[0].trigger, function (evt) {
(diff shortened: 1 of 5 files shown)


src/htmx.js
```diff
@@ -1060,6 +1061,15 @@ return (function () {
                 triggerErrorEvent(elt, "htmx:wsError", {error:e, socket:socket});
                 maybeCloseWebSocketSource(elt);
             };
+
+            socket.onclose = function (e) {
+                if ([1006, 1012, 1013].includes(e.code)) {  // Abnormal Closure/Service Restart/Try Again Later
+                    setTimeout(function() {
+                        processWebSocketSource(elt, wssSource);  // creates a websocket with a new timeout
```
1cg, line 1068: will this retry on a failure to reconnect?

src/htmx.js
```diff
@@ -44,7 +44,8 @@ return (function () {
                 settlingClass:'htmx-settling',
                 swappingClass:'htmx-swapping',
                 allowEval:true,
-                attributesToSettle:["class", "style", "width", "height"]
+                attributesToSettle:["class", "style", "width", "height"],
+                wsReconnectInterval: 10000
```
1cg, line 48: 10 seconds seems like a long time, but I don't do any websocket programming, so I'm willing to defer to expertise here.  Just want to make sure this wasn't an accidental value.  :)

1cg (commented):
Thank you for helping improve the websocket API!  I made a few small comments asking for clarification, but the pull request looks great.

I especially appreciate the documentation!  :)

src/htmx.js
srkunze: Yes. In this case it will be a abnormal closure (1006). Verified in Chrome and Firefox.

src/htmx.js
srkunze: I actually put some more thoughts into it. Going to push a new version soon.

I though about the use-cases and potential users of htmx. From my perspective, users would just expect the most up-to-date reconnection algorithm or would want to provide their own. So, I removed the number completely and added a more sophisticated full-jitter version instead of a constant. Additionally, I added ability to specify an own implementation.

srkunze: A well-described analysis can be found from AWS: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

src/htmx.js
```diff
@@ -1064,11 +1064,15 @@ return (function () {
...
+                    var delay = getWebSocketReconnectDelay(retryCount);
                     setTimeout(function() {
-                        processWebSocketSource(elt, wssSource);  // creates a websocket with a new timeout
-                    }, htmx.config.wsReconnectInterval);
+                        processWebSocketSource(elt, wssSource, retryCount+1);  // creates a websocket with a new timeout
+                    }, delay);
                 }
             };
+            socket.onopen = function (e) {
+                retryCount = 0;
```
1cg, line 1074: Should retryCount be defined before `socket.onclose =...` so that it's part of the closure?  I can't tell for sure but it looks like it's a global right now...

srkunze: Oh damn! My fault. I reverted another intermediate change and removed the parameter from the respective function. Should be a local variable indeed.

srkunze:
Do you have any further improvement ideas?

1cg:
I don't know websockets that well, so it looks good to me.  I might tweak the algorithm code a little bit to more match my style, but nothing functional.

@benpate would you mind reviewing this and we can accept it in?

benpate:
Howdy!  A while ao, I made another PR (#150) related to web socket retries.  Then I moved on to using SSE on my project instead, and abandoned the original WebSocket effort.

This PR is much better and we should use it as the basis of WebSockets going forward.  I'm not an expert in this area, and #150 was more of a first-pass than a final solution.  In particular, it seems like this one understands the specific error codes that might cause a WebSocket to fail, where mine did not.  Plus, I especially like the option to configure the retry timeout, something that I did not include.

I have two questions about the differences in these two PRs.  For comparison, check out my original changes here: https://github.com/bigskysoftware/htmx/pull/150/files

1) I put a try/catch around the code that opens the WebSocket, and attempted to retry the connection if it fails on opening.  Is this a good idea to consider for this PR?  Or, was there something I don't understand about how WebSockets work on opening?

2) What happens if the original opening element is closed?  With the `maybeRetryWebSocketSource` function, I first checked to see that the element still exists in the DOM before reopening a WebSocket connection.  Is this handled in this PR in some way that I'm missing?

Thanks again for a solid contribution to htmx.  It will make developers much more comfortable using WebSocket connections!

srkunze:
> I don't know websockets that well, so it looks good to me. I might tweak the algorithm code a little bit to more match my style, but nothing functional.

Just let know what it is, so I can change that accordingly; especially useful to know for future pull requests also.

srkunze:
> Plus, I especially like the option to configure the retry timeout, something that I did not include.

That one took really some time to figure it out. Customizability is important because this area is full of opinions/requirements. So, we have a default all-rounder solution which works for almost everyone, and the ability to roll you own (hopefully better suited) variant.

I expect the majority of users not to override it because it is "just reconnecting". Nothing really seriously. The server is either restarting or down for a longer period. So that should be handled gracefully now.

> I have two questions about the differences in these two PRs. For comparison, check out my original changes here: https://github.com/bigskysoftware/htmx/pull/150/files

I had a look. :-)


> 1. I put a try/catch around the code that opens the WebSocket, and attempted to retry the connection if it fails on opening.  Is this a good idea to consider for this PR?  Or, was there something I don't understand about how WebSockets work on opening?

As far as I can see it, the onclose event is always triggered. So, no special handling needed.


> 2. What happens if the original opening element is closed?  With the `maybeRetryWebSocketSource` function, I first checked to see that the element still exists in the DOM before reopening a WebSocket connection.  Is this handled in this PR in some way that I'm missing?

I think you refer to these lines here, right?
``` javascript
            // If the DOM node that owns this websocket no longer exists, then do not try to reopen it.
            if (!bodyContains(elt)) return false;
```

Interesting actually. Some thoughts on this from my side:

1) I think it's separate from this issue here (reconnecting after the socket has been closed unexpectedly)
2) I also think it's a valuable addition as a separate pull request (but maybe made a little bit more sophisticated)
3) It should not just concern reconnecting but always when the element is removed, the respective socket is to be discarded as well.

Especially point 3) stems from the main expectation that also was a driver for this pull request: state-dependent behavior.

element exists => sockets exists
element does not exist => socket does not exist

htmx should (at least so my expecation) maintain this state-dependent behavior as good as possible.

@1cg
Is there already a functionality which closes and removes the socket object from the js interpreter when the corresponding element is removed from the DOM?


> Thanks again for a solid contribution to htmx. It will make developers much more comfortable using WebSocket connections!

You are welcome. I liked working on it as well. :-)

benpate:
Thanks for the detailed analysis.  I agree with your point that we should just ensure there's a robust way of disconnecting WebSockets when the DOM node is removed.  I think there's code in there now to do that, so it might just need a once-over to make sure it's bulletproof.

The specific (corner-case) situation I was concerned with was that 1) a server might disconnect/reboot, causing htmx to start incrementing the retry wait counter, then 2) the originating DOM node might be removed while the WebSocket is still in limbo, then 3) the WebSocket reconnects, but is orphaned from the main DOM.  For this (and possibly other) reason, I think a a simple null check at the top of `processWebSocketSource` would be worth it :)

srkunze:
Alright. I added the check. So, this one here is ready to go. :-)

1cg:
I'm ready to merge this, @benpate you mind doing a final review?

benpate:
Hi :)

Everything *looks* good, but I'm having trouble testing it.  It's likely a problem with me and my tests, and not the code, but I'm still going to ask for some help in making a test that works well.  When I run the sample code below, I'm able to send/receive WebSocket messages.  But, when I disconnect/reconnect my network, the WebSocket doesn't seem to work.  @srkunze - would you mind trying this out to let me know what I'm doing wrong?

MANUAL TEST FILE 
**`/htmx/test/manual/websocket-reconnect.html`**:

```html

<html>
<head>
    <script src="../../src/htmx.js"></script>
    <title>WebSockets Test</title>
</head>
<body hx-ws="connect:wss://echo.websocket.org">

    <form hx-ws="send">
        Send Message to Echo Server...<br>
        <textarea name="message" style="width:500px; height:300px;">&lt;div id="idMessage"&gt;This Is The Message&lt;div&gt;</textarea>
        <br/><input type="submit"/>
    </form>

    <!-- 
    Receive doesn't work with this `echo` server because of differences in the way HTMX formats 
    `send` messages vs. what it expects for replies.  It's has no bearing on the reconnect test.

    <br><hr><br>
    <div id="idMessage"></div>
    -->
</body>
</html>
```

srkunze:
@benpate I added your manual test. Thanks for that.

As the websocket object now is a moving part, we need to retrieve it every time when doing a ws-send. Thanks for spotting that.

1cg:
I have not forgotten about this pull request, things have been crazy over on the hyperscript side.  I will merge it this week, I promise.  Thank you for your contribution!

srkunze:
@1cg You are welcome. :-)

@benpate Also thanks for reviewing. :-)

pohlt:
It might be crucial for the front end to react to a WS connection loss (e.g. show some warning) and to a reconnect (e.g. hide warning or request update from back end), so callbacks for both events can be very useful. I guess this is beyond the scope of HTMX, but for my particular project it would be a strong requirement.

Anyway, thanks for this great library!

benpate:
@pohlt - This is a great suggestion, and should be easy to accomplish following "orthodox htmx."  We could just add a couple of events to the Web Socket code that would signal when a socket connection has been created, errored, or reconnected.  That would allow custom code to respond correctly when those events happen on the browser.

Are you comfortable making that kind of change?  I'd bet it's a straightforward PR, placing [`htmx.trigger()`](https://htmx.org/api/#trigger) calls into the key parts of the code.

srkunze:
> That would allow custom code to respond correctly when those events happen on the browser.

👍

pohlt:
@benpate I'd love to help with a PR, but I'm not a JS pro, barely dipped my toes into HTMX, and spread thin between project deadlines, so don't hold your breath. Could be next month, next year or never. :-D

If I understood you correctly, the idea is to add the callbacks for dis/re/connect of the WebSocket to htmx.js. These callbacks would create events (e.g. `htmx:ws:close`) which could be bound to custom handlers via `htmx.on` by a HTMX user, right?

benpate:
Yes, I think that's exactly it.  You could certainly take a shot at it even if you're not a "Javascript pro."  There's no time, like the present, right?

To clarify one term -- a "callback" usually means registering a function to be called when something happens.  I don't think we need to do that here.  We just need to trigger an event in the browser (using `hx.trigger()`). If there are any handlers listening for that event, they can pick it up from there.  If not, then nothing else has to happen.  Does this make sense?

pohlt:
So something along the lines:
``` javascript
socket.onerror = function (e) {
    triggerErrorEvent(elt, "htmx:wsError", {error:e, socket:socket});
    maybeCloseWebSocketSource(elt);
};
```
but for `onclose` (triggering an event and trying to reconnect) and `onopen` (just triggering an event), right?

I guess by now you could have implemented it yourself instead of holding my hand. :-)

benpate:
Yeah, it should be pretty simple, but you've already got the code editor open, so it's quicker for me to just talk about it than to launch GitHub and all that :)

Thanks for taking a shot at it.  But, you're not alone.  There's  a lot going on in the htmx internals, so just holler if you get stuck and I'll dust off VS Code.

1cg:
@pohlt I highly recommend taking a shot at a pull request since you came up with the idea.  Maybe jump on the discord w/ Ben and me and we can walk you through it.

PR #384, opened 2021-02-23
settle on SSE Swap event

whamtet:
Small PR to make SSE consistent with Websockets.  On htmx.js#L1082 we settleImmediately after oob swapping a websocket event.  For SSE events we also swap in content, but settling is missing which means that the node does not get processed properly.


whamtet:
Thanks!!!

PR #387, opened 2021-02-24
Update examples.md

paxperscientiam:
Updated links to examples


1cg:
I  think you are viewing the docs on github?  They should be viewed here:

https://htmx.org 

where the links are correct...

paxperscientiam:
@1cg ohhhhh. got it

Issue #385, opened 2021-02-23
shouldCancel with <a href="#" hx-get>


BoPeng:
https://github.com/bigskysoftware/htmx/blob/b8ffa5c6a5176ac452e5aeee2710a2456dcde8d1/dist/htmx.js#L922-L926

When I trace an error on my side, I notice that

```
elt.href.indexOf('#') !== 0
```

returns `True` with my `<a href="#" ...>` element because somehow the browser has changed the `href` to `href="http://url_of_this_page/#"`.  I am not sure if this is a browser thing (chrome, mac) or something wrong on my side but I would like to bring this issue to the attention of the developers.

BoPeng:
According to [this stackoverflow question](https://stackoverflow.com/questions/1550901/how-to-get-raw-href-contents-in-javascript), 

```
elt.href
```
gets the "expanded" URL, and
```
elt.getAttribute("href")
```
gets the original value of the URL, which is `#`.

Since 
```
elt.href.indexOf('#') !== 0
```
will always be true, should not
```
elt.getAttribute("href").indexOf('#') == 0
```
be used here?

1cg:
Thank you for contributing this fix!

PR #388, opened 2021-02-25
Use getAttribute("href") instead of .href in shouldCancel

BoPeng:
#385

Using
```
elt.getAttribute('href').indexOf('#') !== 0
```
instead of
```
elf.href.indexOf('#') !== 0
```
seems to be the right thing to do because the latter always returns true. The PR does not fix any bug but it does not break any existing tests either.


1cg:
I merged this change in via a local branch and added a test as well.

Thank you and congratulations on your first contribution to htmx!

PR #397, opened 2021-03-01
Update Modal Example

benpate:
Update modal example to use new wait for <eventName> syntax


1cg:
orthodox hyperscript... I like it

benpate:
And, of course, as soon as I update the demo, hyperscript evolves with an even simpler way of handling this.  I really like the `settle` syntax.  I'll get that into these demos soon, and can hopefully add a few more into the "cookbook" as well.

Issue #398, opened 2021-03-02
hx-post ignores form validation if form is not triggered directly.


BoPeng:
This is essentially the same bug as #396  but expressed in a different way with a concrete example.

Suppose I have a simple flask app that supports `GET` and `POST` of  `test.html` with the following page

```html
<!DOCTYPE html>
<html lang="en">
<body>
  <form action="{{ request.path }}" method='post'>
  <label for="choose">Would you prefer a banana or cherry?</label>
  <input id="choose" name="like" required>
  <button>Submit</button>
  <a href='#' hx-post="{{ request.path }}">hx-post</a>
</form>
</body>
  <script src="https://unpkg.com/htmx.org@1.2.1"></script>
</html>
```

If I do not enter anything and press `submit`, the form is not submitted

![image](https://user-images.githubusercontent.com/9889312/109577450-35803600-7abb-11eb-9e7a-22af4c190c4f.png)

If I enter cherry, the form can be submitted.

However, if I leave the `input` blank and press `hx-post`, the form is submitted regardless of the  `required` property of `<input id="choose" name="like" required>`.

![image](https://user-images.githubusercontent.com/9889312/109578348-a2e09680-7abc-11eb-99c3-17bc77248334.png)

The root cause is that `hx-post` does not trigger the `submit` event of the `form` and bypasses HTML form validation. If `hx-post` is designed NOT to dispatch a `submit` event, there should be a way to easily do that. A hyperscript trick might help.

BoPeng:
It turns out that the reason for no validation is that `hx-post` is placed in `<a>`, not on `<form>`. A PR (#399) is submitted.

simondrabble:
I'm seeing this behaviour with a trigger outside of the form:

```
<form action="/save-things" method="post" id="the-form"><input name="data" type="text" required></form>
<button
    hx-post="/save-things"
    hx-include="#the-form">Save</button>
```

The form will be submitted via hx-post even if the input is empty.

Example here: https://jsfiddle.net/g8akjpn1/

PR #399, opened 2021-03-02
Perform form validation even if the form is submitted by an element within it

BoPeng:
#398 and https://github.com/bigskysoftware/_hyperscript/issues/49

https://github.com/bigskysoftware/htmx/blob/a66458768613ac8193ccad1570ac629a9974d927/src/htmx.js#L1664

means that form is only validated with

```
<form hx-post=>
</form>
```
not
```
<form>
<a hx-post=>
</form>
```
The comment above the code even says so

```
// only validate when form is directly submitted and novalidate is not set
```
but I do not understand why it was designed so.

This PR changes the behavior to validate the form even if the form is submitted by an element within the form.

No test because I am not sure if the original behavior was by design for a reason.


BoPeng:
recreated as PR to dev instead of master.

PR #400, opened 2021-03-02
Perform form validation even if the form is submitted by an element within it

BoPeng:
#398 and https://github.com/bigskysoftware/_hyperscript/issues/49

https://github.com/bigskysoftware/htmx/blob/a66458768613ac8193ccad1570ac629a9974d927/src/htmx.js#L1664

means that form is only validated with

```
<form hx-post=>
</form>
```
not
```
<form>
<a hx-post=>
</form>
```
The comment above the code even says so

```
// only validate when form is directly submitted and novalidate is not set
```
but I do not understand why it was designed so.

This PR changes the behavior to validate the form even if the form is submitted by an element within the form.

No test because I am not sure if the original behavior was by design for a reason.


1cg:
Hi Bo,

I had to roll back this change, as there are cases when you would want an input in a form to submit even if the form was invalid.  Consider an input that is doing server side email validation:

```html
<form>
  <input name="email" hx-post="/validate_email" .../>
  <input name="first_name" required .../>
</form>
```

We would not want to prevent the POST of that input because the user had not filled in a value for `first_name` yet.

I think the right thing here is to give a way to expression validation in the event filter, so you can say something like:

```html
<form>
  <a hx-trigger="click:valid" .../>
</form>
```
That will be a bit more work, but will be more generally useful.

BoPeng:
```html
<form>
  <input name="email" hx-post="/validate_email" .../>
  <input name="first_name" required .../>
</form>
```
@1cg I disagree with your argument because the exact reason of frontend validation is to prevent invalid data from `POST`ing to the backend. 

1. This kind of validation does not change the backend data and should conceptually be performed with `hx-get`, not `hx-post` and the following should work, 

```html
<form>
  <input name="email" hx-get="/validate_email" .../>
  <input name="first_name" required .../>
</form>
```

2. Even if you need to partial `POST`, htmx has already resolved your problem with `hx-params` if `validate_email` only validate `email`. Now, if `/validate_email` does somehow use `first_name`, then `first_name` should not be empty.

```html
<form>
  <input name="email" hx-post="/validate_email" hx-params="email" />
  <input name="first_name" required .../>
</form>
```

3. That is my old argument, `hx-post` should behavior as close to `POST` as possible, namely validating all `fields` instead of just the `form` element.

BoPeng:
Sorry, the first example of using `hx-get` does not readily work but I think the idea should hold. For example, the [`django-ajax-select`](https://github.com/crucialfelix/django-ajax-selects/blob/develop/ajax_select/views.py#L8-L19) package that I use frequently uses `GET` to query matched records from backend.

1cg:
Again, the problem is that you may want to have ajax interactions within the form that do not validate. The common example is an input that does server side validation, such as an email input.  You don't want to prevent that server side validation because other required elements in the form are not filled out and, therefore, fail validation.  

This is an obvious and common use case for htmx, so we can't make form validation the default behavior for elements within a form: it would break existing code in a silent manner.

We are going to have to agree to disagree on the ideal default behavior here.

That said, I understand your use case and have proposed a syntax for it.  But it can't be the default.

Note that you could achieve this default behavior with an extension, however.

BoPeng:
> Again, the problem is that you may want to have ajax interactions within the form that do not validate. The common example is an input that does server side validation, such as an email input. You don't want to prevent that server side validation because other required elements in the form are not filled out and, therefore, fail validation.

What is wrong with the `hx-params="email"` solution for your example? What you basically said was that that there is a case of "server side validation" for the fields that cannot pass "client side validation" which does not make sense to me at all given that you can limit the fields for `hx-post` with `hx-params`.

1cg:
There is nothing wrong it it in the objective sense.  There is no objectively right answer here.  

This is my subjective opinion: a request generated in a form should not validate all elements within that form unless explicitly stating so.

The form itself, as a whole, should validate elements within it, if any.  But there are too many use cases where cancelling a request silently due to validation errors is the wrong behavior:

* Doing server side validation
* A cancel button
* A button that opens more UI from the server side
* etc.

We are going to have to agree to disagree on this particular area of default behavior.  I understand you want it to be one way, and I intend to allow you to do what you want, but I do not want to make that the default.  This is a subjective judgement on my part, and I ask you to respect that.

BoPeng:
@1cg Despite different opinions, we have one common goal, that is making the htmx API intuitive and easy to use. I respect your opinion and I trust that you know way better than me on the use cases. I am going to try the last time and stop. You could hold on your opinion, or involve others if htmx is not a one-man project.

> This is my subjective opinion: a request generated in a form should not validate all elements within that form unless explicitly stating so.

The problem here is that you are including all the data, but not validating them according to user-specified rules.

I proposed validating all the submitted fields, and ask users to limit fields with `hx-params` if they only need some of them. The behavior of `<form><button hx-post>` would be consistent with `<form><button type="submit">`. All the features are already in htmx and my on-line PR works.

Your approach, as I understand, is to taking `valid` as an explicit condition, so that full-form validation is only enabled with
```
<form><button hx-post="/url" hx-trigger="click[valid]"></form>
```
I disliked this idea because taking `valid` as a condition for `hx-post` is logically very strange because `valid` should by default be a prerequisite for `hx-post` like what regular `POST` does, and is more likely what users would expect.

dwasyl:
@1cg was your proposed syntax ever adopted or did this end up in limbo? I'm running into fields `hx-post`ing when they are blank and invalid on the front end which causes some backend issues.

dwasyl:
Great, one thing I just noticed in the hx-validate docs, is that form elements do it by default when being submitted with a form, not when submitted on their own - might be worth noting?

PR #402, opened 2021-03-02
hx-preserve: preserved element should not be swapped if it lies outside of hx-select/hx-target

marc3lsk:
I am not sure if swapping hx-preserve outside of hx-select was intended, but to me it seems like a bug. Hopefully the code and test is self explanatory.


1cg:
Thank you, I merged this manually into the `dev` branch

Issue #413, opened 2021-03-11
Logics of shouldCancel


BoPeng:
https://github.com/bigskysoftware/htmx/blob/2dae703d5576b6c678211fa150b4518b610b900a/src/htmx.js#L937-L941

I am a bit confused about the above code. If we are calling `preventDefault` for element

```
<a href="#">
```
Should not the last logic be 

```
elt.getAttribute('href').indexOf('#') === 0
```
?

PR #414, opened 2021-03-11
Fix logic of shoudCancel

BoPeng:
#413 

I double checked and I think the original PR (#385) fixed the use of`.href` but should have also changed `!==`to `===`. Using the latest version (1.3.1) when I click on `<a href='#'>` the page actually scroll to the top which is a big regression.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -937,7 +937,7 @@ return (function () {
         function shouldCancel(elt) {
             return elt.tagName === "FORM" ||
                 (matches(elt, 'input[type="submit"], button') && closest(elt, 'form') !== null) ||
-                (elt.tagName === "A" && elt.href && elt.getAttribute('href').indexOf('#') === 0);
+                (elt.tagName === "A" && elt.href && elt.getAttribute('href') === '#');
         }
 
         function ignoreBoostedAnchorCtrlClick(elt, evt) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -937,7 +937,7 @@ return (function () {
         function shouldCancel(elt) {
             return elt.tagName === "FORM" ||
                 (matches(elt, 'input[type="submit"], button') && closest(elt, 'form') !== null) ||
-                (elt.tagName === "A" && elt.href && elt.getAttribute('href').indexOf('#') !== 0);
+                (elt.tagName === "A" && elt.href && elt.getAttribute('href').indexOf('#') === 0);
         }
 
         function ignoreBoostedAnchorCtrlClick(elt, evt) {


1cg:
I think the logic is correct: we want to cancel the event if it is going to trigger an ajax request.  

If it starts with a "#", then we don't want to cancel it.

Is my logic backwards here?

BoPeng:
Ohmm, what is the intended behavior of `href='$'`?

```
<a href="#" hx-get="/something_else.html" hx-target="#dest">load me</a>
```

In htmx 2.x,  `shouldCancel` returns true because `elt.href.indexOf('#') !== 0` is always true due to url expansion. In htmx 3.x, `shouldCancel` returns false because `elt.getAttribute('href').indexOf('#') !== 0` is `false`. The consequence is that `< href="#">` is passed on and the browser will scroll to the top. The behavior of `htmx 2.x` is much nice so I was calling this a regression.

1cg:
OK, looking at the logic, I wonder if we should just ignore if the thing starts with a `#` at all.

I believe that logic is there to allow relative anchor links to work, but we handle that situation here:

https://github.com/bigskysoftware/htmx/blob/3352d5c3e95dad01470e8fc725c61dee1d476593/src/htmx.js#L920

I believe we can just drop the whole clause and swallow events for all anchors that have had handlers added.  Does this sound right?

BoPeng:
Technically speaking you are right that `href="#whatever"` should be passed on to the browser after `hx-get` is done.

However, `<a href="#">` will cause the browser to scroll to the top and is normally not a desired behavior. The reason why it is widely used is because we need to 

1. make `<a href="#">` a link.
2. the link usually has a `onclick` that stops the scroll to the top behavior.

Now that our `<a href="#">` does not have `onclick` but has a `hx-get` instead, it makes sense to call `preventDefault` for `<a href="#">`.

BoPeng:
I revised the PR to `preventDefault` only for the case of `href="#"`, not for the case of `href="#anchor"`. Without this PR, pages that use `<a href='#' hx-get="/url">` will scroll to the top after `hx-get`, which is really annoying.

bioworkflows:
The example I gave in #418 can be adapted to show this regression bug

This example can also be used to demonstrate the problem with #385 

That it to say, with htmx 1.2.1 as follows, the "swap" is done in place. With htmx 1.3.1, clicking the add to cart will scroll the page to the first paragraph.

```html
<!DOCTYPE html>
<html lang="en">
<body id='body'>
    <p style="height:1800px;">this is fist paragraph</p>
    <a href="#" hx-get="/add_to_cart" hx-swap='outerHTML'>
    add to cart
    </a>

</form>
</body>
  <script src="https://unpkg.com/htmx.org@1.2.1"></script>
</html>
```

1cg:
Great fix, thank you @BoPeng

PR #415, opened 2021-03-12
Minor description addition for triggering custom events with body modifier

scsmash3r:


1cg:
I merged this into dev and moved it out to its own section.  It's important enough to warrant that.  Thank you!

PR #435, opened 2021-03-25
Add file inputs to hx-ws=send either inline or out of band

frankier:
This patch adds files to websocket sends. By default it does the more efficient thing, sending each file afterwards in a separate message and adding pointer to this. This is more efficient since it relies on primitives inside the browser. The other encoding possibility requires hx-encoding to be set to multipart/json-files-inline. In this case, the files are included inline either as text or base64 depending on the mimetype.

How does this approach seem? If it suits I will add short additional documentation to the effect of the above.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -271,6 +271,17 @@ return (function () {
             }
         }
 
+        function blobToBase64(blob) {
+            var reader = new FileReader();
+            var result = new Promise(resolve => {
+                reader.onloadend = () => {
+                    resolve(result.substr(result.indexOf(',') + 1));
+                };
+            });
+            reader.readAsDataURL(blob);
+            return result;
+        };
+
         //==========================================================================================
         // public API
         //==========================================================================================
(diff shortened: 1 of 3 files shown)


1cg:
Thank you!  Can you make the pull against `dev` so I can review it more easily?

frankier:
Done

1cg:
sorry this has taken so long to merge

can you redo it with just the changes in `src/htmx.js`

1cg:
bumping: 

> can you redo it with just the changes in src/htmx.js

frankier:
Rebased, changed to only modify htmx.js, and added documentation

frankier:
Bump

1cg:
Hi There,

I am taking a break from code for a few weeks, will review this when I am back at it, probably end of next week.

Cheers,
Carson

/// big sky software ///

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

frankier:
Please let me know if there's anything I can do to make this one easier to review/merge or if you have any concerns.

1cg:
Hi There,

Sorry to keep pushing this off.  What concerns me about the change is that it is so large, for something that isn't useful to the majority of htmx users.  Do you think it would be reasonable to introduce an extension cut-point here so this could be moved out to an officially-supported extension for htmx?

Again, sorry, have a lot going on and this change is one I want to get right.  Very much appreciate your work on it.

frankier:
It's something I could take a look at. I guess some extension hooks would still have to be added which I don't understand yet.

What I would like to first counter with is that for developer convenience, which is kind of the point of htmx, there is something to be said for "batteries included". So the thing to consider is as you say that P(submits files over websockets|uses htmx) is quite low. W about P(uses websockets|uses htmx) and P(submits files over websockets|uses websockets). Would it not make sense to put all of the websockets and SSE parts of htmx in an extension together with this patch?

1cg:
poll related:

https://twitter.com/htmx_org/status/1436364749055447061

looks like pulling websockets and sse out is the favorite to win.

interested in taking over the websocket extension?  :)

frankier:
Afraid not, but I can rebase this against any such extension as it materializes.

1cg:
Hi @frankier @benpate has pulled web sockets out to an extension, and it is now in `dev`.  

Any interest in updating your pull request against that code?  Thank yoi!

1cg:
@benpate let's have a chat today about potentially integrating this into the new websocket stuff.

1cg:
@Renerick this is an older web socket change that languished.  Worth taking a look at for the new extension?

Renerick:
I've looked through the diff and added some comments about the implementation. Overall, I think that would be a welcomed addition to htmx websockets support. 

To accept this feature, it would have to be rebased onto latest `dev` and moved to `src/ext/ws.js` (since old websockets stuff is deprecated and going to be removed) 

@frankier , if you are still interested in moving forward with this PR, I will be happy to review and approve it.

Alternatively, I can do the integration myself in a new PR in the following weeks

Renerick (changes requested):

Renerick:
So...
I've given it some thought, and I believe that by implementing this feature in the WS extension, we will provide some **really** specific and opinionated solution for a rather uncommon use-case. 

Instead, I propose to implement a more robust event system for websockets. A counterpart of 'htmx:configRequest', in particular, could be used to extract file inputs from the form and serialize and/or send them in any way desired.

frankier:
Hi. Sorry for the radio silence. I think this seems like the correct way to me. If such a plugin mechanism had been available I probably would have just used it and not attempted to submit any kind of PR. At most, a simplified version of my pull request could be a cookbook example or something in the docs.

Renerick:
Just a head up, I opened an issue to track websocket events implementation, as this has been a long standing issue and I feel this should be finally dealt with. It's hard to give any estimation right now, but I will try to take on with the highest priority

Renerick:
FYI, I just finished #1126 , that adds events support to the WebSocket extension. Check out `htmx:wsConfigSend`. This event allows you to configure outgoing message data, as well as provide your own serialization mechanism by setting `messageBody` property. You can also send multiple messages now, since events receive websocket wrapper object, so sending files in separate message should be simple.

Your feedback on this solution is much appreciated. Thank you for raising this issue and your contribution!

1cg:
Closing this pull request as it has been superseded by https://github.com/bigskysoftware/htmx/pull/1126

Thank you @frankier for your inspiration on this!

PR #443, opened 2021-04-14
Fix the <title> of markdown docs

angusholder:
I noticed all the markdown doc files declare their own page title, but on the live website all have the same title of "</> htmx - high power tools for html". It appears the title var wasn't getting used in the base layout template


angusholder:
Nice, it's there now:

Before:
https://htmx.org/attributes/hx-get/

After:
https://deploy-preview-443--htmx.netlify.app/attributes/hx-get/

PR #444, opened 2021-04-14
Allow explicitly opting-out from an inherited hx-select

### LINKED ISSUE #442: How to have a child element opt out of a parent's `hx-select`
I was hoping I could just write `hx-select=""` on the child to have it opt-out of `hx-select` functionality and use the full result of its `hx-get`, but it seems that `maybeSelectFromResponse` only stops searching when the value of `hx-select` is truthy, rather than non-null, so `hx-select=""` gets ignored entirely.

(Rigjt now I'm working around this by setting `hx-select="#id-of-root-element-of-response"` on the child element, basically turning `hx-select` into a no-op.)

Currently what happens is `getClosestMatch` only stops traversing for when it gets truthy value: https://github.com/bigskysoftware/htmx/blob/6f8899f6fddf0bcf4770eb5387597e9afb578cb5/src/htmx.js#L109-L111

it's called by `getClosestAttributeValue`, which passes `getClosestMatch` the attribute value itself (instead of say a bool): https://github.com/bigskysoftware/htmx/blob/6f8899f6fddf0bcf4770eb5387597e9afb578cb5/src/htmx.js#L119-L125

which is called by `maybeSelectFromResponse`: https://github.com/bigskysoftware/htmx/blob/6f8899f6fddf0bcf4770eb5387597e9afb578cb5/src/htmx.js#L635-L637

I think this could be solved by changing `getClosestAttributeValue` to stop traversing when it reaches a non-null attribute value:
```diff
 function getClosestAttributeValue(elt, attributeName) { 
     var closestAttr = null; 
     getClosestMatch(elt, function (e) { 
-        return closestAttr = getAttributeValue(e, attributeName); 
+        closestAttr = getAttributeValue(e, attributeName); 
+        return closestAttr != null;
     }); 
     return closestAttr; 
 } 
```

Does this sound like a reasonable change/am I going about my issue wrong?

angusholder:
Fixes #442

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -117,11 +117,10 @@ return (function () {
         }
 
         function getClosestAttributeValue(elt, attributeName) {
-            var closestAttr = null;
-            getClosestMatch(elt, function (e) {
-                return closestAttr = getAttributeValue(e, attributeName);
+            elt = getClosestMatch(elt, function (e) {
+                return hasAttribute(e, attributeName);
             });
-            return closestAttr;
+            return elt != null ? getAttributeValue(elt, attributeName) : null;
         }
 
         function matches(elt, selector) {
--- a/test/attributes/hx-select.js
+++ b/test/attributes/hx-select.js
@@ -37,4 +37,24 @@ describe("BOOTSTRAP - htmx AJAX Tests", function(){
         this.server.respond();
         div.innerHTML.should.equal("<div id=\"d1\">foo</div>");
     });
+
+    it("inherits the parent's hx-select", function()
+    {
+        this.server.respondWith("GET", "/test", '<section><p id="d2"></p></section>');
+        var parent = make('<div id="parent" hx-select="#d2"><div id="child" hx-get="/test"></div></div>');
+        var child = parent.querySelector('#child');
+        child.click();
+        this.server.respond();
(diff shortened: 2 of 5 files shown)


1cg:
@angusholder thoughts on :point_up:

angusholder:
I think you're right, it's not very clear, although I think `none` sounds confusing too - what about borrowing from the [CSS keywords for controlling inheritance](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_and_inheritance#controlling_inheritance) and going with `hx-select="unset"`?

1cg:
this is why I ask smart people questions

unset is perfect

want to update the PR?

angusholder:
Okay nice! I'll go update the PR another day

1cg:
still very interested in this idea

1cg:
hi there, still interested in pursuing this w/ the `unset` modifier...  lmk if you have time for it...

1cg:
ping...

1cg:
the `unset` option has been implemented in htmx

bblanchon:
@1cg, please add it to [the documentation](https://htmx.org/attributes/hx-select/).

angusholder:
> the `unset` option has been implemented in htmx

Sorry I didn't end up finishing this, thanks a lot for adding it!

PR #454, opened 2021-04-18
Ignore metaKey for boosted links.

mvaled:
metaKey is the Cmd key in MacOS, which is used for the same purpose of opening
a link in another tab.

So it should be also used to ignore clicks on boosted links.


mvaled:
Closed to be merged into `dev` in #455

PR #455, opened 2021-04-19
Ignore meta key for boosted links.

mvaled:
metaKey is the Cmd key in MacOS, which is used for the same purpose of opening
a link in another tab.

So it should be also used to ignore clicks on boosted links.


mvaled:
@1cg Could please you take a look of this?

1cg:
I have just finished up my semester at school and will be returning to htmx development next week.

PR #488, opened 2021-05-16
rename example to "cascading select".

guettli:


1cg:
this change conflicts with the last change, can you please resubmit?

PR #492, opened 2021-05-18
added regression test for image maps

whamtet:


1cg:
thank you for the contribution!  Can you give me a bit of background on what this test is doing?  Is there a related issue for it I can fix?

whamtet:
I've enjoyed using HTMX and I needed it in an image map.  Because image maps are not commonly used nowadays I wrote a test in my own project to check whether it worked.  It did, so I added a similar unit test to HTMX to improve your test coverage.

whamtet:
@1cg fixed merge conflict.

Issue #472, opened 2021-05-01
Have a hx-function attribute that works with hx-trigger


tonyalaribe:
The hx-trigger system is very powerful and expressive, with support for keybindings/shortcuts(which hyperscript doesn't support). I have had multiple usecases where I want to react to a hx-trigger with a javascript or hyperscript function. If this makes sense with you, I'm happy to look into it and make a pull request.

1cg:
Hi @tonyalaribe 

I think for this I would be more inclined to increase the expressivity of hyperscript.  

Back during the intercooler.js days, I had `ic-action`: https://intercoolerjs.org/attributes/ic-action.html

and hyperscript was designed to replace that and remove that functionality, which got very elaborate, from the core of htmx, so htmx could stay focused on extending HTML as a hypertext.

That being said, I'm willing to look at examples with you and think about if there is a feature here that stays within the "extending HTML as a hypertext" philosophical position of htmx.

1cg:
@tonyalaribe I'm going to close this out, but still interested in discussing with you, maybe on discord?  htmx.org/discord

PR #493, opened 2021-05-19
Explore a possible action attribute

tonyalaribe:
This pr is an extension of issue: https://github.com/bigskysoftware/htmx/issues/472

Like I mentioned in the issue, hx-trigger is very expressive and powerful, and just like with HTML, sometimes we just want to trigger a function or action, instead of a HTTP request. This opens a new set of possibilities. I believe this should be simple, and work like the regular HTML onEvent [eg onClick] attributes, and simply execute the attribute value as javascript. 

What this looks like in practice?
```html
<script>
    function doAction(value){
      document.getElementById("btnID").innerHTML = value;
    }
</script>
<button 
              hx-trigger="click[metaKey&&shiftKey] delay:1s" 
              hx-action="doAction(12345)" id="btnID"
>original string</button>
```
This example will replace original string with the numbers passed as argument to the doAction function, when the user clicks the button while holding down CMD+Shift, and after a 1 second delay. This is really non-trivial to implement with HTML and javascript, especially when compared to how easily it was achieved in the example above, and this is also currently not trivial(or I don't know how to), in hyperscript.

This can be combined quite powerfully with hyperscript, especially using an interop function between hyperscript and javascript like `_hyperscript.evaluate()`. 

```html
<button 
        hx-trigger="click[metaKey&&shiftKey] delay:1s" 
        hx-action="_hyperscript.evaluate(put '1234' into #btnID.innerHTML)" id="btnID"
>original string</button>
```
The line above leverages hyperscript to do the same thing as the previous code example. This opens the door for really powerful expressions not possible before.

### DIFF
--- a/src/ext/actions.js
+++ b/src/ext/actions.js
@@ -1,0 +1,35 @@
+(function () {
+  function processAction(elt, action) {}
+
+  function maybeProcessActions(elt) {
+    if (elt.getAttribute) {
+      var action =
+        elt.getAttribute("hx-action") || elt.getAttribute("data-hx-action");
+      var triggers =
+        elt.getAttribute("hx-trigger") || elt.getAttribute("data-hx-trigger");
+      if (action && triggers) {
+        var triggerSpecs = htmx.getTriggerSpecs(elt);
+        var nodeData = htmx.getInternalData(elt);
+        var triggerAction = function(evt){
+          eval(action);
+        }
+        htmx.addEventListenerFn(elt, triggerAction, nodeData, triggerSpecs[0], false)
+      }
+    }
+  }
+
+  htmx.defineExtension("actions", {
+    onEvent: function (name, evt) {
+      if (name === "htmx:afterProcessNode") {
+        var elt = evt.detail.elt;
+        maybeProcessActions(elt);
+        if (elt.querySelectorAll) {
+          var children = elt.querySelectorAll("[hx-action], [data-hx-action]");
+          for (var i = 0; i < children.length; i++) {
+            maybeProcessActions(children[i]);
+          }
+        }
+      }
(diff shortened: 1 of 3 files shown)


tonyalaribe:
> Few ideas:
> Perhaps hx-call to match the term used in hyperscript to call JavaScript?
> hx-hyper to call hyperscript directly?
> Flip it around and add "on htmx" as an event source for hyperscript?

THese are definitely great ideas. I discussed possible solutions with @bigskysoftware on discord. 
hx-call is good, and I don't mind using that instead. For hx-hyper, we resolved instead to support something like hx-script, to specify what scripting language to use. Defaulting to javascript. 

So we could do 
```html
<button 
        hx-trigger="click[metaKey&&shiftKey] delay:1s" 
        hx-action="put '1234' into #btnID.innerHTML"
        hx-script="hs" // for hyperscript. js for javascript, but not necessary since it defaults to js
        id="btnID"
>original string</button>
```
This way,  htmx can be an actual extension to HTML that is not necessarily tied to hyperscript.

wiverson:
Maybe hx-hs for hyperscript?
In Java it's possible to do a reflection call with a method look up by name if you want to include a reference to an external lib that may or may not be present in your code. That way you can have a "soft" dependency. It's clunky if it's a lot of calls but for a single method call it's easy. Not a JS expert but wouldn't it be possible to just check to see if hyperscript is available dynamically and if so invoke it if it's just a call to hyperscript evaluate? That keeps the dependency on hyperscript soft but also makes it easy/short to call w/a single attribute.

1cg:
closing this out, we now have a branch where we are exploring integrating this in the core w/ a `js:` prefix

PR #507, opened 2021-05-25
Update modal-bootstrap.md

benpate:
- Fixes problem with example where modals could not be opened multiple times


1cg:
can you retarget for master, since this is a docs fix?

PR #513, opened 2021-05-27
Consistent quoting

leevigraham:
Updated quotes so they are consistent.


1cg:
i appreciate this, but i worry that it will cause too many merge conficts with other contributors

PR #521, opened 2021-06-05
Allow docs navigation to scroll on shorter screens

onyekaa:
The menu for the docs is rather long, and on Firefox the last few items are cut off. I added an overflow to the list and a max-height to it to scroll on smaller desktop viewports. Also added a margin to the bottom to allow set some space between it and the rest of the content on mobile screens. 

Screenshot 2021-06-05 at 15 25 47

### DIFF
--- a/www/css/site.scss
+++ b/www/css/site.scss
@@ -196,7 +196,6 @@ h4 {
         display: none !important;
     }
     .hamburger {
-        // display: inline-block;
         float: right;
     }
     .nav {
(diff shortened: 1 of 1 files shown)


www/css/site.scss
```diff
@@ -195,7 +196,7 @@ h4 {
         display: none !important;
     }
     .hamburger {
-        display: inline-block;
+        // display: inline-block;
```
tonyalaribe, line 199: Instead of commenting, maybe delete?

onyekaa: Oh that definitely was my intention. Not sure why I commented. 👍🏽

PR #544, opened 2021-07-06
Add typings

paxperscientiam:
This replaces PR #500 

Rather than using jsdoc, i have added a basic type definition file (src/htmx.d.ts). This could later be extended to cover extensions.

For now, this mainly serves to provide autocomplete to IDE users; I have set `compilerOptions.strict` to `false` as otherwise a litany of errors would be generated by typescript. Unless you plan on adopting typescript into your build process, it makes sense to suppress them.

EDIT: Arg, gotta move dependency to devdependency and squash.


package.json
```diff
@@ -41,5 +43,8 @@
     "sinon": "^9.2.4",
     "uglify-js": "^3.13.7",
     "sass": "^1.34.0"
+  },
+  "dependencies": {
+    "typescript": "^4.3.5"
```
1cg, line 48: can this be made a dev dependency?

paxperscientiam: totes. on it.

PR #545, opened 2021-07-06
Add typings

paxperscientiam:
This replaces PR #500

Rather than using jsdoc, i have added a basic type definition file (src/htmx.d.ts). This could later be extended to cover extensions.

For now, this mainly serves to provide autocomplete to IDE users; I have set compilerOptions.strict to false as otherwise a litany of errors would be generated by typescript. Unless you plan on adopting typescript into your build process, it makes sense to suppress them.

EDIT: PR #544 got killed when i squashed cuz i had to (?) delete remote branch before pushing to remote again.

EDIT2: @1cg , made the changes in case you missed this update


paxperscientiam:
@1cg , just realized that we overlooked something. I forgot to add './src/htmx.d.ts' to the files list in package.json.

Want me to do that? Also, perhaps the definitions file should be moved to its own dir? something like "./typings"

qqilihq:
This still seems an open issue, right? NPM shows the TS badge, but the package doesn't contain the typings.

paxperscientiam:
@qqilihq Good catch. Types file is now included in package.json.files
 See https://github.com/bigskysoftware/htmx/pull/658

PR #550, opened 2021-07-11
Introduce hx-allow-error-codes

mushishi78:
Hopefully this change makes sense 😅 I'd like to be able to show messages when requests fail and it seems incorrect to use a success error code. I'd argue this behaviour is more aligned with typical web semantics as when you return a 404 page, it's normally accompanied with the html you'd like to show the user. It's not obvious to me why this behaviour should change when only rendering a portion of the page. I'm also not particularly eager to use the `htmx:responseError` error and write javascript error handling logic, seems to break the preference for locality that motivates this library.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2435,8 +2435,9 @@ return (function () {
             }
 
             var shouldSaveHistory = shouldPush(elt) || pushedUrl;
+            var shouldAllowErrorCodes = getAllowErrorCodes(elt);
 
-            if (xhr.status >= 200 && xhr.status < 400) {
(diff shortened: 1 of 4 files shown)


1cg:
I would prefer to see this reworked to be a config situation (or maybe an event) rather than an attribute.

Does that sound reasonable?

1cg:
I think that this change has been superseded by the following htmx functionality:

https://htmx.org/docs/#modifying_swapping_behavior_with_events

mushishi78:
Great, thanks. Sorry I didn't get round to responding but your solution looks like it can cover my use case.

PR #555, opened 2021-07-18
Encourage users to provide feature requests for the docs.

guettli:
Idea and text was "stolen" from PostgreSQL: https://www.postgresql.org/docs/current/preface.html


1cg:
closing out because we haven't had issues w/ people submitting docs fixes:

PR #562, opened 2021-07-28
search example typo

dzerrenner:
Typo in the description of the search example. The target id and the url didn't match betewwn text and example.


1cg:
thank you!

PR #564, opened 2021-07-31
docs: typo in websockets example

GriceTurrble:
`source` is not present in the example, nor is that an attribute specified in the docs for `hx-ws`.


1cg:
Thank you!

PR #602, opened 2021-09-27
Fix root: option of the intersect event

MamesPalmero:
This PR fixes the root: option of the intersect special event which is currently broken and adds a manual test to demonstrate the issue.


1cg:
thank you!  I also noticed a bug in `querySelectorExt` that I have fixed

PR #603, opened 2021-09-28
Feature support script tags w src

1cg:
hey guys, would you mind taking a look at this change?

it enables htmx to handle script tags that `src` a remote file

this mechanism is a little harder to test, but fixes an annoyance w/ htmx right now


bencroker:
Looks good to me, nice work!!

dz4k (commented):
Guess I'm a bit late...

LGTM

Note: You can handle `script type=module` by using the dynamic import syntax `import(url)`, which can be used from non-module scripts (like htmx). No way to do it synchronously, though.

PR #615, opened 2021-10-11
Prevent errors from being thrown on 204 status messages

benpate:
When the server returns a status code 204 (No Content), the library currently adds an error into the browser's error log.  This little hack prevents this from happening.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2600,7 +2600,7 @@ return (function () {
                 } else {
                     doSwap();
                 }
-            } else {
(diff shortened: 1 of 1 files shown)


1cg:
rejecting not because it isn't correct, but because I fixed it a slightly different way

benpate:
No, that's wonderful.  I'd much rather use your code than mine, because I was no proud of the fix, but didn't want to dig deeper into the logic just for this.

PR #626, opened 2021-10-20
add extension swap-errors

jgirardet:
the `swap-errors` extension allows swapping response when HTTP Error code is returned.
It's nice way to deal with errors server-side in
It works adding the header `HX-SWAP-ERRORS` in the response.


1cg:
@jgirardet do you think that this extension has been obviated by the new support for configuring swap behavior in the `htmx:beforeSwap` event?

jgirardet:
No I don't. This extension allows you to write less javascript where htmx:beforeSwap forces to write more JavaScript.
Since many errors are handled server side, I think it's better for htmx to support the result (with a non 200 status code) out of the box.

src/ext/swap-errors.js
```diff
@@ -0,0 +1,9 @@
+htmx.defineExtension("swap-errors", {
+  onEvent: function (name, evt) {
+    if (name === "htmx:beforeSwap") {
+      if (evt.detail.xhr.getResponseHeader("HX-SWAP-ERRORS") !== null) {
```
1cg, line 4: Only change I would make is to not use all caps, which is more consistent w/ other htmx headers.

1cg (changes requested):

alexpetros:
> This extension allows you to write less javascript where htmx:beforeSwap forces to write more JavaScript.
Since many errors are handled server side, I think it's better for htmx to support the result (with a non 200 status code) out of the box.

I agree with this, though coming up with a conceptual framework that allows configuration via attributes (instead of JS) is difficult!

We're working with this concept in #1620, so I'm going to close this PR in favor of that.

Issue #619, opened 2021-10-14
Discussion for WebSockets extension syntax


benpate:
We're moving WebSockets support into an extension ([so it was decreed](https://twitter.com/htmx_org/status/1448416208244461573) by @1cg).  This thread is to collect suggestions/feedback on the syntax that we use in the new code.  I'm going to start on this after I work out SSEs (#618), so please let me know if I'm missing anything, or if there's something we should add.  Now is the time :)

I would like to keep it as close to the existing syntax as possible, but there are some funny things with the existing code, so now might be our chance to normalize it a little.  

**Questions:**
1) Do we want to keep the swap-oob syntax for received messages?  Or is there another way to target DIVs that might be better?  This would be cool, but I don't think the WebSocket protocol gives us anything standard to work with.  Would love to get input from someone using the current hx-ws tag.

2) Is there a better / more convenient way to send messages?  I'm going to propose multiple encoding options for this data, including JSON (the current encoding), URL encoded, and multipart FormData.

**Here's my first pass:**

```html
<div hx-ext="websocket" ws-connect="/my-websocket-address">
    <form ws-send="json | url | form ">
        <input name="message"/>
    </form>
</div>

```

benpate:
I'm posting progress on this extension to https://github.com/benpate/htmx/tree/pr-websocket-extension.  It's not at all ready to use, but it's starting to take shape, in case anyone wants to follow along or add comments/suggestions.

PR #641, opened 2021-11-02
hx-ws-send-on-open & htmx:wsOnOpen

goksan:
This adds `hx-ws-send-on-open`, my personal use case is making a `send` to the websocket which subscribes to an event once the websocket is open.
I've also included `htmx:wsOnOpen` which may cater to different preferences and use cases outside of my own.

Happy to add accompanying documentation - just wanted to see whether I've missed an existing way to do this stuff & whether it's something that belongs in the lib.

Thanks for the great work on this project.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1177,6 +1177,11 @@ return (function () {
             };
             socket.onopen = function (e) {
                 retryCount = 0;
+                var init = getAttributeValue(elt, "hx-ws-send-on-open")
+                if (init !== null) {
+                    socket.send(init)
+                }
+                triggerEvent(elt, "htmx:wsOnOpen", e)
             }
 
             getInternalData(elt).webSocket = socket;


1cg:
We (well, @benpate) are in the process of pulling web sockets out as an extension.  Would you mind waiting until that work is done for us to review this PR?

benpate:
Hi @goksan -- the Web Socket extension is currently *very close* to being testable.  I'm just working on making a nice test environment for it now.  You can see it here, if you're interested: https://github.com/benpate/htmx/tree/pr-websocket-extension and can track the progress on the ticket #619.  

I'd love to get your input on the new extension.  It is essentially just the original WebSocket code that's been moved into it's own external file.  And, it's not too early to plug it into a dev environment to see how it performs. (Nothing anywhere near production, though.  Not yet)

1cg:
the changes are now in dev for review

goksan:
Thanks for the heads up.

Untimely response from me, but would be happy to drop my changes into the extension if there's any demand. Will close this assuming otherwise.

PR #643, opened 2021-11-03
swap into all elements matching the out-of-band css selector

### LINKED ISSUE #634: discussion / feature request: target every element matching selector in `hx-select-oob`
Hi, I recently started exploring htmx and I could use some advice on a feature I'd like to implement.

I am looking for a way to target multiple page elements with a single `hx-swap-oob` selector.

I have an page with this structure

```
- Item 1 [activate]
  tags
      - A:off
      - B:off
- Item 2 [activate]
  tags
      - C:off
      - B:off
- Item 3 [activate]
  tags
      - D:off
      - B:off
```

when I click `[activate]` on "Item 1" I want to also indicate the associated tags are active on the other items like this:

```
- Item 1 [deactivate]
  tags
      - A:on
      - B:on
- Item 2 [activate]
  tags
      - C:off
      - B:on
- Item 3 [activate]
  tags
      - D:off
      - B:on
```

I hoped I would be able to achieve this with `hx-swap-oob` ie. ship in the response from `items/1/activate`
 the new version of "Item 1" as normal, and in addition to that a new version of all affected tags using the out-of-band swap.

something like:
```
<div>(item 1 content)</div>
<div hx-swap-oob="true:.tag-A-cls">(tag A content)</div>
<div hx-swap-oob="true:.tag-B-cls">(tag B content)</div>
```

But when used like this the `hx-swap-oob` only updates the `B` tag on "Item 2" and not on "Item 3".
Indeed, this is [the documented behavior](https://htmx.org/attributes/hx-swap-oob/) (emphasis mine):

> If a selector is given, **the first element matching that selector** will be swapped. If not, the element with an ID matching the new content will be swapped.

Is there any way I can override `hx-swap-oob` to hit all matching targets?

The [docs](https://htmx.org/examples/update-other-content/) also discuss other ways of updating, I went through them and I don't think any of them fit this use case:
1. Solution 1: Expand the Target
   - This would mean reloading all of the items on the list, I have other data linked to each item which isn't exactly fast to recompute, so the whole operation would be slow
2. Solution 2: Out of Band Responses
   - Using a generalized selector which matches multiple elements doesn't work as discussed above
   - I would have to have a unique id on each item-tag, and the server would have to compute a full list possibly affected item-tags and ship it as out-of-band swaps (this is the solution I'm leaning towards the most)
3. Solution 3: Triggering Events
   - Each tag would have an event based trigger eg. `<div hx-get=/tags/B hx-trigger="reload-tag-B from:body">...</div>` and the item activation endpoint would return list of affected tags to be reloaded in the `HX-Trigger` header: `HX-Trigger: {"reload-tag-A": null, "reload-tag-B": null}`
   - After "Item" activation/deactivation, all elements of the related tags would reload themselves by with separate http requests to the server
   - I tried this approach and unfortunately it causes a somewhat unpleasant rolling update of the interface as the reloading responses are coming in
4. Solution 4: Solution 4: Using the Path Dependencies Extension
   - I haven't played with this, but I think it would suffer from all the drawbacks of Solution 3
5. Hyperscript
    - Add a hyperscript `on click` handler to the tag element which will take care of updating its peers based on a css selector
    - I would like to avoid this as it leads to fragmentation of the logic between server/client

For this use case, I think it would be ideal if the out-of-band swap could target all elements matching the css selector.
It would allow me to drive the content of the page from the server without overfetching.

Are there any other approaches that could work for this?
Any advice would be much appreaciated.

msladecek:
resolves #634
Motivation for these change is described in more detail in the referenced issue.

### Summary

The [`hx-swap-oob`](https://htmx.org/attributes/hx-swap-oob/) attribute can use a css selector to find its target in the document.
In the current implementation, it only swaps into *the first* element found matching that selector.
It would be useful to be able to target *every* matching element, for example when the request leads to an update of some value displayed in multiple places in the page (such as nickname, tags, ...).

### Solution

Use `querySelectorAll` instead of plain `querySelector`.
Loop over the results and swap a copy of the supplied fragment into each of them.

### Notes on this MR

I'm not very familiar with js and the structure/conventions of this repo so I'm not sure if I made the right changes in all the right files etc., I'd be happy to fix it up if something is not in order.
Also, if there are any more scenarios that I should write tests for, I'd be happy to contribute.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -512,15 +512,22 @@ return (function () {
                 swapStyle = oobValue;
             }
 
-            var target = getDocument().querySelector(selector);
-            if (target) {
-                var fragment;
-                fragment = getDocument().createDocumentFragment();
-                fragment.appendChild(oobElement); // pulls the child out of the existing fragment
-                if (!isInlineSwap(swapStyle, target)) {
-                    fragment = oobElement; // if this is not an inline swap, we use the content of the node, not the node itself
-                }
-                swap(swapStyle, target, target, fragment, settleInfo);
+            var targets = getDocument().querySelectorAll(selector);
+            if (targets) {
+                forEach(
+                    targets,
+                    function (target) {
+                        var fragment;
+                        var oobElementClone = oobElement.cloneNode(true);
+                        fragment = getDocument().createDocumentFragment();
+                        fragment.appendChild(oobElementClone);
+                        if (!isInlineSwap(swapStyle, target)) {
+                            fragment = oobElementClone; // if this is not an inline swap, we use the content of the node, not the node itself
+                        }
+                        swap(swapStyle, target, target, fragment, settleInfo);
+                    }
+                );
+                oobElement.parentNode.removeChild(oobElement);
             } else {
                 oobElement.parentNode.removeChild(oobElement);
                 triggerErrorEvent(getDocument().body, "htmx:oobErrorNoTarget", {content: oobElement})
--- a/test/attributes/hx-swap-oob.js
+++ b/test/attributes/hx-swap-oob.js
@@ -92,5 +92,31 @@ describe("hx-swap-oob attribute", function () {
         byId("d1").innerHTML.should.equal("Swapped");
     })
 
+    it('swaps into all targets that match the selector (innerHTML)', function () {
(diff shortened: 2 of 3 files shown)


1cg:
looks good, but can you revert changes in `/www/ and `/dist`?

msladecek:
> looks good, but can you revert changes in `/www/ and `/dist`?

Done, I kept the change in docs though, or should that be handled in some different way?

Issue #645, opened 2021-11-05
Feature Request: Add `delete` mode to hx-swap and hx-swap-oob


ajusa:
This mode would just attempt to delete the element that matches, rather than swapping it. Use cases (that I can think of anyway)

1. Removing references to an element after hitting a delete button in a modal. For example , if there is a list of elements that you can browse, after hitting delete one of them it should be possible to just delete the element in that list.
2. Having an updated list of users for a chat application. When a user leaves, we want to delete them from the list rather than generating the entire list and swapping that (could be expensive with 100s of users).

I did take a stab at implementing it in https://github.com/ajusa/htmx/commit/890b0c1834e7bcbab33d45e4968db05790936d3f but ~~wasn't quite able to get it to work~~ it works fine, my local test just wasn't correct. 
PR: #646

1cg:
merged and fixed, thx

PR #646, opened 2021-11-05
Add delete option for hx-swap

ajusa:
See #645 for motivation.

This is my first contribution, so let me know if this is incorrect or missing some other case.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -770,7 +770,7 @@ return (function () {
             return insertNodesBefore(parentElt(target), target.nextSibling, fragment, settleInfo);
         }
 	function swapDelete(target, fragment, settleInfo) {
-	    cleanUpElement(target);
+	        cleanUpElement(target);
             return parentElt(target).removeChild(target);
         }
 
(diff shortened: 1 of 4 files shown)


1cg:
implementation looks good, but now we need tests and docs :)

tests probably in 

test/attributes/hx-swap.js

and 

test/attributes/hx-swap-oob.js

and docs in

www/docs.md

and 

www/attributes/hx-swap.md

1cg:
Can you add tests for this swap style?

ajusa:
Added tests and a quick mention in the docs. It's been a while since I've made a PR to a project of this size, please let me know if there are any additional steps I should take, especially with regards to Git. I've changed the branch to be dev as well and merged that into my fork, so it shouldn't have any conflicts as of now.

ajusa:
Finally fixed the rebase issue I was running into, the git diff should be sensible now. Tests that I added pass as well. Let me know any additional documentation or tests to add. Shouldn't take long now that I know how to do it

1cg:
:crown:

PR #657, opened 2021-11-13
Add successful and failed properties to events

jvosloo:


jvosloo:
@1cg the way I added the 2 response event properties (failed and successful), means that they will be exposed in more of the response lifecycle events, besides just htmx:afterRequest. Is that ok? Unless I missed the pot completely and there's a whole other/better way to do it 😅

1cg:
looks good to me, only request is that you add some tests for it, hmu on discord if you need a hand

jvosloo:
> looks good to me, only request is that you add some tests for it, hmu on discord if you need a hand
@1cg  I added the tests. No idea if they're passing? Do the tests automatically run on commit?
If so, can you have a quick squiz to see if I did the the way you'd like?

PR #659, opened 2021-11-14
Improve Active Search Example

Krolken:
The example did not use a search type input so no abilty to clear input. To do a new search on clear the search trigger was added.

Only thing now is that on clear it shows all results. This is how I use it myself but maybe not the intended use for the example.

### DIFF
--- a/www/examples/active-search.md
+++ b/www/examples/active-search.md
@@ -15,10 +15,10 @@ We start with a search input and an empty table:
     <img src="/img/bars.svg"/> Searching... 
    </span> 
 </h3>
-<input class="form-control" type="text" 
+<input class="form-control" type="search" 
        name="search" placeholder="Begin Typing To Search Users..." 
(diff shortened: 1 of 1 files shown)


pljspahn:
It appears that clearing `<input type="search">` isn't supported by Firefox. 

I have a simple form for a search autocomplete (Django+Wagtail autocomplete feature) but once the hx-target is populated with results, it seems there isn't a way to clear the search and remove the results in hx-target without doing things that feel dirty, like adding my own javascript to add listeners/hide elements/etc.

I also tried `<input type="reset">` but that doesn't clear the form (and I'd really rather not add extra inputs if I don't need them).

I think at the very least, the documentation for the Active Search Example should mention that it doesn't work in Firefox and to also provide a nice and easy workaround. I really don't want to start back down the road of having to maintain a bunch of different browser specific hacks.

PR #663, opened 2021-11-17
Fixed :closest and :find extended css selectors on hx-trigger

robinvandernoord:
In the [the docs](https://htmx.org/attributes/hx-trigger/) about hx-trigger, it is mentioned that you can use extended CSS selectors such as `from:closest <CSS selector>` and `from:find <CSS selector>`. This however leads to a syntax error and does not seem to be implemented in the `getTriggerSpecs` method for 'from'.
I have added this parsing to this section and in our demo it works as described.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -907,7 +907,17 @@ return (function () {
                                     triggerSpec.delay = parseInterval(consumeUntil(tokens, WHITESPACE_OR_COMMA));
                                 } else if (token === "from" && tokens[0] === ":") {
                                     tokens.shift();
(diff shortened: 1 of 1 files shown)


1cg:
looks great, and this bug was *not* fixed

can  you to add tests for the fix?  (If not, I can add them)

robinvandernoord:
Thanks! I don't know anything about mocha testing, so if you could do this that would be great!

1cg:
https://github.com/bigskysoftware/htmx/commit/3ee3739b4b76b45ef17f3fb588071dc5c0b7de3b

PR #668, opened 2021-11-21
adding support for `hx-trigger="restore"`

makasaurus:
Adds `hx-trigger=restore` support for when a user hits the back button. This is distinct from `htmx:historyRestore` because this doesn't require cache to fire and will fire whenever a window `popstate` is encountered.

I was able to add some docs for this but I'm not sure how to best create a test scenario for this with sinon. @1cg Any suggestions?


makasaurus:
@1cg mentioned that this might be best as an extension for now. Should I change this over to an extension or would this need some more discussion?

1cg:
I like this as an extension to start out.  Those are located in `src/ext` and you can see tests in `test/ext`

For testing this, I would probably just trigger the event manually: the history API is a pain to test

PR #688, opened 2021-11-26
Fix main websockets example attribute in docs

ajusa:
Tiny change, it was still using hx rather than ws
@benpate does this look good to you?


benpate:
Yes, thank you. Good catch!

PR #691, opened 2021-11-26
Update info about `config.defaultSettleDelay`

hirasso:
Reflect correct value of [defaultSettleDelay](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L46) in docs.


1cg:
thank you!

Issue #606, opened 2021-09-29
Convert htmx to TypeScript 😍


bencroker:
TypeScript would bring lots of DX benefits to the htmx source code. Besides the obvious strict type checking, we would be able to write code using the latest and greatest ES2020, but continue compiling to ES5 (to support IE11), or any version of ECMAScript really, with a single config setting.

```json
{
  "compilerOptions": {
    "target": "ES5"
  }
}
```

I've started a WIP branch in which I've converted `htmx.js` to `htmx.ts`, changing the minimum amount of code possible, and it is compiling with all tests passing. 

@1cg in theory, we could merge this branch into the `dev` branch before the next minor release, and start converting the source code incrementally.

gnat:
Of course up to @1cg to decide if the benefits are actually worthwhile but personally would vote against moving over to TS:

* Increased barrier for contributions.
* More complex, slower build process.
* The true "futureproof" choice is to continue evolving with the latest ECMAScript standard. (I'd rather wait to have official types support in ECMAScript.)

gnat:
Of course up to @1cg to decide if this switch is actually worthwhile, but will have to say no thanks because:

* Increased barrier for contributions.
* More complex, slower build process.
* Type checking may not even be of tangible benefit for htmx (htmx has a nice test suite).

The ECMAScript spec will probably supersede TypeScript in the near future anyway. This has happened historically with other dead Javascript dialects (example: when classes were included in ES6).

bencroker:
While you may have some good points, I still think that the benefits of moving to a modern codebase (rather than one that resembles 2014, no offence @1cg!) far outweigh the arguments given, especially for the core contributors of htmx.

> Increased barrier for contributions.

That's debatable, as TypeScript is a superset of JavaScript and helps to catch errors early.

> More complex, slower build process.

The build process remains the same: `npm run dist`

> Type checking may not even be of tangible benefit for htmx (htmx has a nice test suite).

Type checking helps to write more robust code and leads to tighter integration with code editors that can then catch errors much more reliably. This doesn't question the value of the test suite at all!

khalwat:
In general, I think any library should be written in TypeScript... and probably everything else, too, for the reasons cited above.

> The ECMAScript spec will probably supersede TypeScript in the near future anyway. This has happened historically with other dead Javascript dialects (example: when classes were included in ES6).

[...]

PR #696, opened 2021-11-29
Fix and extend TypeScript definitions

qqilihq:
I have been recently trying htmx with the TypeScript declarations (thank you for these, @paxperscientiam!)

Unfortunately, I noticed that these typings would expect me to import HTMX as follows:

```typescript
import { htmx } from 'htmx';
htmx.defineExtension('foo', { … }); // will not work
```

… while the library actually provides all the API in the default import, and thus my (already working) codebase rather uses …:

```typescript
import htmx from 'htmx';
htmx.defineExtension('foo', { … }); // works!
```

Triggered by this, I fixed the import to work as actually provided by the library (unless of course I'm overlooking something fundamental), and went the long way to extend the typings with all the functionality from the API documentation, and added tests based on the given examples.

I realize that having auto-generated typings would be much better on the long run (see #606), but until then this could serve as a crutch?

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "files": [
     "LICENSE",
     "README.md",
+    "src/htmx.d.ts",
     "dist/*.js",
     "dist/ext/*.js",
     "dist/*.js.gz"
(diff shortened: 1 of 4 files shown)


1cg:
I'm good w/ this.  @benpate  I think you did some TS work w/ htmx.  Can you review and accept?

PR #720, opened 2021-12-15
Setup CI

gotcha:


gotcha:
@carson Would you agree to set `Build forked pull requests` at CircleCI ?

1cg:
My CI foo is very weak.  I'm fine with whatever you want to set up!

gotcha:
@1cg Your answer above does not help me to be sure that you actually changed the configuration of htmx project at CircleCI.

Does that question make sense or do I need to provide more context ?

PR #727, opened 2021-12-18
www/extensions/path-deps: rm extra punctuation

waldyrious:
Just a tiny fix to avoid having two periods at the end of the sentence.


1cg:
thank you!

PR #750, opened 2022-01-03
Add handleHistorySwap extension function

### LINKED ISSUE #749: History swap bypasses extension handleSwap
I'm using an extension to implement some custom behavior to work around #748, and it works really well for regular link navigation, but unfortunately doesn't work on browser navigation. This is because `loadHistoryFromServer` and `restoreHistory` both directly call `swapInnerHtml` instead of `swap`, bypassing the extension check. It would be great if those two continued to use the regular `swap`, or if there was a separate extension hook for `handleHistory` to avoid breaking any existing extensions.

wbobeirne:
Closes #749.

This adds `handleHistorySwap` as a method for extensions to allow them to control the HTML swapping of browser history behavior. I considered just adding a new `swapType` of `history` and keep using `handleSwap`, but my concern was that older extensions wouldn't necessarily know to check for that and might start behaving incorrectly.

I didn't add any tests since history didn't have any, but I saw there were manual tests for it. If this is an acceptable change, I could add some manual test pages for it too.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -703,15 +703,7 @@ return (function () {
                         try {
                             var newElements = ext.handleSwap(swapStyle, target, fragment, settleInfo);
                             if (newElements) {
-                                if (typeof newElements.length !== 'undefined') {
-                                    // if handleSwap returns an array (like) of elements, we handle them
-                                    for (var j = 0; j < newElements.length; j++) {
-                                        var child = newElements[j];
-                                        if (child.nodeType !== Node.TEXT_NODE && child.nodeType !== Node.COMMENT_NODE) {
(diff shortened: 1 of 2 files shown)


1cg:
Would it be possible to add a manual test for this in `test/manual`?

1cg:
Pinging again on tests and docs on this.  I'm going to cut a 1.7 release in a week or so, so lmk if you want it in for that.

alexpetros:
Closing this PR because it's a year and a half old. Feel free to rebase and re-open if you're still interested.

wbobeirne:
Apologies on letting this go stale. Not currently working on the project I used htmx for so no worries, if someone else finds a need for this they can pick it back up!

PR #753, opened 2022-01-05
hx-disinherit - configure closest attribute inheritance

alfonsrv:
As discussed this PR adds `hx-disinherit` to optionally disable inheritance for a single attribute or all  `hx`-attributes associated with an element.

* `hx-disinherit="hx-select"` disables the auto-inheritance for `hx-select` of an element
* `hx-disinherit="false"` disabled auto-inheritance of all `hx`-attributes of an element

Ideally the logic to check whether or not an attribute is inherited should've been implemented alongside `if (closestAttr !== "unset")`, however that wouldn't reflect the inheritance-setting of the `getClosestMatch`ing element the attribute was pulled from.

Differentiated testing pending. Mainly checking if this is going in the right direction.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -127,10 +127,10 @@ return (function () {
         function getClosestAttributeValue(elt, attributeName) {
             var closestAttr = null;
             getClosestMatch(elt, function (e) {
-                closestAttr = getAttributeValue(e, attributeName);
-                if (closestAttr) {
-                  var inheritAttr = getAttributeValue(e, 'hx-inherit');
-                  if (inheritAttr === 'false' || inheritAttr.includes(attributeName)) {
+                closestAttr = getAttributeValue(e, attributeName)
+                if (closestAttr && elt !== e) {
+                  var inheritAttr = getAttributeValue(e, "hx-inherit");
+                  if (inheritAttr && (inheritAttr === "false" || inheritAttr.includes(attributeName))) {
                       closestAttr = "unset";
                   }
                 }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -110,7 +110,7 @@ return (function () {
             var data = getInternalData(elt);
             if (elt === closestMatch || attributeName === "hx-boost") return
             if (getAttributeValue(closestMatch, "hx-boost") === "true" && data.boosted) return
-            var inheritAttr = getAttributeValue(closestMatch, "hx-inherit");
+            var inheritAttr = getAttributeValue(closestMatch, "hx-disinherit");
             if (inheritAttr && (inheritAttr === "false" || inheritAttr.includes(attributeName))) {
                 return "unset";
             }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -127,7 +127,14 @@ return (function () {
         function getClosestAttributeValue(elt, attributeName) {
             var closestAttr = null;
             getClosestMatch(elt, function (e) {
-                return closestAttr = getAttributeValue(e, attributeName);
+                closestAttr = getAttributeValue(e, attributeName);
+                if (closestAttr) {
+                  var inheritAttr = getAttributeValue(e, 'hx-inherit');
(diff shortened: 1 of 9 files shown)


src/htmx.js
```diff
@@ -127,14 +136,8 @@ return (function () {
...
-                closestAttr = getAttributeValue(e, attributeName)
-                if (closestAttr && elt !== e) {
-                  var inheritAttr = getAttributeValue(e, "hx-inherit");
-                  if (inheritAttr && (inheritAttr === "false" || inheritAttr.includes(attributeName))) {
-                      closestAttr = "unset";
-                  }
-                }
-                return closestAttr;
+                closestAttr = getAttributeValue(e, attributeName);
+                return closestAttr = getAttributeInheritance(elt, e, attributeName) || closestAttr;
```
alfonsrv, line 140: Other than the documentation states, this will allow for inheritance to be disabled by an element that does not necessarily specify the htmx-attribute it disables inheritance for.

```javascript
<div hx-select="#select">
    <div hx-target="#target" hx-inherit="hx-select">
        <!-- Won't have hx-select -->
        <button hx-get="/test/"></button>
    </div>
</div>
```

However this could be the desired behavior. Leaving it up to you. Otherwise we'll just have to wrap `getAttributeInheritance` in `if (closestAttr)`

1cg:
Do I read correctly that

```html
  <button hx-target="#foo" ... hx-inherit="hx-target" ...> ... </button>
```
will exclude hx-target?  

I wonder if we should change the name of the attribute in that case to something like `hx-disable-inheritance`

Or we could do a more elaborate syntax supporting a `+` or `-` prefix.  

If we change the name to `hx-disable-inheritance` I wonder if we should have an `all` or `*` option that disables inheritance of all attributes.

Am I reading this right?

alfonsrv:
Exactly. 

> Do I read correctly that
> 
> ```
>   <button hx-target="#foo" ... hx-inherit="hx-target" ...> ... </button>
> ```
> 
> will exclude hx-target?

Exactly. This could alternatively be renamed to `hx-disable-inheritance` or `hx-exclude-inheritance`. While less verbose, I avoided a tripple-word-attribute however to keep it short – also it seems like 3-word-hx-attributes are reserved to rather advanced use cases, but double-thinking about it I feel like manipulating `inheritance` could indeed be considered a more advanced use-case. So a-ok with me. 👍 


> I wonder if we should change the name of the attribute in that case to something like 
> 
> Or we could do a more elaborate syntax supporting a `+` or `-` prefix.

I don't see any other attributes with that syntax. It feels more like something hyperscript would use – which btw I have difficulties working with htmx since I got used to the natural verbosity of hyperscript.

1cg:
I agree trying to keep it short is probably best.  How about `hx-disinherit` :)

alfonsrv:
Sounds good. It's done. 

Also been using it for a couple of weeks now and the global `hx-boost` aspect while having child nodes with different configuration works very well.

1cg:
will be merging this in the next hour!  thank you for your hard work!

alfonsrv:
Quickly fixed the alphabetical order for the reference and tests.

Sorry for the extra commit. My Yubikey went rogue.

1cg:
hey @alfonsrv I merged your PR before I saw this and made some changes (I cleaned up the way targets worked so it isn't a special case now)

Can you take a look at my pushes on dev and lmk if they look good?

alfonsrv:
I honestly think the docs are not comprehensive / verbose enough. Fully understanding what's possible for a given feature often involves picking from different parts of the documentation. Like reading the attribute's reference page often doesn't give all or too little information of the full picture – so it's necessary to then browse through the examples, hoping something comes up. 

I always thought this was mainly due to documenting "quick and good enough", but not actually intended. tl;dr: I don't think it's beneficial to short down the documentation this drastically.

---

I also tested the code and it seems to be working fine.

PR #759, opened 2022-01-11
add instructions for using `htmx.ajax()` with a callback

arcanemachine:
Added some info about how `htmx.ajax()` returns a Promise, which can be paired with a callback to execute code after the content has been inserted into the DOM.


1cg:
thank you!

PR #769, opened 2022-01-15
Adding swap argument for ajax calls

gone:
Gives ajax calls the ability to pass a swap string to control swapping behavior


1cg:
looks like some auto-added semicolons snuck in that are making the change look more dramatic than it is.  Do you mind removing them?

gone:
Updated!

1cg:
perfection! :pinched_fingers:

Issue #717, opened 2021-12-13
hx-post content-type header


pkel:
I'm new here. Thank you for this wonderful piece of software.

I started to play around with HTMX out of pure interest. On the server side, I use a niche web framework that is a little strict about the Content-Type header on POSTed form data. It expects `application/x-www-form-urlencoded` and only that. HTMX appends `; charset=UTF-8` which causes proper confusion on my backend.

https://github.com/bigskysoftware/htmx/blob/e3c0dadbe7de09fde4df32fb7d730ab8385379b4/src/htmx.js#L2310

I've never seen such a content type on form POST requests and wonder how it makes sense.

#74 mentions that the current content type is invalid and provides some links to external discussions.

gmso:
Hi!

I'm also new here, but hopefully I can add something to the conversation 😃 

Out of curiosity, which backend web framework are you using? I use primarily Django, which seems to answer also with `charset=UTF-8`. That's why I had personally never found this issue.

I also don't know why the `charset` information is added to the `Content-Type` header in this case. Maybe someone from the team can shine a light on the reasoning behind it.

In any case, if you need to change the headers sent by the htmx POST requests, you can listen to the `htmx:configRequest` event, and change the event's `detail.headers['Content-Type']` attribute to the value you need.
```html
<script>
    document.body.addEventListener('htmx:configRequest', (event) => {
      event.detail.headers['Content-Type'] = 'application/x-www-form-urlencoded';
    });
</script>
```

This seems to to the trick for me:
![image](https://user-images.githubusercontent.com/55711215/147496783-f0213868-6c6b-4422-8d8e-2e3f28d3e8c4.png)

On my side, I added the event listener to my base template to have this script execute for all pages. But whatever works best for your templates in your framework should do the trick.

1cg:
I am not sure why I went with this (it's almost certainly copy-pasta from somewhere else) so if someone with expertise can weigh in I am happy to change it.

thestics:
Hey there! Is there a way to adjust the content type not on the document level but on the component level?
Say I want this button to issue `x-www-form-urlencoded` contenttype and that `application/json`
Is there a way to do that?
Thanks a lot!

1cg:
https://htmx.org/attributes/hx-encoding/

MuRo-Cronos:
Hey,
I tried to send Requestbody via POST using HTML-button, eg.
					<button id="executeButton"  
					  		hx-post="/dosomething"
							
[...]

Issue #74, opened 2020-05-31
json-enc extension does not update Content-Type


aimichal:
I'm having trouble using the json-enc extension, because it appears to not update the Content-Type header. It is always `application/x-www-form-urlencoded; charset=UTF-8`, even though I've confirmed with console.log that I'm reaching the code in the extension that is calling `xhr.setRequestHeader("Content-Type", "application/json");`.

After some digging around, I found a kind of plot twist. Turns out, the parameter `charset` isn't allowed for mime type `application/x-www-form-urlencoded`. (For example, see discussion at [here](https://stackoverflow.com/questions/16819502/application-x-www-form-urlencoded-and-charset-utf-8) and [here](https://stackoverflow.com/questions/19694503/ajax-setrequestheader-content-type-application-x-www-form-urlencoded-and-ch).) Accordingly, this code in htmx.js:

```js
if (verb !== 'get') {
    headers['Content-Type'] = 'application/x-www-form-urlencoded; charset=UTF-8';
    ...
```

should become

```js
if (verb !== 'get') {
    headers['Content-Type'] = 'application/x-www-form-urlencoded';
    ...
```

For whatever reason, after this change, the extension `json-enc` does update the `Content-Type` header -- hurray! (I confirmed this in FF 76.0.1 and Chrome 83.0.4103.6 on Mac.)

Unfortunately, the header value is actually appended to:

```
Content-Type: application/x-www-form-urlencoded, application/json
```

And that's because the `setRequestHeader("Content-Type", ...)` is called twice, and behavior for `setRequestHeader` is [documented](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/setRequestHeader) to be:

> If this method is called several times with the same header, the values are merged into one single request header.
>
> Each time you call setRequestHeader() after the first time you call it, the specified text is appended to the end of the existing header's content.

😬 

So... it looks like setting the Content-Type header should be done once for the XMLHttpRequest instance.

One idea is to somehow allow extensions to modify the `headers` map, and call this code once immediately before the xhr.send() call:

```js
// request headers
for (var header in headers) {
    if (headers.hasOwnProperty(header)) {
        if (headers[header] !== null) xhr.setRequestHeader(header, headers[header]);
    }
}
```

What do you think?

chg20:
Fixed in dev: https://github.com/bigskysoftware/htmx/pull/92

Master-Y0da:
[...]

PR #771, opened 2022-01-16
fix content-type header for urlencoded form data

pkel:
This PR changes the content-type header for urlencoded form data
(hx-post) as proposed in #74 and #717.

From #74:
> Turns out, the parameter `charset` isn't allowed for mime type `application/x-www-form-urlencoded`. (For example, see discussion at [here](https://stackoverflow.com/questions/16819502/application-x-www-form-urlencoded-and-charset-utf-8) and [here](https://stackoverflow.com/questions/19694503/ajax-setrequestheader-content-type-application-x-www-form-urlencoded-and-ch).)


1cg:
thank you!

PR #789, opened 2022-02-02
Document required node version

adamchainz:
After trying to set up the project, I found that Node 16 is incompatible with the pinned version of node-sass, and node 15 must be used. This PR documents the requirement - perhaps an upgrade to Node 16 can be done after.


1cg:
absolute :crown:

PR #803, opened 2022-02-06
fix link to hotwire/turbo website

msladecek:


1cg:
I actually ended up just pointing to hotwire.dev (which is what I meant)

thank you though!

PR #806, opened 2022-02-07
Fix focus scroll page jumps using new `htmx config.scrollOnFocus` setting.

### LINKED ISSUE #781: Page jump on focus after swap
Posting this as a follow up to a Discord discussion so it doesn't get lost. 

When htmx focuses an element after swap it will scroll to that element even if the user had already scrolled away from it. One case in point is a long form where validation is performed on input changes. The issue can be reproduced with the example below where you select an option from the list, then scroll down to make other changes or submit, but the page jumps up to the select that was changed to focus it.

```html
<form hx-post="" hx-trigger="submit, change">

        <select id="select">
            <option value="">Select</option>
            <option value="2">2</option>
            <option value="3">3</option>
        </select>
        <p>Scroll down to submit</p>
        <div style="height: 800px;"></div> 

        <button id="button" type="submit">Submit</button>

</form>
```

The problem takes place here:

https://github.com/bigskysoftware/htmx/blob/f1cef92e528871993df1a5569b30bb94c99ec847/src/htmx.js#L2569

The focus method includes an option to prevent scrolling which could be use to avoid such cases.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#focus_with_focusoption

Even though it says it's not supported in Safari, I tested it and it works. On Safari mobile the focus is lost regardless so the issue doesn't happen there.

jreviews:
This fixes #781 . It adds a global `htmx.config.scrollOnFocus` which is disabled by default to avoid page jumps on longer responses if the user has already scroll away from an input before the response is swapped. I am not sure it makes sense to add a new attribute or swap modifier for this type of (undesired) behavior. When using a swap scroll modifier, a user may want to scroll to the top if using pagination, and maintain focus (without auto scrolling) when using filters within the same target area, so multiple scroll modifiers would be needed in this case.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -60,6 +60,7 @@ return (function () {
                 disableSelector: "[hx-disable], [data-hx-disable]",
                 useTemplateFragments: false,
                 scrollBehavior: 'smooth',
+                scrollOnFocus: false,
             },
             parseInterval:parseInterval,
             _:internalEval,
@@ -2640,7 +2641,7 @@ return (function () {
                                     // @ts-ignore
                                     newActiveElt.setSelectionRange(selectionInfo.start, selectionInfo.end);
                                 }
-                                newActiveElt.focus();
+                                newActiveElt.focus({preventScroll: ! htmx.config.scrollOnFocus});
                             }
                         }
 


1cg:
i feel like this would be better as a swap option than a global, since I imagine people will want different behaviors in different placers within an app

how do you feel about that?

jreviews:
Unless it's disabled by default, my preference would be to keep global setting for it because I don't see any use-case where I would want this behavior. Then allow overriding it via a swap modifier. 

Not sure what the modifier should be. It should work together with other modifiers because this one would only be used if the triggering elt is an input. This isn't about preventing the input from being focused after swap. It's about preventing the automatic scroll to that input.

So maybe `focus-scroll:true` would activate the scrolling and `focus-scroll:false` would disable it if it's enabled globally. What do you think?

jreviews:
@1cg I added a new `focus-scroll` swap modifier. 

- It's used only when focusing a previously focused input in the target area to prevent scrolling to that input, while still focusing it.
- It overrides the global `htmx.config.defaultFocusScroll`

src/htmx.js
```diff
@@ -2022,6 +2023,12 @@ return (function () {
                             swapSpec["show"] = showVal;
                             swapSpec["showTarget"] = selectorVal;
                         }
+                        if (modifier.indexOf("focus-scroll:") === 0) {
+                            var focusScrollSpec = modifier.substr("focus-scroll:".length);
+                            var splitSpec = focusScrollSpec.split(":");
```
1cg, line 2028: do we need to split here?  It looks like you'd just have the value `true`

1cg (commented):
one small comment

src/htmx.js
jreviews: The modifier value could also be `false` when the global config is `true`, but you want to change set it as `false` for a particular request using the modifier.

1cg: Right, but the 

```js
var splitSpec = focusScrollSpec.split(":");
```

line, wouldn't `focusScrollSpec` already just be "true" or "false" or whatever?  So splitting it on ":" and taking the first result seems unnecessary?  I may be reading the code wrong though...

PR #810, opened 2022-02-09
SSE Test Server

benpate:
This PR is progress refactoring the SSE extension to use the original calling conventions.  To run, 

```bash
> brew install go
> go run server.go
```


benpate:
One note: All tests are passing except for the "multi-channel" test.  It is breaking when `hx-ext` and `hx-sse="connect:..."` are on different DOM nodes.  This was allowed by the original architecture, and should probably be allowed with the extension, too.

PR #811, opened 2022-02-10
Restore hx-ws and hx-sse tags

benpate:
Per discord discussion, this PR restores the original `hx-ws` and `hx-sse` tags into htmx.  The plan is to leave these legacy tags in place until htmx 2.0, and remove them then.  We will continue all new development on these two features in the existing extensions.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -767,6 +767,12 @@ return (function () {
 
         function cleanUpElement(element) {
             var internalData = getInternalData(element);
+            if (internalData.webSocket) {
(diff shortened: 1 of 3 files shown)


1cg:
you can create a draft PR (I just learned about this) for situation when you don't want it merged but want feedback

benpate:
Thank you.  I will figure out how to do that for next time.  

I think this server is complete.  Or at least, good enough for somebody else to look at now.  Please let me know if there's anything else that this iteration is missing and I'll be happy to make updates :)

PR #818, opened 2022-02-13
Treat target="_self" the same as being empty

mauricioarangoosorio:
HI, just run into this issue and propose not to ignore anchors with the target="_self" attribute.

If I'm writing templates I have control on what attributes I add. But if the code is generated by something else, for example, a text editor on a CMS, I don't have control on such attributes. target='_self' is a standard and just tell us to load the page on the same browser window. Therefore I don't think an anchor with such attribute should be ignored by HTMX.

With Htmx 1.6.1 now I have to now write an additional little routine to scrap the content of target, making sure I only do it on those that are equal to '_self'. So it seems a little clunky.


1cg:
happy to look at merging this, but can you target `dev` instead of `master`?

mauricioarangoosorio:
********
UPDATE
*********
I tested on HTMX 1.7.0 and the issue still persists, anchors with the `target='_self'` attribute are ignored by HTMX.

PR #832, opened 2022-02-18
Draft: Adding support for client side boosted redirects

gone:


gone:
This grew into a slightly bigger feature than I anticipated - adding the json parameters was necessary to be able to test the header without blowing up the whole test page.

Two questions that probably need to be answered before merging: 

1. Is there a better name than boost-redirect?
2. what is the right thing to do for https://github.com/bigskysoftware/htmx/pull/832/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR2801  - just allow the code to error?

gone:
Name updated to hx-location based on discord communication

src/htmx.js
```diff
@@ -2792,6 +2792,22 @@ return (function () {
                 var pushedUrl = xhr.getResponseHeader("HX-Push");
             }
 
+            if (hasHeader(xhr, /HX-Location:/i)) {
+                saveCurrentPageToHistory();
+                var redirectPath = xhr.getResponseHeader("HX-Location");
+                var swapSpec;
+                if (redirectPath.indexOf("{") === 0) {
+                    swapSpec = parseJSON(redirectPath);
```
1cg, line 2800: what if we supported a plain path too, so do this in the case of a JSON value or keep the target, etc. in the case of a plain path string?

gnat:
Just gave this a try. Seems to work OK but possible to have this support status codes 300 to 399 in order to support progressive fallback? 

Returning a status code 200 makes browsers blank screen on re-direct (new "location" header).

`hx-target="body" hx-swap="outerHTML"` is progressive and works with 300 to 399 status codes.

gnat:
Another nice realisation regarding the `hx-target="body" hx-swap="outerHTML" hx-push-url="true"` method..

You do not have to disable `hx-boost` :smile: It works with boosted parents.

Keepin the progressive enhancement real!

PR #840, opened 2022-02-24
EventSource: Better error handling of broken connections

benpate:
If a malformed URL (such as `https://localhost:99999999999/`) is passed to the SSE extension, this would previously cause the extension to throw an exception and break entirely and log a difficult-to-understand error message.  

This update adds error checks around this part of the process, and tries to throw a reasonable error, along with triggering standard htmx error events.

There is a new test in the realtime server to demonstrate this process.


benpate:
@1cg - can you review this before we merge?  I'm getting different results when I use `api.triggerErrorEvent()`, depending on when it's being used.  

For regular connection errors (like an HTTP 404) the event is triggering just fine.  The realtime server demonstrates this well.

But, when this function is called [inside a `catch` statement](https://github.com/bigskysoftware/htmx/pull/840/files#diff-aa48e60a783d218419f5293f84f824d34a48607a15160328a8855fe3b7da5e1fR114) (i.e. when I have a malformed URL) I can see the error event getting triggered, but it does not propagate to the DOM nodes above.  To see this working, change [the connection URL on the error page](https://github.com/bigskysoftware/htmx/pull/840/files#diff-1313164d5c756f89554ffbcda4a06d9aef8c079d24bc1c0a4d4d3d2761c4b8f4R19) to `http://localhost:999999999` (which is not a valid port number) and you'll see what I mean.

1cg:
Hmm.  The only way I can see that happening is if the elt is not in the hierarchy or bubbling is being cancelled for some reason.

It all ends up going through `makeEvent()` which creates events that bubble...

Cheers,
Carson

1cg:
Hmm.  The only way I can see that happening is if the elt is not in the hierarchy or bubbling is being cancelled for some reason.

It all ends up going through `makeEvent()` which creates events that bubble...

Cheers,
Carson

alexpetros:
@benpate is this still an issue and do you want to update it and try to get it merged?

benpate:
Eh, not really.  It was just trying to close up a weird corner case that we found because of some other bug.  Honestly, if you're connecting to port 99999999999, you kinda deserve what you get :)

But since plenty of other things have moved forward since then, it's probably just best to close this one out. Yes?

Issue #800, opened 2022-02-04
Discourage use of unpkg.com


adamchainz:
There are many reasons to avoid CDN's, as summarized [in this post](https://blog.wesleyac.com/posts/why-not-javascript-cdn). I'd like to update the htmx docs to:

1. only recommend unpkg.com for development/testing use
2. always recommend the subresource integrity and crossorigin=anonymous attributes, rather than as a "hey maybe you'd like to use them"
3. provide a direct download link to `htmx.min.js` to make it easy for self-hosting

What do you think?

1cg:
I'm good w/ that.  Please propose a PR.

PR #851, opened 2022-03-01
Update installation instructions

adamchainz:
As per #800.

* Recommend downloading htmx as the main way of installing it.
* Cover npm installation in a bit more detail.
* Note the downsides of unpkg.com and always recommend with the integrity SHA.
* Use `<script defer>` in all `<script>` examples. This tells the browser to defer the parsing and processing of the script until all content is on display, increasing performance. [See HTML spec diagram](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-defer).

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,7 @@ By removing these arbitrary constraints htmx completes HTML as a
 ## quick start
 
 ```html
-  <!-- Load from unpkg -->
-  <script src="https://unpkg.com/htmx.org@1.7.0" ></script>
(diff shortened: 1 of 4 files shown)


gnat:
Apologies, I know people are not going to like a post-merge review but I had no idea these changes were going in.

### We need to be extremely mindful of documentation that first touches new users. New users will leave on a whim based on what they see in the Quick Start.


I strongly feel we need to revert this and do more work on it, if at all. The old docs were fine.

@1cg

PR #812, opened 2022-02-10
Create disable-element.js

alexsemerad:
Extension to disable an element (e.g. a button) for the duration of a request. Can either target the current element with the keyword "this" or any other element by specifying a CSS selector. It works by disabling said element and can be further enhanced in combination with the :disabled CSS pseudo-class to style it. Example:

```
<button hx-post="/example" hx-ext="disable-element" hx-disable-element="this" class="button">Submit</button>

<style>
    button:disabled {
        background-color: red;
    }        
</style>
```

PR #869, opened 2022-04-03
feat(ext): add disable-element extension

David-Guillot:
Credits for the idea and the feature code go to @alexsemerad, the author of [the original PR](https://github.com/bigskysoftware/htmx/pull/812) for this extension. Mine is just adding tests and documentation.

### DIFF
--- a/src/ext/disable-element.js
+++ b/src/ext/disable-element.js
@@ -5,7 +5,7 @@ htmx.defineExtension('disable-element', {
     onEvent: function (name, evt) {
         let elt = evt.detail.elt;
         let target = elt.getAttribute("hx-disable-element");
-        let targetElement = (target == "this") ? elt : document.querySelector(target);
+        let targetElement = (target == "self") ? elt : document.querySelector(target);
 
         if (name === "htmx:beforeRequest" && targetElement) {
             targetElement.disabled = true;
--- a/src/ext/disable-element.js
+++ b/src/ext/disable-element.js
@@ -1,0 +1,16 @@
+"use strict";
+
+// Disable Submit Button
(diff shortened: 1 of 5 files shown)


alexsemerad:
I just had a quick peak and the documentation. Everything is perfect, but there is one small detail. 

In the documentation, specifically in the "_Nominal case: disabling the element triggering the request_", you pass "**self**" to hx-disable-element, although I think it should've been "**this**" to disable the element triggering the event (according to the source code, see below):

```
    let elt = evt.detail.elt;
    let targetElement = (target == "this") ? elt : document.querySelector(target);
```

You have the option to replace "this" by "self" in the source code or change it in the documentation. I have to admit, "self" might indeed be more clear than "this". 

Again, awesome job and thanks!

Renerick:
Great PR, but isn't this feature already implemented in https://htmx.org/extensions/loading-states/ (`data-loading-disable`)?

1cg:
Does seem like a bit of a dupe, but I'm fine with multiple extensions.  jQuery had loads of overlapping extensions, and that's the inspiration here.  @David-Guillot can you address @alexsemerad 's concerns?

David-Guillot:
Thanks @alexsemerad for your attention on the `self` vs `this` matter. I wrote the documentation after taking care of the code, and `self` came to me naturally. So I'll change the code.

And thanks @Renerick for pointing `loading-states`, I didn't see it before taking over this new `disable-element` extension. Since @1cg agrees to have overlapping extensions, I'll go with keeping this PR open, as this new extension is more lean than `loading-states`.

David-Guillot:
Here we go, I just added a fixup commit. Let me know if you're fine with it :wink:

PR #872, opened 2022-04-04
Check for localStorage availability

### LINKED ISSUE #866: Running Safari with "Block all Cookies" breaks htmx
Running Safari with setting "block all cookies" breaks htmx, due to "insecure operations" Safari is blocking.

Unfortunately since JavaScript and htmx basically runs, the default fallback (e.g. submitting a form the regular way) doesn't get executed instead.

safari

One of them seems to be related to `S(localStorage.getItem("htmx-history-cache"))` – tried both en-/disabling the history using `<meta name="htmx-config" content='{"historyEnabled":false, "refreshOnHistoryMiss": false}'>`, but that didn't change the behavior.

David-Guillot:
Fix #866

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -77,6 +77,7 @@ return (function () {
         /** @type {import("./htmx").HtmxInternalApi} */
         var internalAPI = {
             bodyContains: bodyContains,
+            canAccessLocalStorage: canAccessLocalStorage,
             filterValues: filterValues,
             hasAttribute: hasAttribute,
             getAttributeValue: getAttributeValue,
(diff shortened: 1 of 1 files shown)


src/htmx.js
```diff
@@ -402,6 +403,17 @@ return (function () {
             }
         }
 
+        function canAccessLocalStorage() {
+            var test = 'test';
```
1cg, line 407: Maybe something a little more htmx-specific?  `htmx:localStorageTest` or something like that?  I'd hate to accidentally overwrite something someone really wanted.

1cg (changes requested):

src/htmx.js
David-Guillot: Ah you're right :+1: ! Let me fix that right away.

David-Guillot: There you go @1cg , I just added a fixup commit. Thanks for your review :slightly_smiling_face:

1cg:
Ah, can you retarget to `dev` for me?

David-Guillot:
> Ah, can you retarget to `dev` for me?

Done

David-Guillot:
@1cg up on this one, which has been properly rebase against `dev` with the fixup commit fixed-up :wink:

David-Guillot:
@1cg up on this one, which has been properly rebase against `dev` with the fixup commit fixed-up :wink:

PR #899, opened 2022-05-03
Style changes proposal

tunguski:
First of all thank you for this great project!

Second of all, this is just a suggestion. If you see anything valuable in this PR, that's great. But if you don't, just reject it and I won't try to discuss it too much. :)

I'm reading htmx site regularly on my phone (pixel 4a) and site looks as it could be improved. Because of font size, weight and multiple alignment lines, readability on not-too-big screens could be improved.

This PR suggests a couple of changes in style:

* Base `font-size` of `16px` - on small screens current font size results in very small part of the content being visible
* Additional `padding-bottom` of `5em` for body - on some pages content ends up direcly at the end of the window; having small margin at the bottom makes it easier to read last line and visually separate it from the browser/background
* Code `font-weight` `bold` removed, `font-size` changed to `0.9em` - currently on small screen I don't see even short lines fully
* `pre` `margins` removed - this way one line of alignment is removed
* `blockquote` `margin` changed to `1em`, added `padding-left` `0.5em` - currently text touches left border
* content `div` `max-width` changed to `40em` - even more subjective than other changes, but it makes content closer to 80 chars in line
* `p` `margin-left/right` removed - aligns content with headers
* headers `font-weight` changed to `600` - less heavy headers
* `ul` `padding-left` changed to `1em` - removes one additional alignment line which currently is created by dots in lists
* Removed search input in `navbar` - on most screen sizes I see logo is broken into two lines. Removing search helps keeping aesthetic style of the navbar.
* Default background color for `pre` - now code is unreadable if `js` is disabled
* Similar chnges for table style changes - smaller font, cell padding, right alignment to content of page

Outcome:

compare_main_htmx

compare_htmx

Again, feel free to incorporate it in any part you think it's worth it. Including dropping this PR fully!

Thanks!

### DIFF
--- a/www/_includes/layout.njk
+++ b/www/_includes/layout.njk
@@ -55,6 +55,13 @@
                     <div>
                         <a href="https://github.com/sponsors/bigskysoftware?o=esb">sponsor</a>
                     </div>
+                    <div>
+                        <input _="on keyup[key is 'Escape'] or click elsewhere
+                                    blur() then set my.value to ''
+                                  on keyup[key is '/'] from <body/>
+                                    focus()"
+                                placeholder="&#128269;" type="text" id="search"/>
+                    </div>
                 </div>
                 <div class="github-stars">
                     <iframe style="margin:auto;" src="https://ghbtns.com/github-btn.html?user=bigskysoftware&repo=htmx&type=star&count=true" frameborder="0" scrolling="0" width="150" height="20" title="Star htmx on GitHub"></iframe>
--- a/www/_includes/layout.njk
+++ b/www/_includes/layout.njk
@@ -55,13 +55,6 @@
                     <div>
                         <a href="https://github.com/sponsors/bigskysoftware?o=esb">sponsor</a>
                     </div>
-                    <div>
-                        <input _="on keyup[key is 'Escape'] or click elsewhere
-                                    blur() then set my.value to ''
-                                  on keyup[key is '/'] from <body/>
-                                    focus()"
-                                placeholder="&#128269; / to Search Docs" style="width:150px; padding: 6px; border: 1px solid #ccc; border-radius: 4px; margin-right: 6px" type="text" id="search"/>
-                    </div>
                 </div>
                 <div class="github-stars">
                     <iframe style="margin:auto;" src="https://ghbtns.com/github-btn.html?user=bigskysoftware&repo=htmx&type=star&count=true" frameborder="0" scrolling="0" width="150" height="20" title="Star htmx on GitHub"></iframe>
--- a/www/_includes/layout.njk
+++ b/www/_includes/layout.njk
@@ -72,7 +72,7 @@
 </div>
 </div>
 
-<div class="c content">
+<div class="c content {{ customClasses | default("") }}">
     {{ content | safe }}
 </div>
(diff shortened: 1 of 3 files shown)


gone:
I use the search a lot, so I wouldn't want to lose that feature

1cg:
@dz4k you mind reviewing and pulling in the stuff you like?

I think we should keep the hyperscript search shortcut, otherwise I leave it to you Deniz

dz4k:
The search could be shrunk, the way the new MDN does it:
tiny search bar

1cg:
keeping this open, I like the change except for the removal of search, but want @dz4k to approve/integrate it

dz4k:
Consider increasing the max-width of .c on the docs page, since it has a sidebar. Currently, the content is only about 28rem wide.

This part of the reference page seems broken:
image

The talk page too:
image

Otherwise LGTM.

tunguski:
> Consider increasing the max-width of .c on the docs page, since it has a sidebar. Currently, the content is only about 28rem wide.

Yeah, I felt it should be changed, but did not search for solution yet. Is it possible to modify width of single page?

> This part of the reference page seems broken: image

What exactly should be fixed here? Is it about the page width or something else?

> The talk page too: image

Would you consider it if I would make it single column? Despite seing it multiple times, I still have a feeling it's kind of messy. Every header is on slightly different height.

What do you think?

> Otherwise LGTM.

Thanks!

dz4k:
> > Consider increasing the max-width of .c on the docs page, since it has a sidebar. Currently, the content is only about 28rem wide.
> 
> Yeah, I felt it should be changed, but did not search for solution yet. Is it possible to modify width of single page?

You can add a `<style>` tag to docs.md to override it.
 
> What exactly should be fixed here? Is it about the page width or something else?

The two forms of the function are on a single line with no space. I would expect something that looks like

```
htmx.findAll()
htmx.findAll(elt, selector)
```

You can leave this one though.

> Would you consider it if I would make it single column? Despite seing it multiple times, I still have a feeling it's kind of messy. Every header is on slightly different height.

That will be fine for now. The talk page needs cleaning up anyway.

tunguski:
> > > Consider increasing the max-width of .c on the docs page, since it has a sidebar. Currently, the content is only about 28rem wide.
> > 
> > 
> > Yeah, I felt it should be changed, but did not search for solution yet. Is it possible to modify width of single page?
> 
> You can add a `<style>` tag to docs.md to override it.

I wasn't sure where to put it. I've added template param `customClasses` that allows specifying additional page-level classes. I've defined `wide-content` class that overrides page width.

> > What exactly should be fixed here? Is it about the page width or something else?
> 
> The two forms of the function are on a single line with no space. I would expect something that looks like
> 
> ```
> htmx.findAll()
> htmx.findAll(elt, selector)
> ```
> 
> You can leave this one though.

Ok, now I understand. In current layout this issue exists too - these are two inline elements one after another. Only because of accidental sizes of page, cell and font it was split into two lines.

I've introduced additional css style that displays `code` elements in first column of info table as block. It behaves as you described.

> > Would you consider it if I would make it single column? Despite seing it multiple times, I still have a feeling it's kind of messy. Every header is on slightly different height.
> 
> That will be fine for now. The talk page needs cleaning up anyway.

Ok, I'm leaving it as is.

PR #911, opened 2022-05-13
Docs: Hotwire / Turbo to htmx Migration Guide

gnat:
Very short right now but to get the ball rolling on a guide we can add common practices to.

Why?

* To eliminate friction for people migrating from hotwire to htmx.
* So we do not lose people to hotwire just due to a lack of basic common "how to do X" knowledge.


gnat:
Squashed all changes so far.

gnat:
Problem:

Hotwire uses kebab case.
HTMX uses camel case. Most examples are in camel case.
This is a transitional guide, so I explained it like this:

image

Thoughts? @1cg 

Also any additional feedback would be nice. Thanks.

1cg:
htmx fires both camel and kebab style event names:

https://htmx.org/docs/#event_naming

gnat:
Yup, I reference it in the guide, I think it's the best compromise.

PR #863, opened 2022-03-15
Improvements in WebSocket extension

Renerick:
* Implement proper `hx-trigger` support
* Expose trigger handling API to extensions
* Implement safe message sending with sending queue
* Fix `ws-send` attributes connecting in new elements (#859)
* Fix OOB swapping of multiple elements in response (#857)
* ~~Fix `api.shouldCancel` call (#846)~~ Fixed in eafd316d5f436a6dda4b3af4a89d85de20d537ba

Issue #858, opened 2022-03-13
Morphdom extension not working with changes through the Websocket extension


advanderveer:
I'm experimenting with a setup where updates to my UI are received through a websocket connection. These updates may contain arbitrarily complex html so I wanted to make sure to cover may bases and use morphdom for the swapping mechanism. My HTML looks like this: 

```HTML
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3" crossorigin="anonymous">
    <script src="https://unpkg.com/htmx.org@1.7.0"></script>
    <script src="https://unpkg.com/htmx.org@1.7.0/dist/ext/ws.js"></script>
    <script src="https://unpkg.com/morphdom@2.6.1/dist/morphdom-umd.min.js"></script>
    <script src="https://unpkg.com/htmx.org@1.7.0/dist/ext/morphdom-swap.js"></script>
    <script>
   // _NOTE: I copied the morhdom extension into my own script tag so I could debug it. But the behaviour is the same If I 
   // include it through a CDN.

    console.log("loading");
    htmx.defineExtension('morphdom-swap', {
        isInlineSwap: function(swapStyle) {
            console.log("isInlineSwap", morphdom, swapStyle);
            return swapStyle === 'morphdom';
        },
        handleSwap: function (swapStyle, target, fragment) {
            console.log("handleSwap", morphdom)
            if (swapStyle === 'morphdom') {
                if (fragment.nodeType === Node.DOCUMENT_FRAGMENT_NODE) {
                    morphdom(target, fragment.firstElementChild);
                    return [target];
                } else {
                    morphdom(target, fragment.outerHTML);
                    return [target];
                }
            }
        }
    });    
    </script>
  </head>
  <body hx-ext="morphdom-swap" hx-swap="morphdom"> 
    <div hx-ext="ws" ws-connect="/logz/log-ws-trigger">
      <div id="log-container"></div>
    </div>  

    <button hx-get="/test">Click Me!</button>
  </body>
</html>
```

[...]

PR #928, opened 2022-05-28
Update websockets extension documentation

Renerick:
Add mention of message queuing mechanism, following merge of #863 

Expand on the topic of swapping logic (relevant to #858). Thought that it should better describe the proper way to customize swapping mechanism for WS messages


1cg:
Thank you!

PR #930, opened 2022-05-30
Dark theme

tunguski:
Just a proposal. Let me know what do you think.

### DIFF
--- a/www/css/site.scss
+++ b/www/css/site.scss
@@ -425,3 +425,47 @@ table td:first-child code {
   width: 150px;
   border-radius: 4px;
 }
+
+.note {
+  border: 1px solid whitesmoke;
+  background-color: #d0dbee;
+  padding: 8px;
+  border-radius: 8px;
+}
+
+@media (prefers-color-scheme: dark) {
+  body {
+    color: #eee;
+    background-color: #202945;
+  }
+  .top-nav {
+    box-shadow: none;
+  }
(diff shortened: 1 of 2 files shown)


gnat:
I like the idea of supporting `@media (prefers-color-scheme: dark) {` for people who don't have something like Dark Reader installed.

The color selections here are a little over saturated, and the gradient should probably be removed, but the fundamental idea is sound.

1cg:
Sorry I let this sit so long.  I think I'd prefer a black background to blue for a dark mode.  Is that doable?  (my design-fu is awful, of course)

alexpetros:
@tunguski unfortunately, we did a website refactor pretty recently and removed the `scss` files entirely. Are you interested in rebasing this PR against the current website build?

gnat:
image

gnat:
That's [Dark Reader](https://darkreader.org/) extension. I recommend it as an interim solution (..although I imagine most devs desiring lower brightness already are aware :wink:)

First impressions really do count.. so if we're "taking over" this responsibility, it has to look good (dark reader's design is good enough IMO). Everyone who has wanted to contribute a dark mode so far has "programmer art" design sensibilities (sorry OP, but it is a good idea!)



The site will get one eventually but short term I recommend dark reader extension. If it's a worse dark mode than dark reader, we need to avoid it.

PR #937, opened 2022-06-05
Proposal: talk page in single column

tunguski:
Fixes header margins.

Current style sets no margin for first child (https://github.com/bigskysoftware/htmx/commit/b3dbb999#diff-5de0d55baacf311719c2102f912d06ecf3c7c24f023dc6af8f32e2f0a7a9cd41R133).


1cg:
Can I see a screen shot?

tunguski:
## Current style on small screen. 

See the difference in top margins between `Essays` (wrong margin) and `Podcasts` (correct margin).

wrong_margin

## After the change - small screen 

correct_margin

## After the cange - big screen

Two column layout was removed.

changed_margin_full_screen

Issue #425, opened 2021-03-17
Trouble with `hx.process()`, playing around with making an extension


paxperscientiam:
So, I've started some experimenting with making a pagination extension. What I've got so far is really messy, but that's beside the point. Current sticking point is that `htmx.process(element)` seems not work.

My goal is to update the `hx-get` attribute each time there's a page turn event. I figured that once I update the attribute value and then run `htmx.process` on the relevant element, all would be good. Unfortunately, the hx-get never gets updated internally.

Here's my code so far. What might I be doing wrong?

As an aside, is there a deeper explanation of how the extension API should be used? (https://htmx.org/extensions/#reference). Like with `onEvent()`, there's a slew of events, but it's unclear to me what augmentations should be done when. To me, initial DOM changes (creating previous and next buttons and attaching them) should be done when `htmx:beforeProcessNode` is fired.

Anyway, thanks!

```html
<!doctype html>
<html lang="en">
    <head>
        <meta charset="UTF-8"/>
        <title>Document</title>

        <script type="text/javascript" src="src/htmx.js"></script>
        <script type="text/javascript" src="src/ext/my.js"></script>

        <style type="text/css" media="screen">
         header,#body,#hx-pagination-container,footer {
             border: 1px solid black;
             min-height: 200px;
         }
         
        </style>
    </head>
    <body>
        <header>header</header>
        <div id="body"
             hx-get="/body.html"
             hx-trigger="load,click"
             hx-ext="pagination-ext"
             hx-pagination-ext="initialpage:5,resourceRoot:/book1/,filePrefix:page-,fileExt:html"></div>
        </div>
        <footer>footer</footer>
    </body>
    
</html>
```

```javascript
/*
  1. prefetch
  2. option to append at beginning of stack or end of stack
  3. ltr and rtl support

*/
(function(){

    function updateElements(elRoot, pageNumber) {
        elRoot.setAttribute("hx-pagination-current-page", pageNumber);
        let x = document.getElementById("hx-pagination-prev-button");
        let y = document.getElementById("hx-pagination-next-button");

        const parameterString = elRoot.getAttribute("hx-pagination-ext");
        const pairStrings = parameterString.split(",");
        const parameters = {};
        for (const pam of pairStrings) {
            const [k,v] = pam.split(":");
            parameters[k] = v;
        }

        const nextUrl = `${parameters['resourceRoot']}${parameters['filePrefix']}${pageNumber + 1}`;
[...]

PR #943, opened 2022-06-16
Added cleanUpElement to public API

viliammihalik:
https://github.com/bigskysoftware/htmx/issues/425


1cg:
Looking at the referenced issue, maybe this should be on the extension API and the public API should have something higher level like `reinit()`?

PR #980, opened 2022-07-19
Docs: Reference guide readability. Essential attributes.

gnat:
Removed words while improving overall descriptiveness, consistency.

Added Essential Attributes, to ensure important attributes are not lost in the supplementary ones. Important for newcomers to htmx.


1cg:
I like it.  Can we change the terminology to "Core" and "Additional" attributes?

PR #984, opened 2022-07-23
Added reinit method to public API

rudiv:
In reference to #425 (and subsequent PRs #943) this adds a `reinit` method to clean up elements and then reprocess.

`cleanUpElement` was extended to have a parameter that resets the `initialized` flag too (to allow `initNode` to reactivate).

Documentation updated with an Alpine sample that would reinitialise when a `hx-get` has changed on an element.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -822,7 +822,7 @@ return (function () {
             }
         }
 
-        function cleanUpElement(element) {
+        function cleanUpElement(element, withInitialized) {
             var internalData = getInternalData(element);
             if (internalData.webSocket) {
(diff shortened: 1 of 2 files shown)


1cg:
As of 1.8.1, `htmx.process()` now will inspect a node and, if any attributes have changed, it will re-process them.

See https://github.com/bigskysoftware/htmx/commit/1dbb22b682c625c5054de7127da975f30dfbe6ea

So you can simply modify a node and call `htmx.process()` on it.  This should fix issues with morphing, etc. and so I believe makes this PR obsolete.

PR #1003, opened 2022-08-11
Update docs information architecture + tweak the look & feel

xdrdak:
I'm attempting to make the docs fall inline a bit more with other project structures. Currently, some content is a bit haphazardly placed and needs just a bit of re-org to get the information across.

Additionally, I'd like to improve the homepage a bit to better explain what the project is and guide users to the right section of the docs.

There might be some very minor design tweaks along the way (mostly typographic/containers since content is king), but the overall vibe should be the same.

Current goal is to: 

- reduce the content sprawl and to streamline a bit how the document is organized
- create links from the homepage to the appropriate doc page
- cleanup/organize the sass styles


1cg:
Gah, I pulled some changes in that conflict with this.

Can you hit me up on discord so we can coordinate this?

https://htmx.org/discord  (Im 1cg there too)

alexpetros:
We launched a major website refactor since this PR was created, so I think in its current form it's likely unsalvageable. If you still want to make some of these changes, feel free to re-open if you rebase. I do strongly recommend splitting it into smaller chunks, which will make it easier to review some of proposed website changes individually (and quickly). For instance one PR or the homepage changes, one PR for CSS cleanup, and so on.

PR #1007, opened 2022-08-15
[RFC] Always use latest `href`/`hx-*` in `hx-boost`

leiffoged:
hx-boost seems to capture the href/hx-get/etc of an element when the boosting logic is setup. 

But if the element changes its `href` since it was created, `hx-boost` will cause it to redirect to the wrong location. While opening that link in a new tab/window (bypassing hx-boost) will sent to the correct URL.

This proposed change would -- when performing an AJAX request -- revalidate that the element should be boosted and compute the verb/path at that time.

Thanks for your consideration! :)

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1221,23 +1221,29 @@ return (function () {
                 getRawAttribute(elt,'href').indexOf("#") !== 0;
         }
 
+        function shouldBoost(elt) {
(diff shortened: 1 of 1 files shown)


1cg:
I like this change *but* I'd like to do it for hx-get, etc. as well.  Are you willing to look at that bigger change?

leiffoged:
We recently moved our hx-boost usage to turbolinks (while still using HTMX for everything else), so I probably won't have the bandwidth to get to this. Sorry! :)

PR #1013, opened 2022-08-21
Allow 'unset' directive for hx-vals and hx-vars

B-Lenton:
This PR aims to fix the issue of the "unset" directive not working for `hx-vals` or `hx-vars` due to it being parsed as JSON.

See [this codepen](https://codepen.io/b-lenton/pen/NWYoQMo) for an example of `hx-vals="unset"` both causing an error and not unsetting the inherited values.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2418,7 +2418,7 @@ return (function () {
             if (attributeValue) {
                 var str = attributeValue.trim();
                 var evaluateValue = evalAsDefault;
-                if (str.indexOf("unset") === 0 && str.length === 5) {
+                if (str === "unset") {
                     return null;
                 }
                 if (str.indexOf("javascript:") === 0) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2418,6 +2418,9 @@ return (function () {
             if (attributeValue) {
                 var str = attributeValue.trim();
                 var evaluateValue = evalAsDefault;
+                if (str.indexOf("unset") === 0 && str.length === 5) {
+                    return null;
+                }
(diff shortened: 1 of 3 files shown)


src/htmx.js
```diff
@@ -2418,6 +2418,9 @@ return (function () {
             if (attributeValue) {
                 var str = attributeValue.trim();
                 var evaluateValue = evalAsDefault;
+                if (str.indexOf("unset") === 0 && str.length === 5) {
```
Renerick, line 2421: Maybe I'm missing something, but why not just `str === "unset"`?

test/attributes/hx-vals.js
```diff
@@ -184,5 +184,72 @@ describe("hx-vals attribute", function() {
...
+    });
+
+    it('multiple hx-vals can be unset', function () {
+        this.server.respondWith("POST", "/vars", function (xhr) {
+            var params = getParameters(xhr);
+            params.should.be.empty;
+            xhr.respond(200, {}, "Clicked!")
+        });
+        make(
+            "<div hx-vals='\"v1\":\"test\", \"v2\":42'>\
```
Renerick, line 228: Invalid JSON value in `hx-vals`

Renerick: Also below

src/htmx.js
B-Lenton: Fair point. Will change that tomorrow.

Renerick (approved):

1cg:
thank you!  great change, thank you for the tests!

PR #1014, opened 2022-08-23
Properly set title during history cache miss

leiffoged:
I am running HTMX with hx-boost and the history cache disabled (it's important to not show stale content to my users unless my backend uses cache-control headers to do so, in which case I simply rely on the fetch being cached). However, during 'history cache misses', the page title is not properly set.

This fixes it (mirroring the logic in `handleAjaxResponse`).


1cg:
thank you!

PR #1022, opened 2022-09-01
Add Michael's partials projects for Python section

mikeckennedy:
There are two packages I created specifically for integrating HTMX with Python web frameworks like Flask, FastAPI, Pyramid, etc. This PR adds links to them in the _known implementations_ section.

### DIFF
--- a/www/essays/template-fragments.md
+++ b/www/essays/template-fragments.md
@@ -134,6 +134,8 @@ Here are some known implementations of the fragment concept:
 
 * Python
   * [Django Render Block Extension](https://pypi.org/project/django-render-block/)
+  * [jinja_partials package](https://github.com/mikeckennedy/jinja_partials) ([discussion](https://github.com/mikeckennedy/jinja_partials/issues/1) on motivation)
+  * [chameleon_partials package](https://github.com/mikeckennedy/chameleon_partials)
 * Java
   * [Thymeleaf](https://www.thymeleaf.org/doc/tutorials/3.0/usingthymeleaf.html#fragment-specification-syntax)
   * [Chill Templates (currently in early alpha)](https://github.com/bigskysoftware/chill/tree/master/chill-script)


sponsfreixes:
For much that I love jinja_partials, I think this PR needs to be reverted.

The point of the essay is having templates as a single file which includes fragments that can be retrieved out of it. Instead, jinja_partials offers an easy way to reuse templates that contain a full fragment, having one file per fragment.

mikeckennedy:
Hi @sponsfreixes Thanks for the love. While this library does not offer a 100% automatic way to implement this idea, it does make it trivial and simple to accomplish without code duplication. 

Consider two files: content.html and full_page.html. You may want to show just content or the whole page with the content. Just define full_page.html as:

```
<html> ...
<h1>...</h1>
{{render_parital(content.html, data) }}
...
</html>
```

And you have the idea implemented. I'll leave it up to Carson to decide whether it is invalid if the details contained within `content.html` must be literally written within `full_page.html` or whether there is this type of function calling reuse that "counts".

I'm fine if you revert it if it's not a fit. But also the ASP.NET version is identical FWIW.

sponsfreixes:
Unless I am misunderstanding it, with this particular essay Carson wants to highlight the pattern of having a single file where fragments can be extracted, instead of relaying on multiple files. Particularly, the beginning of the article mentions:

> [...] without pulling fragments of the template out to separate files for rendering, creating a large number of individual template files.
> 
> By keeping all the HTML in a single file, it is also easier to reason about how a feature works. This follows the Locality of Behavior design principle.

That would be the case if we had a single `full_page.html` like:
```jinja2
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>This is the title</title>
</head>
<body>
<h1>This is a header</h1>
    {% block content %}
        <p>This is the content.</p>
    {% endblock %}
</body>
</html>
```

And then we were able to do:
```python
@app.get('/')
def search():
    data= ... # Get data from somewhere as a dict
    if is_htmx_request:
        html = new_library.render_fragment('full_page.html', block="content", **data)
    else:
        html = flask.render_template('full_page.html', **data)

    return flask.make_response(html)
```

PS: I am using jinja_partials in my projects with HTMX, thanks for it! :)

mikeckennedy:
Hey @sponsfreixes glad the library is work well for your project! In terms of whether jinja_partials, chameleon_partials, and ASP.NET partials belong on this doc is really a call I'll leave to Carson. I feel like the example fits well enough given the whole is created without duplicate markup, but I do see your point.

PR #1023, opened 2022-09-02
Add ASP.NET Core sample code and link

AlexZeitler:
This PR adds ASP.NET Core Sample Code for Partial Views and a link to the full sample code on GitHub

PR #1024, opened 2022-09-02
Remove ASP.NET Core from list

AlexZeitler:
As clarified in #1023, ASP.NET Core doesn't have the concept similar to fragments.


AlexZeitler:
@1cg Now that we have a working approach for ASP.NET Razor Views, do you think it makes sense to create a new PR with a link to the blog post?

https://alexanderzeitler.com/articles/htmx-razor-fragment-single-view-approach/

PR #1034, opened 2022-09-06
Avoid: "Element.querySelector: '...' is not a valid selector"

deveil:
When hx-get swap for an xml page (inkscape svg for exemple) with namespaces on elements, htmx failed.
This trick avoid the error raised.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -779,7 +779,7 @@ return (function () {
         function handleAttributes(parentNode, fragment, settleInfo) {
             forEach(fragment.querySelectorAll("[id]"), function (newNode) {
                 if (newNode.id && newNode.id.length > 0) {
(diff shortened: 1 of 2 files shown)


1cg:
Would you  mind adding a test demostrating this issue, and retargeting to `dev`?

deveil:
I need to see how unit tests are performed in your repository to take
examples.
I will try to do that, perhaps next week or later.

1cg:
Can you retarget this to the `dev` branch?

alexpetros:
Hi @deveil, sorry that this has been sitting for a bit. If you can perform the following updates I'll get it looked at ASAP:
* Rebase against the current dev branch
* Explain exactly what your "trick" does, because we it's not obvious and we will probably want to document what's broken exactly.

alexpetros:
Feel free to reopen if you're still interested.

PR #1036, opened 2022-09-07
Add Pushup to template fragments implementation

paulsmith:
Pushup is a Go web framework (unreleased but coming soon) that has an "inline partials" feature which is the same as template fragments.


1cg:
Would you mind if we added this when it is released?

I will co-market it on the htmx_org twitter account too...

paulsmith:
Yep, thank you! Should be soon …

1cg:
has this been released?

paulsmith:
No, but working on it. Hopefully in the next week.

paulsmith:
I'm happy to close this out now to help clear the decks and re-PR after there is a URL to point to if you like.

1cg:
that would be great, please PR when ready

PR #1051, opened 2022-09-19
[DRAFT] Don't overwrite Content-Type header if it has been set

dkniffin:
This allows the `Content-Type` to be set via headers in the case of `htmx.ajax`

NOTE: I have not tested this yet. This is a proof-of-concept.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2684,7 +2684,7 @@ return (function () {
             var allParameters = mergeObjects(rawParameters, expressionVars);
             var filteredParameters = filterValues(allParameters, elt);
 
-            if (verb !== 'get' && !usesFormData(elt)) {
+            if (!headers['Content-Type'] && verb !== 'get' && !usesFormData(elt)) {
                 headers['Content-Type'] = 'application/x-www-form-urlencoded';
             }
 
@@ -3294,4 +3294,3 @@ return (function () {
     }
 )()
 }));
-


dkniffin:
Actually... hold off on looking at this. I don't think this fixes my problem, it just moves it. With this change, I can set the `Content-Type`, but I don't think the content of the request is encoded correctly

dkniffin:
@1cg Alright. So here's a summary of what I'm trying to do.

I've got a Rails app that was using jQuery and SJR (server-side javascript rendering). Basically, what the previous version of my code does is use jQuery to make an AJAX request to an endpoint, sending some JSON over with the request, and the server responds with a javascript snippet that includes some HTML inside it, and the Javascript swaps the HTML into place.

jQuery is old and I want it gone from this app. And the SJR pattern isn't a great one either. So I'm replacing those with with vanilla JS + Stimulus, and HTMX.

In a lot of cases, this means using HTML attributes for HTMX, and things work nicely. But I'm working on a piece of the app that is more complicated and requires more Javascript code. I'd like to still use HTMX to swap in the content, but the request needs to be made via JS instead of HTML.

I've got that hooked up in my app and in some places it's working fine. But in one particular spot, where I need to send additional data to the backend, it's not. Basically, I've got an object like this:

```
const values = { a: 1, b: { c: 1, d: 2 } }
```

and I'm trying to send it like this:

```
htmx.ajax("post", url, { swap: "none", values: values, source: element })
```

When I do that, it sends a request, but it sends the data in the wrong format. In the previous version of my app's code, it sent it with `Content-Type: application/json`. So I tried changing the header (hence this PR). But as mentioned above, the body of the request seems to be wrong still.

So now I'm thinking: maybe I don't need `application/json`, as long as I can send the data along in the right format for the default `Content-Type`. But doing what I did above sends it over wrong still

Inspecting the request payload in the browser, I see something like:

```
a: 1,
b: { "c": 1, "d": 2 }
```

Instead, I think that should be something like:
```
a: 1,
b[c]: 1
b[d]: 2
```

Any thoughts?

1cg:
ah, yeah, that last encoding is a rails convention.  But that should work.

dkniffin:
Oh, I see. After looking at the spec mentioned [here](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams), I see the value is supposed to be a string. So the Rails works around that by encoding the structure into the key name.

Thanks for pointing that out. I'm not sure exactly what I'll do to solve my problem yet, but that should help me move in the right direction.

dkniffin:
I ended up just setting the key to `b[c]` and `b[d]` manually. Thanks for the help!

PR #1057, opened 2022-09-23
Fix oob before swap

elabx:


Telroshan:
So if I understand correctly, this change passes around the `xhr` reference to have it available in the `htmx:oobBeforeSwap` and `htmx:oobAfterSwap` events to the user binding a callback to those events, just like `htmx:beforeSwap` and `htmx:afterSwap` [expose it](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L3337) already?

alexpetros:
Closing due to inactivity, if you want to re-open, just rebase it and we'll take a look.

PR #1071, opened 2022-10-02
Update docs.md

vladstudio:
I believe ".my-indicator" is a typo, and correct CSS code should read ".htmx-indicator" instead.


1cg:
Believe it or not, that is correct: it is showing you how to create your own indicator class for use w/ htmx.

Needs to be clarified!

PR #1099, opened 2022-10-23
Add support for websocket binary data

TristanCacqueray:
This change enables ws extension user to handle binary data.

### DIFF
--- a/src/ext/ws.js
+++ b/src/ext/ws.js
@@ -29,6 +29,11 @@ This extension adds support for WebSockets to htmx.  See /www/extensions/ws.md f
 			if (htmx.config.wsReconnectDelay == undefined) {
 				htmx.config.wsReconnectDelay = "full-jitter";
 			}
+
+			// Default setting for binary data handler
+			if (htmx.config.wsOnBinary == undefined) {
+				htmx.config.wsOnBinary = () => false;
+			}
 		},
 
 		/**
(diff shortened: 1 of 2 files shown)


TristanCacqueray:
Hello, thank you for htmx, it's great :) Please let me know if this change is appropriate.

1cg:
@Renerick do you mind reviewing this?

Renerick:
Hi! Thanks for your PR, I'll review this ASAP

Renerick:
Hello, @TristanCacqueray and thank you again for your contribution. I've reviewed this PR and here are my some of my thoughts.

First of all, I'm 100% on board with adding support for raw binary data in websockets. There already were some discussions regarding files support in outgoing messages so it is only logical to support it in the incoming messages as well.

That being said, having a page global callback in the config is not exactly htmx style. Typically this is done through events and I think we should stick with it. Unfortunately, this means I can not approve this change in its current state.

However, you may have noticed that at the moment WS extension has basically no events support, and this is entirely on me. While I've been doing some tinkering with it, I had no chance to publish anything finished yet. To signify that this should be fixed at the first opportunity (there are already two PRs that could be done through events), I opened the issue (see above).

For now, I'll leave this PR open and will get back to it once WS events are implemented, so you can see for yourself if the implementation is satisfactory for your use-case.

TristanCacqueray:
@Renerick thanks for the review, good to hear ws binary data support is planned, I'd be happy to use another API you see fit for htmx.

1cg:
I am going to close this PR to help clean up my backlog.

@TristanCacqueray @Renerick do you want me to create an issue or discussion for this?

Renerick:
This will be covered by #1110

Renerick:
FYI, I just finished #1126. It introduces `wsBinaryType` configuration option that control corresponding WebSocket property. You can then access message content through `htmx:wsBeforeMessage` and `htmx:wsAfterMessage` events.

Any feedback about this solution will be much appreciated. Thank you for raising this issue!

PR #1101, opened 2022-10-24
Web Types: rename, update schema, add css classes and register in package.json

piotrtomiak:
Hi! I hope you'll find this finishing touches useful and you'll be able to ship Web Types with your library. I haven't touched any config file to actually include Web Types file as I am not sure how you build the package.

It would be cool to actually generate documentation from MD files, so if you would like that, I can experiment a bit and try to dynamically produce Web Types during build. I would only need to know into which file I should plug-ing the generation code :)


1cg:
Thank you @piotrtomiak, I'm very interested in whatever you can contribute to the project!

PR #1105, opened 2022-10-29
Add `hx-swap-title` to prevent title change

reedspool:
Simpler than I thought. If `hx-swap-title` has a value of "no" then don't swap the page title. This attribute can be inherited.

Let me know if this makes sense, and I will add/edit docs

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1003,7 +1003,10 @@ return (function () {
         }
 
         function selectAndSwap(swapStyle, target, elt, responseText, settleInfo) {
-            settleInfo.title = findTitle(responseText);
+            var shouldSwapTitle = getClosestAttributeValue(elt, 'hx-swap-title');
+            if (shouldSwapTitle !== "no") {
+                settleInfo.title = findTitle(responseText);
+            }
             var fragment = makeFragment(responseText);
             if (fragment) {
                 handleOutOfBandSwaps(elt, fragment, settleInfo);
--- a/test/core/ajax.js
+++ b/test/core/ajax.js
@@ -845,6 +845,18 @@ describe("Core htmx AJAX Tests", function(){
         window.document.title.should.equal("</> htmx rocks!");
     });
 
+    it('hx-swap-title value of "no" prevents title update', function () {
+        var originalTitle = window.document.title
+        this.server.respondWith("GET", "/test", function (xhr) {
+            xhr.respond(200, {}, "<title class=''>htmx rocks!</title>Clicked!");
+        });
+        var btn = make('<button hx-get="/test" hx-swap-title="no">Click Me!</button>')
+        btn.click();
+        this.server.respond();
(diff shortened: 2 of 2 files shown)


1cg:
Would you mind implementing this as an event/extension rather than a core attribute?

I'm trying to minimize the total number of attributes, if possible, and this seems like a minor feature to dedicate an attribute to.  Happy to show you how to do it if so!

Thank you!

reedspool:
I don't mind doing that. But while this may not be worth an attribute, I
argue it is a core concern. The default behavior to always swap title when
htmx is supposed to be more fine-grained is a strong choice that several
have asked to turn off somehow.

is there a way you envision that behavior getting toggled in core? e.g. a
global config? Or maybe another qualifier in the hx-swap value, like
`hx-swap="#output notitle"`?

If you say no, I'm down to try making it an extension

1cg:
Interesting idea to make it part of the swap syntax. I'd like to think about that...
------- Original Message -------

1cg:
I wonder if this dovetails w/ the new head-support extension coming in 1.8.3, and we should allow swap specifications to drive head merge behavior.  Have you looked at that at all yet?

reedspool:
I've looked at it a bit and I think it does dovetail conceptually. I'm
happy to move this in whatever direction you see fit.

jyrimatti:
Hi,

force-swapping the title is giving me troubles in some use cases.

I just tried adding hx-preserve="true" (with head-support included and enabled) to title-element but it still did not preserve it. Should this work?

Tagging the title might not be a proper solution, though, since sometimes I'd want it to swap (changing large portion of the page), and sometimes not (changing only something small). So I'd feel the best place to indicate whether to change the title or not is the triggering element (inherited or not), either as a new attribute, or as a modifier to an existing one, or as some fancy logic like "if hx-select is given then the title is not swapped".

Is a custom extension currently the only way to preserve the title? I'm in need of _some_ solution until a proper one is implemented...

jyrimatti:
Hi,

I'm currently using an extension like this:
```
htmx.defineExtension('swap-notitle', {
    handleSwap: function(swapStyle, target, fragment, settleInfo) {
        delete settleInfo.title;
        return false; // always return false to continue processing
    }
});
```

Is this the proper/best way to prevent the title from changing?

I haven't used Htmx much so I probably cannot understand all the different aspects and edge cases, but currently I feel that the best way to handle title change would be logic:
"Don't swap title if hx-select is given".
Rationale would be that by giving hx-select, the user states to be interested only in subset of the response. If a title element happens to be outside of this subset, then it should not be used.

What do you think, are there any problems with this logic?

Telroshan:
Hey, does @1cg's commit c230931 (not merged on master yet, only available on dev) solve that issue?
> support ability to ignore new title tags via config, event and `hx-swap`

alexpetros:
Yup it does, the config is `ignoreTitle` now.

PR #1033, opened 2022-09-06
Minor changes in docs

David-Guillot:

PR #1106, opened 2022-10-29
fix typo attribute

klavman:
Attribute change according to the explained example.


limugob:
Is this not a bad example to make a search with POST? See also #1033
Would it be not better to change the description to mention GET?

PR #1109, opened 2022-10-30
Add 'distracted boyfriend' meme

dekoza:
New meme for you :kissing_heart:


1cg:
:) this makes me laugh, but in general I don't like going directly at any particular technology out there, and Next.js is a fine piece of software

dekoza:
That's why it's so funny :wink: But I understand, I'll come up with something more general :smile:

PR #1111, opened 2022-10-31
Cleaner Bootstrap 5 modals example.

dekoza:
After some experimentation I managed to get htmx working smoothly with Bootstrap 5 modals and their original JS bundle.

Edit:
OK, got it working after few tries. Turns out the initial skeleton needs to be a bit bigger. But it doesn't require writing any JS or even using hyperscript :slightly_smiling_face:

### DIFF
--- a/www/content/examples/modal-bootstrap.md
+++ b/www/content/examples/modal-bootstrap.md
@@ -4,8 +4,7 @@ template = "demo.html"
 +++
 
 Many CSS toolkits include styles (and Javascript) for creating modal dialog boxes. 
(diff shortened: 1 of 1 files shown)


1cg:
hey @dekoza  can you retarget this at `dev` so I can preview it?

alexpetros:
@dekoza I'm sorry for the delay on this, if you rebase and ping me we'll get it looked at.

dekoza:
@1cg @alexpetros I think I managed to update this properly :wink:

PR #1113, opened 2022-10-31
History cache embargo

croxton:
1. Allow the history snapshot feature to be disabled when `hx-history="false"` is present anywhere in the current document. This can be used to prevent sensitive or user session data being saved to the localStorage cache. Particularly important for public/shared use computers.
2. Adds a test for the attribute.


1cg:
@croxton can you please add documentation for this and mention it in the history section of docs.md?

croxton:
Will do

1cg:
Thank you! 🎉

PR #1144, opened 2022-11-26
Fix handling of repeated empty inputs

### LINKED ISSUE #487: Input array not send empty fields
Hello, I have a form where it has additional multifield and I use the name of the input to transform it into an array and receive it in the backend with PHP, something like:

`<input type = "text" name="subject[]">`

It turns out that as I can have several of these and the fields are optional, HTMX only sends those that are filled, ignoring the fields that are empty and this ends up generating the server-side matrix in the wrong way. With JQuery Ajax works correctly, is there any way to resolve this?

https://www.php.net/manual/en/faq.html.php#faq.html.arrays

bluekeyes:
Native form handling submits repeated empty inputs as an array with an empty value for each input, but htmx would only submit the first non-empty input and any following inputs. Fix the value collection code so that it correctly distinguishes between empty values and new inputs.

You can compare the browser-native and current htmx (1.8.4) behavior here: https://jsfiddle.net/sjzypx9e/

This fixes #487, which I think was mistakenly closed. I wasn't sure whether to create this against `master` or `dev`, so please let me know if I should rebase against a different branch.


1cg:
Thank you, I agree this is more consistent with plain HTML.

I'm a little worried about changing the semantics without an option to disable.  Let me think on it.  I hope to merge it this week.

bluekeyes:
Thanks, I understand the concern about the change in behavior without an option. For what it's worth, I think it would be hard to build an application that relied on the existing behavior because it depends so much on which particular fields a user fills out:

|Inputs|HTMX 1.8.4 Value|
|---|---|
|`""`<br>`""`<br>`""`|`""`|
|`"foo"`<br>`""`<br>`""`|`["foo", "", ""]`|
|`""`<br>`"foo"`<br>`""`|`["foo", ""]`|
|`""`<br>`""`<br>`"foo"`|`"foo"`|

fletcheaston:
Just to bump this, we're running into this issue as well and would really appreciate similar behavior to what standard forms/HTML provide.

1cg:
Can you do the pull against `dev` and I'll integrate it?

bluekeyes:
Sure, updated to target `dev`.

1cg:
Thank you!  Now that I've looked at the change more closely, I agree: this is how it should have worked all along and it is unlikely that anyone is relying on the current (broken) behavior.  

I will make this part of 1.8.5, hoping to release next week.

PR #1147, opened 2022-11-28
Ability to trigger preload always (not once)

alexbezhan:
When a response's max-age is very short, it becomes stale and in this situation we may want to trigger preload every time instead of once.
This PR enables to write `preload="mouseover always"` and have preload trigger every time.

### DIFF
--- a/src/ext/preload.js
+++ b/src/ext/preload.js
@@ -26,7 +26,9 @@ htmx.defineExtension("preload", {
 			// Called after a successful AJAX request, to mark the
 			// content as loaded (and prevent additional AJAX calls.)
 			var done = function(html) {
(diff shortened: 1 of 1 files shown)


1cg:
@benpate can you review this change?

benpate:
Hey, this change looks fine to me, and it's nice to see this extension getting attention.  

BTW, if/when we get access to the `htmx.swap()` function in htmx 2.0, I'll want to re-do the way this extension works.  Once we can trigger swaps independently of ajax calls, we won't need to rely on the browser's cache, which *should* greatly simplify the way this extension works, and remove all those finicky cache validation problems.

Issue #1153, opened 2022-12-05
eleventy-plugin-sass breaks build and is deprecated - cannot npm install


ieugen:
Hello, 

I am trying to contribute to htmx and I cloned the project and `npm install` 

I get an error that I tracked to -    "eleventy-plugin-sass": "^1.2.0" .

Removing the package from devDependencies, I can `npm install` .
I have python3. I am on Debian testing.

The page mentions https://www.npmjs.com/package/eleventy-plugin-sass :

> This package has been deprecated
Author message:
Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.

```
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated tar@2.2.2: This version of tar is no longer supported, and will not receive security updates. Please upgrade asap.
npm ERR! code 1
npm ERR! path /home/ieugen/proiecte/htmx/node_modules/node-sass
npm ERR! command failed
npm ERR! command sh -c -- node scripts/build.js
npm ERR! Building: /home/ieugen/.nvm/versions/node/v16.18.1/bin/node /home/ieugen/proiecte/htmx/node_modules/node-gyp/bin/node-gyp.js rebuild --verbose --libsass_ext= --libsass_cflags= --libsass_ldflags= --libsass_library=
npm ERR! gyp info it worked if it ends with ok
npm ERR! gyp verb cli [
npm ERR! gyp verb cli   '/home/ieugen/.nvm/versions/node/v16.18.1/bin/node',
npm ERR! gyp verb cli   '/home/ieugen/proiecte/htmx/node_modules/node-gyp/bin/node-gyp.js',
npm ERR! gyp verb cli   'rebuild',
npm ERR! gyp verb cli   '--verbose',
npm ERR! gyp verb cli   '--libsass_ext=',
npm ERR! gyp verb cli   '--libsass_cflags=',
npm ERR! gyp verb cli   '--libsass_ldflags=',
npm ERR! gyp verb cli   '--libsass_library='
npm ERR! gyp verb cli ]
npm ERR! gyp info using node-gyp@3.8.0
npm ERR! gyp info using node@16.18.1 | linux | x64
npm ERR! gyp verb command rebuild []
npm ERR! gyp verb command clean []
npm ERR! gyp verb clean removing "build" directory
npm ERR! gyp verb command configure []
npm ERR! gyp verb check python checking for Python executable "python2" in the PATH
[...]

PR #1154, opened 2022-12-05
[#1153] Removed dev plugin that breaks build on modern linux

ieugen:
Fixes https://github.com/bigskysoftware/htmx/issues/1153

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,6145 @@
 {
   "name": "htmx.org",
   "version": "1.8.4",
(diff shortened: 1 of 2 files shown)


ieugen:
I can't see the logs so I can't really figure out what is wrong :( .

1cg:
Man, upgrading eleventy is a huge pain.  I'm going to have to close this for now, will look at fixing for htmx 2.0.  Upgrading a static website generator shouldn't be this hard!

PR #1155, opened 2022-12-05
[#1115] Client-side-templates fix - support arrays

### LINKED ISSUE #1115: htmx client-side-template does not work with JSON array
Hi,

If the server returns an array, then client-side-template can't be used.

https://github.com/bigskysoftware/htmx/blob/master/src/ext/client-side-templates.js

The code to process the data uses it directly, without wrapping it in an object.
```js
           var data = JSON.parse(text);
           .......
           return Mustache.render(template.innerHTML, data);
```
If the server returns an array then I believe it's not possible to render that.

ieugen:
* Provide data as property named `data` so we can render arrays

Fixes #1115 (closed before I had a chance to make this PR).

### DIFF
--- a/src/ext/client-side-templates.js
+++ b/src/ext/client-side-templates.js
@@ -7,7 +7,7 @@ htmx.defineExtension('client-side-templates', {
             var templateId = mustacheTemplate.getAttribute('mustache-template');
             var template = htmx.find("#" + templateId);
             if (template) {
-                return Mustache.render(template.innerHTML, data);
+                return Mustache.render(template.innerHTML, {"data": data});
             } else {
                 throw "Unknown mustache template: " + templateId;
             }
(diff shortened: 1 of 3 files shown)


ieugen:
hi @1cg , I made this PR for issue #1155 .

1cg:
this would be a breaking change for current users of this extension, correct?

ieugen:
Yes. I don't think it's a way around it. 
The upside is that now you can render arrays.

ieugen:
ping @1cg : do you think you'll have time to look at this?

1cg:
hey @ieugen I'm hesitant to publish a breaking change like this in a minor version bump

i feel like what we need to do is create a new extension (maybe specifically for moustache templates?) and make the changes there

another alternative would be to detect if the data is an array and then use the `data` symbol in that case, which should preserve the current behavior

ieugen:
hi, I totally understand not wanting to push breaking changes in a patch.

I wold not complicate logic with array detection. 
This could wait for a future (minor ?! ) release.
I think the existing implementation should be fixed since it does not work right now with arrays - hence it's a bug IMO. 

If the change is highlighted well people should have an easy time to upgrade.

PR #1156, opened 2022-12-06
Actually pass original event on to `htmx:configRequest`

thomaseizinger:
Previously, we passed the _element_ as the original event. I think this bug was introduced in https://github.com/bigskysoftware/htmx/commit/8bb0399052f5a52a49e9e4f0f9624d25f1c2e890.

We can easily fix this by actually using both arguments of the `handler` closure.

Let me know if there is a better way to write this test.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1275,7 +1275,7 @@ return (function () {
                     path = getRawAttribute(elt, 'action');
                 }
                 triggerSpecs.forEach(function(triggerSpec) {
-                    addEventListener(elt, function(evt) {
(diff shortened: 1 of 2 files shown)


1cg:
oosh, nasty bug, thank you for catching it

can you retarget to `dev` branch for me?

thomaseizinger:
> can you retarget to `dev` branch for me?

Done.

thomaseizinger:
Is anything blocking this from being merged @1cg?

1cg:
just time, laziness and intellectual ability ;)

PR #1160, opened 2022-12-08
Change scroll behavior default

gone:
Defer committing until 2.0


1cg:
i'd like to make this a 2.0 change

PR #1168, opened 2022-12-12
[Fixes #1128] submit button child elements not sending name or value of parent attributes in form request

### LINKED ISSUE #1128: Submit button child elements do not send the name or value of the parent attributes in the request
When using a standard submit button:
```
<button type="submit" id="edit-save" name="edit-save" value="save">Save</button>
```
The form will submit with name and value in the request.
However when the button has child elements like a span for example:
```
<button type="submit" id="edit-save" name="edit-save" value="save"><span>Save</span></button>
```
The request is missing the parent buttons name and value if you click on the span element.

kjanuska:
Fixes #1128 

This fixes child elements nested within a submit button not sending the appropriate attributes on a form request.
For example, the following code would submit `{ "foo": "bar" }`:
```
<form>
    <input/>
    <input/>
    <button type="submit" name="foo" value="bar">Submit</button>
</form>
```
whereas prior to the change, the following code would **not** send those same attributes (when the `span` was clicked):
```
<form>
    <input/>
    <input/>
    <button type="submit" name="foo" value="bar">
        <span>Submit</span>
    </button>
</form>
```
This change allows for more complex button styling while keeping form functionality simpler.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1781,6 +1781,12 @@ return (function () {
                 if (matches(evt.target, "button, input[type='submit']")) {
                     var internalData = getInternalData(form);
                     internalData.lastButtonClicked = evt.target;
+                } else { // in case clicked element is nested within a button
+                    var elt = closest(evt.target, "button");
+                    if (elt !== null) {
+                        var internalData = getInternalData(form);
+                        internalData.lastButtonClicked = elt;
+                    }
                 }
             };
 
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1777,18 +1777,18 @@ return (function () {
         }
 
         function initButtonTracking(form){
-            var maybeSetLastButtonClicked = function(evt){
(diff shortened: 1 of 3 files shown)


src/htmx.js
```diff
@@ -3338,11 +3344,57 @@ return (function () {
...
+            }
+            var curr = sub[len - 1];
+            key = curr.substring(0, curr.indexOf("="));
+            key = convertSyntax(key);
+            val = curr.substring(curr.indexOf("=") + 1, curr.length);
+            val = convertSyntax(val);
+            object[key] = val;
+            return object;
+        }
+
```
1cg, line 3392: seems like two changes bleeding into one here?

1cg:
I have integrated the spirit of this PR here:

https://github.com/bigskysoftware/htmx/commit/666c6aa8e57ad986741adf85b6714d1d40c3d7e8

Thank you!

Issue #1132, opened 2022-11-13
hx-post on simple button does not send name or value


pyc-mo:
Hello,

The examples in the documentation figure hx-post without any paramers.
According to the documentation "_By default, an element that causes a request will include its value if it has one._ "

It does not seem to work with a simple button:

```
<button name="btn" value="delete" hx-post="/reset">
	Enable Your Account
</button>
```

Does not post anything (request body is empty).

HTMX version used : 1.8.4

NBA707:
It looks like it works for html  input, checkbox, textarea but not button or div tags, you could do something like:

  <input type="hidden" name="enableAccount" value="delete"/>

    <button name="btn" hx-post="/testhtmx" hx-include="[name='enableAccount']">
        Enable Your Account
    </button>

Then it does the payload with enableAccount with value delete.

flibustenet:
You can maybe use `hx-vals`, but i believe you're right, the value of the button should be posted according to the documentation.

spookylukey:
For me this is also not working with:

```html
<input type="submit" hx-trigger="click" name="delete" value="Delete" >
```
I would expected `delete=Delete` to be submitted in this case.

htmx 1.8.4

julianwachholz:
I too have run into this issue but it doesn't happen all the time. From my perspective it looks like this scenario does work **if**:
- The button is of type="submit"
- and has both `value` and `name` attributes

This is how it would also work without any javascript. It works on most browsers except Safari on iOS.

My use case has multiple different submit buttons on a single form where the action depends on the button clicked. E.g.:

```html
<form method="post" action="">
  <button type="submit" name="action" value="increase">Increase Counter</button>
  <button type="submit" name="action" value="decrease">Decrease Counter</button>
</form>
```

samueleaton:
@julianwachholz not quite the same, although using a form (or an anchor tag) seems to be the main workaround. 

Form submissions import *all inputs with values* (which includes buttons). This issue is for when you aren't using a form element. 

I have an open PR i submitted a couple weeks ago that fixes this.

1cg:
If my understanding is correct, with normal forms the button/submit input value is only included when that button is clicked on.  We attempt to mimic this behavior with the code here:

https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1773

https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1819

[...]

PR #1174, opened 2022-12-20
Don't ignore button params if button is input element

samueleaton:
Addresses the issue in #1132 

Currently, button parameters are never sent to the server. I believe that should still be the default but this PR allows users to be opt-in by using an input element instead of a button element. 

The input element must be type `button` or `submit`, and it must have a name and a value.

I think this makes sense since a user would expect an `<input ... />` element to send data to the server, whereas it makes sense why a button element wouldn't.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2120,8 +2120,14 @@ return (function () {
                 return false;
             }
             // ignore "submitter" types (see jQuery src/serialize.js)
-            if (elt.type === "button" || elt.type === "submit" || elt.tagName === "image" || elt.tagName === "reset" || elt.tagName === "file" ) {
-                return false;
+            if (elt.type === "button" || elt.type === "submit") {
+              // If is an input button (i.e. <input type="submit" name="..." value="..." />)
+              // then include the name & value if they exist.
+              // We know this input button already has a name from above logic.
+              if (elt.tagName !== "INPUT" || !elt.value) return false;
+            }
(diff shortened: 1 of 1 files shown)


samueleaton:
If this is "approved in principle", I will add tests.

src/htmx.js
```diff
@@ -2120,8 +2120,14 @@ return (function () {
...
             // ignore "submitter" types (see jQuery src/serialize.js)
-            if (elt.type === "button" || elt.type === "submit" || elt.tagName === "image" || elt.tagName === "reset" || elt.tagName === "file" ) {
-                return false;
+            if (elt.type === "button" || elt.type === "submit") {
+              // If is an input button (i.e. <input type="submit" name="..." value="..." />)
+              // then include the name & value if they exist.
+              // We know this input button already has a name from above logic.
+              if (elt.tagName !== "INPUT" || !elt.value) return false;
+            }
+            if (elt.tagName === "image" || elt.tagName === "reset" || elt.tagName === "file" ) {
```
samueleaton, line 2129: Was wondering if tagName is expected to be lower case. Shouldn't it be "IMAGE"?

1cg:
iirc, buttons (including input[type=submit]) should only be included in a request if they are the source of the form submission

i think your comment on the tagName capitalization is correct!

samueleaton:
@1cg got it. can you address this in #1132. thanks!

PR #1179, opened 2022-12-22
fixes where dynamically adding hx-disable is not processed

### LINKED ISSUE #1176: Dynamically adding hx-disable to element doesn't disable htmx actions on element
Adding `hx-disable` to an element (e.g. `myElement.setAttribute("hx-disable", ""); htmx.process(myElement)`) doesn't disable the element. 

Using `htmx.logAll.()` I didn't see any events triggered after running `htmx.process(...)`.

I think that when the `htmx.process(myElement)` is run, it sees that the element has "hx-disable" and stops processing it, instead of removing all triggers.

### Why Add hx-disable Dynamically?

I feel like adding "hx-disable" more is more inline with hypermedia principles than doing `evt.preventDefault()` inside an `htmx:confirm` event when i want to prevent an event. For example, when I click a dropdown button when the dropdown menu is already visible, I don't want to re-fetch the menu.

samueleaton:
Fixes #1176 

Adding the `hx-disable` to an element and then running `htmx.process(...)` now de-registers the element.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1791,18 +1791,26 @@ return (function () {
         }
 
         function initNode(elt) {
+            var nodeData = getInternalData(elt);
+
             if (elt.closest && elt.closest(htmx.config.disableSelector)) {
+                if (nodeData.initHash) {
+                  triggerEvent(elt, "htmx:beforeProcessNode")
+                  nodeData.initHash = null;
+                  deInitNode(elt);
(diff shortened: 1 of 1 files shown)


Telroshan:
Hey, following up late, did #1567 resolve that issue?
If it's not the case, could you please update the branch against `dev` and resolve conflicts to have an up-to-date diff?

PR #1181, opened 2022-12-30
Replace keyup with input event in Search Example

samueleaton:
`keyup` only works with a keyboard, and ignores pasting with a right click or programmatically updating the value with JS. 

`input` is a better event for detecting any "inputs to the input".


alexpetros:
Rebasing and closing in favor of #1952

Issue #1186, opened 2023-01-03
Triggering a swap from an element that doesn't support setSelectionRange causes error


gone:
See https://jsfiddle.net/udcq74nm/1/ for a repo

1cg:
thank you for your service

PR #1187, opened 2023-01-03
Hotfix for swapping fields that don't support setSelectionRange

gone:
fix for #1186 

Tests aren't passing - can't figure out how to submit the form based on enter key

### DIFF
--- a/test/core/regressions.js
+++ b/test/core/regressions.js
@@ -187,22 +187,25 @@ describe("Core htmx Regression Tests", function(){
 
 
     it("can trigger swaps from fields that don't support setSelectionRange", function(){
-        const template = '<form method="get" action="/test" hx-boost="true">\n' +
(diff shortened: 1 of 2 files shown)


test/core/regressions.js
```diff
@@ -185,4 +185,24 @@ describe("Core htmx Regression Tests", function(){
...
+
+        const response = '<form>\n' +
+              '<input value="supertest@test.com" type="email" id="id_email" />\n' +
+              '<input type="submit" id="id_button" />\n' +
+              '</form>';
+        this.server.respondWith("GET", "/test", response);
+        make(template);
+        var input = byId("id_email");
+        input.focus();
+        input.dispatchEvent(new KeyboardEvent('keypress', {"key": "Enter"}));
```
1cg, line 203: maybe make the trigger a click and then `click()` the input?  I don't think which event happens matters so much...

gone:
Updated tests, passing now.

PR #1191, opened 2023-01-04
Add Shadow Dom support 

OnurGumus:
It basically adds shadow dom support to htmx. Node that hyperscript won't support shadow dom but Alpine does if you do

```html
<my comp "x-init"="htmx.process($el); Alpine.initTree($el.shadowRoot);"><my-comp>
```


alexpetros:
Unfortunately you rebased against the `master` branch, not the `dev` branch, so now there are a ton of unrelated commits here, making the changes impossible to review.

I will say that adding shadow DOM support is complicated, and while we're definitely interested in doing it, I do remember having some concerns about your approach (especially how it was initialized) when I read the PR from before the rebase.

OnurGumus:
Yes sorry about that. But even so, what I did was fairly simple. And I believe these changes are easily portable. To be honest I am not actively using htmx these days. When I have some time, I will try to report them into dev branch if I can.

alexpetros:
In that case, I'm going to close this PR for now. Feel free to re-open or start a new one when you've got the bandwidth for it :)

PR #1203, opened 2023-01-15
Fix line spacing

Rowlando13:
The example looks really nice in general but looks a little cramped at the bottom.


Rowlando13:
Also first PR with the project so, let me know what I need to fix, Could also do it with a <p> tag.

PR #1210, opened 2023-01-20
Extend client-side-templates to handle XSLT transformations

jyrimatti:
This PR extends client-side-templates with XSLT support.
Uses browser built-in XSLTProcessor, so no external dependencies are needed.
Since this is only a fourth supported templating engine, I don't think it adds any complexity.

Please let me know if I can improve this in any way!

A couple of other things, which I didn't include:

I don't know how external script support for the other engines is supposed to work, but at least for XSLT it seemed to work using an object tag (see https://lahteenmaki.net/davtest/ for an example). However, I didn't include anything about it in the documentation since I guess the handling should somehow be unified across all supported engines, or what do you think?

I noticed that the HTML template tag is often not useful for templates (dah...) since the browser seems to enforce HTML validity inside the tag. But templates often contain invalid HTML, for example when looping over rows inside a table, which the browser doesn't seem to accept. Therefore I used the script tag in the example. Maybe this problem should be somehow mentioned in the extension documentation, or what do you think?

### DIFF
--- a/src/ext/client-side-templates.js
+++ b/src/ext/client-side-templates.js
@@ -49,7 +49,24 @@ htmx.defineExtension('client-side-templates', {
             } else {
                 return nunjucks.render(templateName, data);
             }
-          }
+        }
+        
+        var xsltTemplate = htmx.closest(elt, "[xslt-template]");
+        if (xsltTemplate) {
+            var templateId = xsltTemplate.getAttribute('xslt-template');
+            var template = htmx.find("#" + templateId);
+            if (template) {
+              var content = template.innerHTML ? new DOMParser().parseFromString(template.innerHTML, 'application/xml')
+                                               : template.contentDocument;
+              var processor = new XSLTProcessor();
+              processor.importStylesheet(content);
+              var data = new DOMParser().parseFromString(text, "application/xml");
+              var frag = processor.transformToFragment(data, document);
+              return new XMLSerializer().serializeToString(frag);
+            } else {
+              throw "Unknown XSLT template: " + templateId;
+            }
+        }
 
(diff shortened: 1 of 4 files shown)


jyrimatti:
Ran into another use case, where this would be useful:
Some data in the Internet is available as RSS (like Mastodon toots or GoodReads lists), which you could then include in your web page like this, by just writing a suitable XSLT transformation (in this case file rss.xml):
```
<script src="https://unpkg.com/htmx.org@1.8.5"></script>
<script src="https://unpkg.com/htmx.org/dist/ext/client-side-templates.js"></script>
<object id="tmpl" data="rss.xml"></object>
<div hx-get=”https://mastodon.online/@jyrimatti.rss” hx-trigger=”load” hx-ext=”client-side-templates” xslt-template=”tmpl”>loading…</div>
<object id=”tmpl” data=”rss.xml” style=”position:absolute; left:999999px”></object>

```

1cg:
Very cool!

Would you  mind pulling the documentation out into it's own little subsection (maybe we need headers for the mustache version and the xslt version) so it doesn't conflict w/ the existing `dev` branch?

jyrimatti:
Hi,

is this what you had in mind? Please feel free to modify as you wish, you are much better at writing documentation than I am :)

alexpetros:
Hi @jyrimatti, I'm sorry this has been sitting for so long! Would you mind rebasing this against the current dev branch so that we can look at it for an upcoming release?

jyrimatti:
Hi,
don't worry about it! I'm already using this myself, so getting it merged would just help other people to discover the possibility.

Please let me know if there's something more I could help with :)

alexpetros:
This looks super! I know that we have XSLT users and I'm sure they would take advantage of it.

PR #1222, opened 2023-01-26
Clear cache when historyCacheSize is set to 0

### LINKED ISSUE #1221: History cache not cleared after historyCacheSize has been set to 0 and cache was not empty
If the history cache is not empty and the `historyCacheSize` is set to 0, the history cache is not cleared, but still used. This leads to very subtle bugs.

https://github.com/bigskysoftware/htmx/blob/7ea35d03c6226aaf278b47d80170132bad9384f0/src/htmx.js#L1945-L1956

aomader:
Fixes #1221

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2175,6 +2175,12 @@ return (function () {
                 return;
             }
 
+            if (htmx.config.historyCacheSize <= 0) {
+                // make sure that an eventually already existing cache is purged
+                localStorage.removeItem("htmx-history-cache");
+                return;
+            }
+
             url = normalizePath(url);
 
             var historyCache = parseJSON(localStorage.getItem("htmx-history-cache")) || [];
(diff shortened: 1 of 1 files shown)


Telroshan:
Hey, if you're still interested in merging those changes, could you please retarget your PR to the `dev` branch ?

src/htmx.js
```diff
@@ -2193,6 +2193,7 @@ return (function () {
                     historyCache.shift(); // shrink the cache and retry
                 }
             }
+            if (historyCache.length === 0) localStorage.removeItem("htmx-history-cache");
```
1cg, line 2196: can you add a comment here?

Also maybe this should be up near the top and avoid all the logic in this method?

1cg (changes requested):

src/htmx.js
aomader: @1cg I added a comment and moved it to the top to exit early. Note however that this changes the current behavior w.r.t. the `htmx:historyItemCreated` that was triggered up until now. However, I'd argue that the new behavior is more correct.

1cg: agree

PR #1227, opened 2023-01-30
fix indicator styles flash

michaelperel:
In a SPA I am writing, I noticed that when adding an element with an indicator to the DOM, the indicator will "flash" (display for a brief moment). Specifically, in my case this is when content for a page is loaded via the "HX-Location" header.

I am not a CSS expert, but it appears as though the CSS transition is on the wrong element. Instead, the transition should be placed on the elements with opacity: 1 (when the request class is added).

I made that change locally in my app, and the "flashing" behavior went away. Here is a PR for that change.


alexpetros:
I am also no CSS expert, but this looks correct to me.

PR #1235, opened 2023-02-01
Add config option to ignore nested oob-swaps instead of processing them

### LINKED ISSUE #1133: Preserve elements with hx-swap-oob that are not at the root of a response
Currently when an element with the `hx-swap-oob` attribute is contained within an outer element in an hx response, the element with this attribute is deleted and not rendered to the document anywhere. I propose changing this behavior to instead *ignore* the `hx-swap-oob` attribute and include it in the final document at the original location within its outer element.

[initially discussed on discord](https://discord.com/channels/725789699527933952/725789747212976259/1041203047092727918)

### What is `hx-swap-oob`

When an hx request is made and its response is interpreted, the standard behavior is to swap in the contents of the response according to the `hx-swap` and `hx-target` attributes on the requesting element. In addition, if the response contains extra root elements with the `hx-swap-oob` attribute, they are instead swapped into the document at arbitrary locations selected by the value of the attribute. This is a good way to update some node in a remote / outer scope when a node in a smaller scope changes. For example: an input is used to add a new item to a list, the main response is just the inserted element, and an extra `hx-swap-oob` element is returned at the same time to update the count that is located outside the scope of the list container. If not for this feature, an operation like adding an element to a list could require downloading and re-rendering a much larger portion of the document just to enclose the outer scope of all elements that need to be modified after some operation.

### Intended use and edge cases

htmx expects elements with the `hx-swap-oob` attribute to be returned in an hx response as a *root* element of the response document. Like this: `<li><label>added list item content</label</li><span hx-swap-oob="true" hx-target="count">7</span>`. For the purpose of affecting remote elements, this restriction is reasonable to make its behavior understandable and predictable; if there were no such restriction an `-oob` tagged element could be nested *anywhere* in the response -- no matter how obscure -- which is too much unnecessary flexibility.

So the rule is "if you want to affect elements outside the scope of the response target, add a new element at the root of the response with the `hx-swap-oob` attribute". **This is great, but it doesn't say what htmx should do if an element with the hx-swap-oob` attribute *does* appear in an inner element in an hx response.**

There are a few reasonable behaviors to choose from:

1. Treat the whole response as an error and abort the response handling. (This would be valid but doesn't hold with the usual hypermedia approach to ignore things you don't understand and apply your best effort to render something useful.)
2. Assume that the presence of the inner element with `hx-swap-oob` is a programming error and omit the entire element from the rendered document. **This is the current behavior.**
3. Ignore that the `hx-swap-oob` attribute is present and render the inner element as if this attribute wasn't there. **This is my proposal.**

### Why prefer behavior 3?

* This is more in keeping with the 'best effort' rendering strategy mentioned before. 
* This unifies the fresh-page-load and hx response loading behavior. If the user refreshes or loads a whole new page these elements are rendered normally, as if the attribute is omitted, but if they are present in an hx response they are currently omitted.
* This would simplify the renderer code by allowing it to always include the `hx-swap-oob` attribute when rendering a specific element, which could be reused in different contexts such as rendering the whole page fresh and in response to an hx request.

### Example

Here's a completely static example of what I wish would work in 3 endpoints, `hi`, `list`, and `add-to-list`:

Full page load at `/hi`:

```html
<script src="https://unpkg.com/htmx.org@1.8.2/dist/htmx.min.js"></script>

<title>Hi</title>

<div class="tabs">
    <a href="hi" hx-get="hi" hx-push-url="hi" hx-target="body">Hi</a>
    <a href="list" hx-get="list" hx-push-url="list" hx-target="body">The List</a>
</div>
<p>Hello!</p>
```

Full page load at `/list`:

```html
<script src="https://unpkg.com/htmx.org@1.8.2/dist/htmx.min.js"></script>

<title>The List</title>

<div class="tabs">
    <a href="hi" hx-get="hi" hx-push-url="hi" hx-target="body">Hi</a>
    <a href="list" hx-get="list" hx-push-url="list" hx-target="body">The List</a>
</div>
<button hx-get="add-to-list" hx-target="#mylist" hx-swap="beforeend">Add Item</button>
<ul id="mylist">
    <li>item1</li>
    <li>item2</li>
</ul>
<p>There are <span id="count" hx-swap-oob="true">2 items</span> in total.</p>
```

HX response at `/add-to-list`:

```html
<li>item3</li>
<span id="count" hx-swap-oob="true">3 items</span>
```

Demo steps:

1. Start at `/hi`
2. Click on `The List` link to use htmx to navigate to the The List page
3. Note how the span with "2 items" is missing
4. Click on the Add Item button
5. Note how the list is updated to add item3, but the span with the item count is still missing
6. Refresh the page at `/list`
7. Note how the span with "2 items" is visible
8. Click on the Add Item button
9. Note how the list is updated to add item3 like before, but also the span with the item count is updated successfully

I think that the steps from 2-5 should behave the same as the steps from 6-9.

---

I'm sorry this issue description is so long, if I had more time I would have made it shorter. :)

Thoughts?

infogulch:
This adds a new configuration option, `allowNestedOobSwaps`, which preserves existing behavior when set to the default `true`, but when set to `false` it changes `handleOutOfBandSwaps` to only process elements with `hx-swap-oob` and `data-hx-swap-oob` that appear at the *root* of the response fragment, i.e. elements whose parent element is `body`. Elements deeper in the dom tree fragment are ignored and the `oob-swap` attribute is stripped instead.

> The current system means you can just randomly intersperse oob fragments anywhere within the main response and they are swapped in from wherever to wherever. It seems much more conceptually sound to require all oob-swaps to originate adjacent to the response's main element.

Enabling this configuration option is helpful when you want to reuse a small html fragment to render both as as an element within a larger fragment (say a page) as well as more fine grained interactions where the small fragment is updated out of band. By setting `allowNestedOobSwaps: false`, you can place `hx-swap-oob="true"` on the fragment and leave it: when a full page renders then the attribute ignored since the fragment is nested inside the main response, but if you do want to update the fragment out of band then render it adjacent to the to the root response element and htmx performs the oob swap as expected. However in this scenario if `allowNestedOobSwaps: true` (default) then the small fragment is removed from the root element first and then the root element is swapped leaving you confused about why the small fragment vanished.

Here's an example of how this simplifies htmx usage in Go templates. In particular, note how this eliminates the conditional logic that controls whether `hx-swap-oob="true"` is included in the response or not: https://github.com/infogulch/go-htmx/commit/ea227b68fe8d7e6df9996f7d13eb951ed046f1ee

This PR implements the feature request in #1133

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -259,7 +259,13 @@ var htmx = (function() {
         { code: '204', swap: false },
         { code: '[23]..', swap: true },
         { code: '[45]..', swap: false, error: true }
-      ]
+      ],
+      /**
+       * Whether to process OOB swaps on elements that are nested within the main response element.
+       * @type boolean
+       * @default true
+       */
+      allowNestedOobSwaps: true
     },
     /** @type {typeof parseInterval} */
     parseInterval: null,
@@ -1767,9 +1773,14 @@ var htmx = (function() {
    */
   function findAndSwapOobElements(fragment, settleInfo) {
     forEach(findAll(fragment, '[hx-swap-oob], [data-hx-swap-oob]'), function(oobElement) {
-      const oobValue = getAttributeValue(oobElement, 'hx-swap-oob')
-      if (oobValue != null) {
-        oobSwap(oobValue, oobElement, settleInfo)
+      if (htmx.config.allowNestedOobSwaps || oobElement.parentElement.nodeName === 'BODY') {
+        var oobValue = getAttributeValue(oobElement, 'hx-swap-oob')
+        if (oobValue != null) {
+          oobSwap(oobValue, oobElement, settleInfo)
+        }
+      } else {
+        oobElement.removeAttribute('hx-swap-oob')
+        oobElement.removeAttribute('data-hx-swap-oob')
       }
     })
   }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1773,7 +1773,7 @@ var htmx = (function() {
    */
   function findAndSwapOobElements(fragment, settleInfo) {
     forEach(findAll(fragment, '[hx-swap-oob], [data-hx-swap-oob]'), function(oobElement) {
-      if (htmx.config.allowNestedOobSwaps || oobElement.parentElement.nodeName === 'BODY') {
+      if (htmx.config.allowNestedOobSwaps || oobElement.parentElement === null) {
         var oobValue = getAttributeValue(oobElement, 'hx-swap-oob')
         if (oobValue != null) {
           oobSwap(oobValue, oobElement, settleInfo)
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1774,7 +1774,7 @@ var htmx = (function() {
   function findAndSwapOobElements(fragment, settleInfo) {
     forEach(findAll(fragment, '[hx-swap-oob], [data-hx-swap-oob]'), function(oobElement) {
       if (htmx.config.allowNestedOobSwaps || oobElement.parentElement === null) {
-        var oobValue = getAttributeValue(oobElement, 'hx-swap-oob')
+        const oobValue = getAttributeValue(oobElement, 'hx-swap-oob')
         if (oobValue != null) {
           oobSwap(oobValue, oobElement, settleInfo)
         }
--- a/test/attributes/hx-swap-oob.js
+++ b/test/attributes/hx-swap-oob.js
@@ -26,6 +26,32 @@ describe('hx-swap-oob attribute', function() {
   }
 
   for (const config of [{ allowNestedOobSwaps: true }, { allowNestedOobSwaps: false }]) {
+    it('oob swap works when the response has a body tag with config ' + JSON.stringify(config), function() {
+      Object.assign(htmx.config, config)
+      this.server.respondWith('GET', '/test', "<body>Clicked<div id='d2' hx-swap-oob='true'>Swapped0</div></body>")
+      var div = make('<div hx-get="/test">click me</div>')
+      make('<div id="d2"></div>')
+      div.click()
+      this.server.respond()
+      div.innerHTML.should.equal('Clicked')
+      byId('d2').innerHTML.should.equal('Swapped0')
+    })
+  }
+
+  for (const config of [{ allowNestedOobSwaps: true }, { allowNestedOobSwaps: false }]) {
+    it('oob swap works when the response has html and body tags with config ' + JSON.stringify(config), function() {
+      Object.assign(htmx.config, config)
+      this.server.respondWith('GET', '/test', "<html><body>Clicked<div id='d3' hx-swap-oob='true'>Swapped0</div></body></html>")
+      var div = make('<div hx-get="/test">click me</div>')
+      make('<div id="d3"></div>')
+      div.click()
+      this.server.respond()
+      div.innerHTML.should.equal('Clicked')
+      byId('d3').innerHTML.should.equal('Swapped0')
+    })
+  }
+
+  for (const config of [{ allowNestedOobSwaps: true }, { allowNestedOobSwaps: false }]) {
     it('handles more than one oob swap properly with config ' + JSON.stringify(config), function() {
       Object.assign(htmx.config, config)
       this.server.respondWith('GET', '/test', "Clicked<div id='d1' hx-swap-oob='true'>Swapped1</div><div id='d2' hx-swap-oob='true'>Swapped2</div>")
--- a/test/attributes/hx-swap-oob.js
+++ b/test/attributes/hx-swap-oob.js
@@ -1,34 +1,44 @@
 describe('hx-swap-oob attribute', function() {
+  const savedConfig = htmx.config
   beforeEach(function() {
     this.server = makeServer()
+    htmx.config = Object.assign({}, savedConfig)
     clearWorkArea()
   })
   afterEach(function() {
     this.server.restore()
+    htmx.config = savedConfig
     clearWorkArea()
   })
 
-  it('handles basic response properly', function() {
-    this.server.respondWith('GET', '/test', "Clicked<div id='d1' hx-swap-oob='true'>Swapped0</div>")
-    var div = make('<div hx-get="/test">click me</div>')
-    make('<div id="d1"></div>')
-    div.click()
-    this.server.respond()
-    div.innerHTML.should.equal('Clicked')
-    byId('d1').innerHTML.should.equal('Swapped0')
-  })
+  // Repeat the same test to make sure it works with different configurations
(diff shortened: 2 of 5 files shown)


src/htmx.js
```diff
@@ -765,7 +765,7 @@ return (function () {
                     }
                 }
             }
-            forEach(findAll(fragment, '[hx-swap-oob], [data-hx-swap-oob]'), function (oobElement) {
+            forEach(findAll(fragment, 'body > [hx-swap-oob], body > [data-hx-swap-oob]'), function (oobElement) {
```
dkniffin, line 768: Does this work even if the response doesn't have a `<body>` tag?

infogulch: Yes, when the response content is parsed into an html document the `<body>` element is implied and created automatically. (I tried to use `:root > ...` but it didn't work because of the implied body tag).

That said, just recently I was looking at the `useTemplateFragments` config/feature which changes how the response content is parsed; I'm not sure if this change is compatible with that feature. FYI.

alexpetros:
This looks good—could you add tests that validate the new functionality?

infogulch:
I added a test that should show the new behavior, but I'm not sure how to run them.

infogulch:
I updated the test so it should pass if it were to run again.

That said I'm not sure if the behavior is exactly what we want. Specifically, should the final innerHTML contain the `hx-swap-oob` attribute or should it be stripped? The point of this PR is to preserve the `<span>` in this scenario, but my question now is should it preserve the hx attribute?

```js
        btn.innerHTML.should.equal('<button>Clicked <span id="when" hx-swap-oob="true">now!</span></button>');
```

alexpetros:
> Specifically, should the final innerHTML contain the hx-swap-oob attribute or should it be stripped?

What happens for a successful OOB swap? Is the attribute there?

> I updated the test so it should pass if it were to run again.

Unfortunately it does not pass, but you can run it locally with `npm run test`

Also, if you could update the `swap-oob` docs to explicitly say that elements with `swap-oob` not in the root are ignored.

infogulch:
Ok I looked into this further and this PR seems to be in tension with the `'oob swaps can be nested in content'` test.

The question is: What to do when an HX response body contains a nested element with the `hx-swap-oob` attribute? (The case of subsequent root elements with the `hx-swap-oob` attr is already clear.)

1. The status quo's answer is to remove them from their parent element and swap them into the document in parallel with the parent element.
  * The downside is that if you *want* the child element to stay with its parent then you have to strip off the `hx-swap-oob` before sending depending on if you send the parent and child or just the child element, and it's a bit of a pain to plumb conditionals everywhere.
3. My conception, and this PR, would change that behavior to assume that the intention is to keep the nested element together with its parent and ignore the attribute. This lets you reuse nested fragments depending on the need and not have to worry about whether the inner element contains a `hx-swap-oob` attr.
  * The downside is that this changes current behavior.

IMO, `'oob swaps can be nested in content'` seems kinda silly; if you want to do an oob swap then return it at the root, adjacent to the main response. This would mean you could just randomly intersperse parts of the page anywhere within the main response and its just swapped in; I don't see how that's preferable to requiring oob swaps to all be adjacent to the root element.

I wanted to surface this difference and get feedback before I proceed due to the behavior change. Thoughts?

alexpetros:
> It seems much more conceptually sound to require all oob-swaps to originate adjacent to the response's root element, and consider oob-swaps elsewhere to be superfluous and ignore them.

*Very* hard to know if anyone is relying on this behavior, though I agree with you that, conceptually, it's a little absurd.

infogulch:
> _Very_ hard to know if anyone is relying on this behavior

Agreed. So, what are our options?

1. Add a config to toggle between the current and new behavior, which can be either:
  a. Default on
  b. Default off
2. Switch to the new behavior and wait for someone to complain in the next releases, then either:
  a. Revert back and abandon the change
  b. Add a config option (see 1)
  c. Convince them this is for their own good, comrade
3. Break it in htmx 2.0
  a. If someone complains then haughtily explain how bumping the major version in semver means things can break, dumbass
5. Abandon the change now to a great weeping and gnashing of teeth lamenting at the absurdity of life, er, I mean the absurdity of the current behavior. Then shrug and respect code working today even if not perfect

Itzdlg:
I will note that the documentation currently warns: “Out of band elements must be in the top level of the response, and not children of the top level elements.”

Perhaps adding another property like hx-swap-oob=“true with-depth” or a server response header HX-Swap-OOB-Depth: true? Just some way to either reverse new functionality (or to set new functionality to keep backwards compatibility)

infogulch:
This has been paused for a while, but I'm currently making this change as a configuration option to switch behaviors. A 2.0 release might consider changing the default for this configuration option.

alexpetros:
Yeah I apologize about this being stalled, it's a tough one for the reasons identified. We're closing in on having many fewer PRs, which will make it easier to focus on the hard choices. I think a config is a good call.

infogulch:
It's been a bit but I updated this PR to only modify the behavior when the config option is changed to `false`.

I'm not too sure about how to test config options. Saving the config's `before` value, running the validation in a `try` and restoring it in the `finally` at the end seems to work ok.

infogulch:
The description has been updated to add my motivation for opening this PR.

Just off the top of my head here are a couple things that I think would improve it:

- [ ] Update config listing
- [ ] Update docs to reflect how oob swaps work when the config is set to both settings

**Should I continue working on this? What are htmx collaborators' disposition to this change?**

alexpetros:
I like the config option and I'm going to push for it. I'll get you back a final up or down soon.

alexpetros:
A good reason to support this, which I think you're calling out, is that it composes really well with [template fragments](https://htmx.org/essays/template-fragments/)

infogulch:
> A good reason to support this is that it composes really well with [template fragments](https://htmx.org/essays/template-fragments/). You can send the fragment as an OOB swap, or just send the template normally.

Go ahead and just summarize my giant meandering paragraphs in two pithy sentences. It's fine. I'm not mad. Why would I be mad? I'm a great writer I'll have you know. 😡😆😭

alexpetros:
If you weren't a good writer I wouldn't have know what you were talking about ☺️ I have the advantage of not having to also explain all the technical details.

Anyway, got the go-ahead; can you rebase again the 2.0 branch and add the docs?

infogulch:
Great! I'll get it done.

infogulch:
- [x] Rebase changes onto `v2.0v2.0` branch, assuming that's the right one
- [x] Update the configuration listings (both of them, thank you @infogulch for that)
- [x] Add a short explanation of the config option on the `hx-swap-oob` attribute page

Anything else?

alexpetros:
@infogulch ready for review?

infogulch:
@alexpetros yessir

alexpetros (approved):
Thanks for all your hard for on this PR, it's really excellent and we're excited to have it as a part of htmx 2 :)

PR #1249, opened 2023-02-10
Speedup haveSeenNode() for long inputs lists

alexbezhan:
This PR improves performance of `getInputValues` by 100x: from 5s to 50ms on a page that has a lot of form parameters (1k+).

I use Set to compare ids of inputs. If the input has an id, then Set is a lot quicker, than traversing the entire array (if it's long).

See profiling before and after https://gist.github.com/alexbezhan/faa3ef2b975000e2a3b97d098813e3e5.

TLDR: If you want speed, add `id` attribute to your input and you are good.


alexpetros:
@alexbezhan if you're still interested in this, can you rebase it and I'll get it looked at?

alexpetros:
Feel free to rebase and re-open if you're still interested, and we'll get it looked at.

PR #1250, opened 2023-02-11
Convert build system to rollup and support ESM

dkniffin:
I set up a rollup build system and fixed the import system so it works with ES6/ESM. I'm fairly sure everything should still work as it did before. No need for a major version bump, though minor might be a good idea, since it's a fairly large change.

Things to note:
1. To build, use `npm run build` now instead of `npm run dist`
2. Minification is using `terser` now instead of `uglifier`, because `terser` has support for ES6.
3. `import htmx from "htmx.org"` will use ES6 imports now, if available. If not, it will continue using UMD (which has ASM and ES functionality built in).
4. I also set it up so that the code in `src` does not have any of the compiled AMD/UMD/ESM stuff, and just has export default for htmx. 
5. As a result of 4, this means that we can now start to refactor things into separate modules/files, and when we run `npm run build`, rollup will compile it all together into one file (and convert it to the appropriate formats). I really think this is a good idea, as it will make the code MUCH more maintainable and readable. As someone who has gone digging through HTMX code several times, I really hope to see some refactoring like this.
6. Also as a result of 4, we _could_ convert some things to typescript. Rollup has a plugin for converting from typescript, so it should be pretty simple to add that functionality in and start using it, if you we chose to do that. But I know that's being discussed over in #606


dkniffin:
Also, I ran `npm run test` and it looks like it's passing. `npm run test-types` doesn't, but I also don't think it's passing on the master branch.

I don't see a test suite running in CI here, but maybe I'm missing it. That seems like a really good idea to add. A lot of CI tools have free support for OSS projects.

dkniffin:
@1cg Can you take a look at this when you get a chance?

1cg:
hey @dkniffin thank you for taking a look at this for us

i brought this PR/idea up on the discord and there was a lot of discussion around it.  I've shied away from a build step thus far in the name of keeping things simple.  There was one person who suggested vita rather than rollup and a lot of people that felt it wasn't worth it.

for the 1.x line, would it be possible to improve the current system in a less dramatic way?

for 2.x I'm interested in looking into something like this, w/ a debate over the various options in the htmx-2.0 channel on our discord:

https://htmx.org/discord

sorry to not immediately accept what is a good step forward, just trying to keep the project in a place I can get my head around and that gels w/ the communities style...

dkniffin:
As discussed further in Discord, this is the most minimal version of the change here: https://github.com/bigskysoftware/htmx/pull/1263

This PR (1250) adds a build system which does this automatically, and removes it from the original source code. IMHO, that's still a worthwhile change. But strictly speaking, for ESM support, it's not required. And ESM support is what I'm actually interested in. So I'm happy to close this PR if the other gets merged.

1cg:
closing in favor of https://github.com/bigskysoftware/htmx/pull/1263

PR #1254, opened 2023-02-15
Add "empty" swap strategy.

gwk:
This commit adds an "empty" swap strategy that simply sets `target.innerHTML = ''`. In my mind it completes a conceptual 2x2 square of strategies:
```
innerHTML  outerHTML
empty      delete
```

I found this useful when building multistep forms, where I want to clear later steps/results whenever an earlier step is altered. It lets me add a class like "empty-on-step-1" to a number of elements. Then the step-1 htmx response contains an out-of-band element that triggers the empty-swap on all elements with that class: `Div(hx_swap_oob="empty:.empty-on-step-1")`.

If this looks promising to the authors, I am happy to add documentation and tests too.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -924,6 +924,10 @@ return (function () {
             return parentElt(target).removeChild(target);
         }
 
+        function swapEmpty(target, fragment, settleInfo) {
+            target.innerHTML = '';
+        }
+
         function swapInnerHTML(target, fragment, settleInfo) {
             var firstChild = target.firstChild;
             insertNodesBefore(target, firstChild, fragment, settleInfo);
@@ -970,6 +974,9 @@ return (function () {
                     return;
                 case "delete":
                     swapDelete(target, fragment, settleInfo);
+                    return;
+                case "empty":
+                    swapEmpty(target, fragment, settleInfo);
                     return;
                 default:
                     var extensions = getExtensions(elt);


gwk:
Having slept on this, I realize that this is probably redundant with innerHTML in all cases. My misunderstanding was (or perhaps continues to be) around the semantics of the innerHTML swap strategy for in-band vs out-of-band responses. As I understand it, for in-band innerHTML, any number of response elements are swapped into the target contents. For out-of-band, the OOB response element acts as a payload container, rather than getting swapped in as the sole child of the target like I had imagined.

Having played with this some more, I realize that there is consistency in this design but it is more subtle than the documentation lets on. The way I understand it now is:
* IB inner: many replace children.
* IB outer: many replace target.
* OOB inner: children replace children.
* OOB outer: one replaces target.

I had initially imagined the relationship to be more symmetrical, where the rule would be "OOB inner: one replaces children". Perhaps we could add some more examples to https://htmx.org/attributes/hx-swap-oob/?

1cg:
Hi @gwk, thank you for your contribution.  I agree I think this is maybe redundant and that the OOB swap behavior is pretty subtle.

I am actually going to look at this today and try to think of a backwards compatible way to make this more obvious.

gwk:
@1cg for what it's worth, there are two points that I missed and had to figure out experimentally:
* an in-band outerHTML swap can have many elements replacing one (many-replace-one, not one-replace-one).
* an OOB innerHTML swap takes the children of the oob response element, not the whole element (many-replace-many, not one-replace-many).

If I were writing documentation for HTMX I would show these various combinations in the intro. It might be a little verbose, but together they show the power of the system. I'd be happy to take a crack at it if this interests you.

PR #1263, opened 2023-02-21
Update UMD wrapper to include ESM support

dkniffin:
This is a simplified version of #1250 that only adds the ESM support via updating the UMD wrapper. In laymen term: this makes things available to ES6 modules, which are widely used in modern browsers


1cg:
@dkniffin can you retarget for `dev` and I'll merge it?  Thank you!

dkniffin:
@1cg done

Delapouite:
Hi

I'm confused by the title of this PR.
The piece of added coded is only targeting Node's Common.js module system. (`require`/ `module.exports`).
There's no introduction of ESM syntax (`import` / `export`)

@dkniffin Do you intend to provide a follow up PR?

Thanks

mozey:
I've used [esm.sh](https://esm.sh) to convert a few CommonJS (npm packages) to [ESM](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) using [esbuild](https://esbuild.github.io/). 

For htmx the conversion to ESM seems to work fine, but then when I try to bundle my app ([using esbuild with deno](https://deno.land/x/esbuild_deno_loader@0.8.1)), I get [direct-eval](https://esbuild.github.io/link/direct-eval) error
```
Using direct eval with a bundler is not recommended
```

mozey:
For my use case (writing a web app front-end) above is not a problem. I only use `deno` for running tests, and then the `esbuild` plugin to bundle everything for the browser target. Then two script tags, one for htmx, and another for the rest.

My front-end logic consists mostly of custom Alpine directives (see [Hypermedia-Friendly Scripting](https://htmx.org/essays/hypermedia-friendly-scripting/)) for *"invoking library functions"*, and event handlers, written in TypeScript. 

Using the custom directives with HTML files to satisfy *"Locality of Behavior"*, it's not necessary hook into htmx source code (as per #606), from inside my TypeScript app.

PR #1265, opened 2023-02-22
:label: types ajax return

ekwoka:
The types incorrectly labeled the return type of `htmx.ajax` as being `void` when it, in fact, returns a `Promise<void>`

This just fixes that.


ochedru:
Hi! Thank you for making htmx.
Looks like this PR has unfortunately been overwritten by commit 15d4cacbf3c4bf59c09fb4eb78b07dfb3ea3711a. Is it possible to bring this change back? Thanks

Issue #1202, opened 2023-01-14
Possibility to pass some parameters as path variables instead of in query string


jyrimatti:
Hi,

I believe it's more or less an API design issue whether an endpoint accepts parameters in the query string or as path variables. Often both are used. I couldn't find a way to make htmx pass parameters in the path. Is there a way, did I just not find it?

This can be achieved with an extension like this:
```
htmx.defineExtension('path-params', {
    onEvent: function(name, evt) {
        if (name === "htmx:configRequest") {
            evt.detail.path = evt.detail.path.replace(/{{([A-Za-z0-9_]+)}}/g, function (_, param) {
                let val = evt.detail.parameters[param];
                delete evt.detail.parameters[param]; // don't pass in query string since already in path
                return val;
          })
        }
    }
});
```

which would produce a URL `api/2023-01-14.json` from `<a href="api/{{date}}.json" hx-ext="path-params" hx-include="[name='date']">` assuming a field `date` exists. Something similar has been discussed at least in (https://stackoverflow.com/questions/68052027/in-htmx-how-do-i-request-a-specific-url-for-each-option-within-a-select-form-el).

The suggested escape syntax `{{paramName}}` could be something else as well, but at least this form is widely used and forms an invalid URI path.

An extension solves the issue, but since I don't think path parameters differ from query string parameters semantically, should this functionality be bundled in htmx and enabled by default without any extensions or additional attributes?

1cg:
This isn't something I would want to integrate into htmx itself, but you can implement it yourself via the `htmx:configRequest` event:

https://htmx.org/events/#htmx:configRequest

You can look at the `detail.path` for a regex matching your suggestion and replace it with a value in the `detail.parameters` object, then delete it from the parameters object.

jyrimatti:
Hi,

it's a pity, but I understand. If using out-of-own-control APIs becomes a bigger use case in the future, maybe this will be reconsidered.
For now I'll keep using the custom extension. It's quite trivial after all, thanks to the great extension possibilities 👍

lanmaster53:
> Hi,
> 
> I believe it's more or less an API design issue whether an endpoint accepts parameters in the query string or as path variables. Often both are used. I couldn't find a way to make htmx pass parameters in the path. Is there a way, did I just not find it?
> 
> This can be achieved with an extension like this:
> 
> ```
> htmx.defineExtension('path-params', {
>     onEvent: function(name, evt) {
[...]

PR #1307, opened 2023-02-26
add extension to populate path variables with request parameters

jyrimatti:
This PR adds an extension to use request parameters to populate path variables. Used parameters are removed so they won't be sent in the query string or body anymore. Unrecognised variables are just ignored leaving the placeholder as it is.

For example, this:
```html
<div hx-ext="path-params">
    <a hx-get="/items/{itemId}" hx-vals='{"itemId": "42", "foo": "bar"}'>test</div>
</div>
```
would invoke the URL `/items/42?foo=bar`.

Motivation is mainly that it's more or less just an API design choice whether to use query parameters, path variables, or both. When the API is under your own control and a "backend for frontend" kind of an API not meant to be used in other ways, you can of course make the decision to use just query parameters. But when interacting with external or more general APIs also path variables are often needed.

It _might_ be nicer if Htmx supported this out-of-the-box as suggested in #1202 , but this extension provides a usable alternative.

What do you think? Any suggestions for improvements? Any ideas for a better name for the extension?

### DIFF
--- a/src/ext/path-params.js
+++ b/src/ext/path-params.js
@@ -1,0 +1,11 @@
+htmx.defineExtension('path-params', {
+    onEvent: function(name, evt) {
+        if (name === "htmx:configRequest") {
+            evt.detail.path = evt.detail.path.replace(/{([^}]+)}/g, function (_, param) {
+                var val = evt.detail.parameters[param];
+                delete evt.detail.parameters[param];
+                return val === undefined ? "{" + param + "}" : encodeURIComponent(val);
+          })
+        }
+    }
+});
--- a/test/ext/path-params.js
+++ b/test/ext/path-params.js
@@ -1,0 +1,51 @@
+describe("path-params extension", function() {
+    beforeEach(function () {
+        this.server = makeServer();
+        clearWorkArea();
+    });
+    afterEach(function () {
+        this.server.restore();
+        clearWorkArea();
+    });
+
+    it('uses parameter to populate path variable', function () {
+        var request;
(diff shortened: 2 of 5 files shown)


alexpetros:
I don't mind putting this on one the website, because it comes up a lot and I think it's a good demonstration how to use extension to implement some new semantics.

If you rebase it, I'll put it up as a candidate to be merged.

ddevienne:
The motivating example above uses hard-coded values. Is it possible to get them from a `<form><input>` instead?
(w/o Javascript, using only HTMX). The HTML I serve is to demonstrate a JSON API (barebone Postman / Swagger UI,
but w/o the OpenAPI spec), and I'd like users to enter route params via forms. Is that possible?

Perhaps via a `#name` syntax, that would read the value from `<input id="name"/>` ?

jyrimatti:
> The motivating example above uses hard-coded values. Is it possible to get them from a `<form><input>` instead? (w/o Javascript, using only HTMX). The HTML I serve is to demonstrate a JSON API (barebone Postman / Swagger UI, but w/o the OpenAPI spec), and I'd like users to enter route params via forms. Is that possible?
> 
> Perhaps via a `#name` syntax, that would read the value from `<input id="name"/>` ?

Yes, of course. The example isn't a realistic use case, just an easy example. The extension uses the data provided by Htmx, wherever it gets it (form fields, hx-vals...). It justs "moves" a parameter to the path if there's an appropriate placeholder present.

PR #1330, opened 2023-03-16
Fix for memory leak

croxton:
Nullify unreachable references to dom nodes inside `setTimeout()` calls, to allow garbage collection to release memory once the nodes have been swapped out of the dom. This helps prevent detached nodes from accumulating after multiple swaps without a browser refresh, and results in lower memory use (in my testing).


1cg:
Can you retarget to `dev` for me?  Thank you!

croxton:
Done!

1cg:
Thank you!

PR #1331, opened 2023-03-16
Trigger event when a SSE-source is opened

### LINKED ISSUE #1325: SSE onopen event
Hey,

I have a webapp that needs some kind of connection-tracking (online-offline). I use the SSE ext. for triggering updates in my frontend.

On connection loss, I open a modal like this:

```
  htmx.on("htmx:sseError", function(evt){
                        console.log("SSE disconnect");
                        htmx.addClass(htmx.find("#disconnect-modal"), "is-active");
```

It would be great to also have a lifecycle event for reconnection or connection at all. I tried to access the `eventSource` object from the internal data to add a callback to `onopen` but it does not work.

How about adding an event `htmx:sseOpen` to the sse extension?

hirschenberger:
fixes #1325

### DIFF
--- a/src/ext/sse.js
+++ b/src/ext/sse.js
@@ -155,6 +155,10 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 				}, timeout);
 			}			
 		};
+
+		source.onopen = function (evt) {
+			api.triggerEvent(elt, "htmx::sseOpen", {source: source});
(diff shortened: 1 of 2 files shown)


1cg:
@benpate can you approve this change (lgtm)

benpate:
i'm looking at this now.

PR #1358, opened 2023-04-11
Untrack non-top-level node_modules directories

waldyrious:
Currently the repo has ~200MB in the `www/static/test/` directory, most of it from a bunch of `node_modules` directories, whose contents are essentially the same (so the repo contains multiple copies of the same files).

The line `/node_modules` currently in the `.gitignore` only catches the `node_modules` directories at the top level of the repository. That line was [added back in 2020](https://github.com/bigskysoftware/htmx/commit/9e8f101c0f425f56955cab0e7df11232c9d53ec5), though, so I suspect it wasn't deliberate that it doesn't capture the non-top-level `node_modules` folders within `www/`.

This PR removes the leading `/` so that it will match `node_modules` directories at any level of the repository file tree.

For comparison, the `www/` folder was at 227.6 MB before, and is now 72.2 MB (specifically, the `www/static/test/` directory was reduced from 203.0 MB down to 55.2 MB).


waldyrious:
Btw, @danieljsummers — I noticed that you introduced a [separate `www/.gitignore` file](https://github.com/bigskysoftware/htmx/blob/v1.9.0/www/.gitignore) in #1332. While this could be useful to separate concerns if the two gitignore files contained a lot of patterns, I think that with the current size of both files it's more ergonomic and intuitive to just have the top-level one. WDYT about moving the single pattern from `www/.gitignore` into the top-level one?

danieljsummers:
@waldyrious That was part of the Zola template; it doesn't matter to me where the exclusion rules live. I do think you may find a bit of pushback on this PR, though, as those `node_modules` directories allow each version of htmx's tests to be run with the versions of the dependencies as they were at the time it was released. Preserving that was what made #1332 so astoundingly huge.

Unless Carson's OK'd this already, you may want to discuss it with him over in Discord. _(or here; again, I have no strong opinions on location... :) )_

waldyrious:
Thanks for the quick response. Carson has [replied on Discord](https://discord.com/channels/725789699527933952/1079191774976802867/1095414283740917820), so I continued the conversation there. I'm hoping we can replace the vendored scripts with CDN links. The considerable reduction in the repo size seems worth it.

waldyrious:
For convenience, here are condensed quotes from the discussion on Discord:

@1cg
> unfortunately those are kinda correct
> I include the specific version of sinon/chai/etc. for each version I release, so that they don't break old tests when they invariably release new versions
> so we effectively vendor per version

@waldyrious
> Oh, I think I get it. The `test/<version>/test/*.html` files use those scripts directly with `<script src="../node_modules/chai/chai.js"></script>` and so on.
> I wonder if it wouldn't be better if those were CDN links? That would avoid polluting the git repo, and it would make it explicit which version of each script is being used.
> There's a downside of having to be online to run the tests, but other than that, the pros seem to outweigh the cons IMHO.
> Something like this: 
> image

@1cg
> we could use CDN links, I've shied away from it for mumble, mumble
> I should rewrite the whole thing but I'm lazy
> manually going back and CDN-ing them would probably be a good move
> Although I'd like to keep the latest vendor'd for some reason

waldyrious:
I'm up for doing those changes (replacing the scripts with CDN links), but I won't be able to complete them very soon. Probably this weekend or the next one. If anyone wants to take over in the meantime, let me know and I'll give write access to my fork (or just start anew with a different PR).

Btw, shouldn't the CI in this PR exercise at least part of those tests? Removing the `node_modules` should've broken _something_.

1cg:
closing for now, this is an issue that we need to address at some point

waldyrious:
@1cg are you still unsure about the CDN approach (for the non-latest versions)? I haven't had the chance to work on this since my last comment, but I am still interested in doing the changes if there are no objections.

PR #1364, opened 2023-04-11
Normalize whitespace in base website HTML template

waldyrious:
- Make all indentation consistent at 4 spaces
- Remove two spurious `<div>` tags
- Collapse `<div>`-wrapped links that fit in a single line
- Split a `<label>` with nested tags within into multiple lines

### DIFF
--- a/www/themes/htmx-theme/templates/base.html
+++ b/www/themes/htmx-theme/templates/base.html
@@ -34,25 +34,18 @@
             <div id="nav" class="navigation" hx-boost="true">
             {% endif %}
                 <div class="navigation-items" preload="mouseover">
-                    <div>
-                        <a href="/docs/">docs</a>
-                    </div>
-                    <div>
-                        <a href="/reference/">reference</a>
-                    </div>
-                    <div>
-                        <a href="/examples/">examples</a>
-                    </div>
-                    <div>
-                        <a href="/talk/">talk</a>
-                    </div>
-                    <div>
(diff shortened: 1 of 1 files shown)


gnat:
Thanks but please tone down the number of zero-substance PR's to this repo- you've had a few already.

You're taking developers time away from doing legitimate work in order to review this.

dz4k:
do you think it would be a good idea to set up an auto-formatter for the repo, and do you know a good one?

waldyrious:
> Thanks but please tone down the number of zero-substance PR's to this repo- you've had a few already.

I'm not sure what you consider zero-substance, but even considering all my PRs to this repo, the interval between them is measured in months, in average. I don't think that amounts to anything that would prevent maintainers from doing "legitimate" work. Either way, please reconsider the way you respond to contributions sent in good faith.

waldyrious:
> do you think it would be a good idea to set up an auto-formatter for the repo, and do you know a good one?

Yes, definitely. I did use HTMLHint several years ago, but I'm not sure it's actively maintained these days. Possibly something like eslint or prettier would be a better choice. I have a personal preference towards eslint because it's more configurable, but Prettier should be easier to set-it-and-forget-it if we don't care about the specifics.

waldyrious:
Ok, I can tell you're not willing to have a good faith discussion here. I'm not interested in exchanging snarky remarks. Cheers.

PR #1371, opened 2023-04-12
Fix swapError with hx-on and innerHTML swap

SamDudley:


Telroshan:
Hey, good catch as the for of loop [isn't supported by IE11](https://caniuse.com/?search=for...of) anyway

This loop seems to have [already been replaced](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L922) with a standard for loop since then, which should probably fix the issue.

Could you update your branch against the `dev` branch? Even if you don't need to update the core code anymore, an extra test is still worth taking!

SamDudley:
Hey - yes the issue has been fixed and released.

I've updated the branch against the `dev` branch to include the additional test.

Telroshan:
Ok so it took 3 runs but the test pipeline passed, so you don't have to worry about it, some random failures happen, we'll have to fix that

Issue #1062, opened 2022-09-28
CSP issue with IndicatorStyles - needs a htmx.config.inlineStyleNonce config option


Kimotu:
Please add config option ` htmx.config.inlineStyleNonce:''` and extend function `insertIndicatorStyles` to include:

```
"<style nonce=\"" + htmx.config.inlineStyleNonce + "\">\
                      ." + htmx.config.indicatorClass + "{opacity:0;transition: opacity 200ms ease-in;}\
                      ." + htmx.config.requestClass + " ." + htmx.config.indicatorClass + "{opacity:1}\
                      ." + htmx.config.requestClass + "." + htmx.config.indicatorClass + "{opacity:1}\
                    </style>");
```

Tested it in a locale modified installation. Works.

jyrimatti:
I also ran into this issue. 
For example when a recommended Content Security Policy is used:
`<meta http-equiv="Content-Security-Policy" content="default-src 'self'" />`
no runtime style tags are allows to be inserted to the document (without a known hash/nonce).

This behaviour can be disabled by using
`<meta name="htmx-config" content='{ "includeIndicatorStyles": false }' />`
but it feels kind of awkward to add this everywhere.

Is there a reason why this style is added programmatically, instead of providing a separate htmx.css file that the users could include if needed?

terrablue:
Would you accept a PR here? I needed to add `style-src 'unsafe-inline';` to my CSP to temporarily resolve this issue, but it loosens up the default security. I'd go with either of the proposed solutions here, though I'd prefer the CSS solution.

1cg:
fixed

PR #1376, opened 2023-04-13
solve CSP issue with indicator style via opt-in

terrablue:
Hi, this solves https://github.com/bigskysoftware/htmx/issues/1062 by allowing the user to explicitly specify indicator classes via CSS variables. A CSS file is included for illustration but is not strictly necessary.

The new mechanism is an opt-in, it is used only if the user specifies class names using both variables, otherwise the previous mechanism that injects styles into the page is used. It allows willing users to circumvent the CSP issue.

### DIFF
--- a/src/htmx.css
+++ b/src/htmx.css
@@ -1,0 +1,13 @@
+:root {
+  --indicator-class: "htmx-indicator";
+  --request-class: "htmx-request";
+}
+
+.htmx-indicator {
+  opacity: 0;
+  transition: opacity 200ms ease-in;
+}
+
+.htmx-indicator .htmx-request, .htmx-indicator.htmx-request{
+  opacity: 1;
+}
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3494,12 +3494,21 @@ return (function () {
 
         function insertIndicatorStyles() {
             if (htmx.config.includeIndicatorStyles !== false) {
-                getDocument().head.insertAdjacentHTML("beforeend",
-                    "<style>\
-                      ." + htmx.config.indicatorClass + "{opacity:0;transition: opacity 200ms ease-in;}\
-                      ." + htmx.config.requestClass + " ." + htmx.config.indicatorClass + "{opacity:1}\
-                      ." + htmx.config.requestClass + "." + htmx.config.indicatorClass + "{opacity:1}\
-                    </style>");
+                var customIndicatorClass = getComputedStyle(document.body).getPropertyValue("--indicator-class");
+                var customRequestClass = getComputedStyle(document.body).getPropertyValue("--request-class");
+                if (customIndicatorClass !== "" && customRequestClass !== "") {
+                  // override custom indicator and request classes
+                  htmx.config.indicatorClass = customIndicatorClass;
+                  htmx.config.requestClass = customRequestClass;
+                } else {
+                  // previous behavior
+                  getDocument().head.insertAdjacentHTML("beforeend",
+                      "<style>\
+                        ." + htmx.config.indicatorClass + "{opacity:0;transition: opacity 200ms ease-in;}\
+                        ." + htmx.config.requestClass + " ." + htmx.config.indicatorClass + "{opacity:1}\
+                        ." + htmx.config.requestClass + "." + htmx.config.indicatorClass + "{opacity:1}\
+                      </style>");
+                }
             }
         }
 


Telroshan:
Hey, I must admit I'm a bit confused
Feels kinda hacky to me to use a CSS variable to define a CSS class name to use, especially given that unless I'm missing something, this can already be achieved with the current state of the lib : htmx's [own config properties](https://htmx.org/api/#properties) here, that are `indicatorClass` and `requestClass` should do the trick, set `includeIndicatorStyles` to false and roll your own CSS for those indicator classes
Why would you use those CSS variables instead of configuring htmx as with any other configuration property?

terrablue:
Hi, the idea here is to make the default behavior compliant with a strict CSP requirement, by extracting the CSS part into a file for which a nonce/hash can then be generated.

I'm not directly using HTMX, but offering it as a frontend option as part of a framework with a default strict CSP requirement, so the current workaround is to turn off that strict requirement for `style-src`, which isn't really clean. Rolling my own CSS would amount to essentially this patch, which would duplicate HTMX's internal code and would deviate from it down the line in case it changes.

I also wouldn't mind a solution that involves providing nonces, but I feel the current situation isn't very compatible with a secure approach.

Telroshan:
Tell me if I'm wrong but from what I read in this PR:
- If you define both `--indicator-class` and `--request-class` to whatever values you want
  - htmx will NOT append the inline `<style>` element to the head
  - htmx will use the custom defined request class as the class to add to indicator elements when a request is fired
  - it's up to the user to then provide some CSS to style the indicators
  - _(htmx won't do anything with `--indicator-class` as  `htmx.config.indicatorClass` is only used in that `insertIndicatorStyles` function anyway)_

What I'm asking, is how is that different, and what added value does it bring compared to what you can already do with the standard configuration? i.e.
- If you define `htmx.config.includeIndicatorStyles` to false, and `htmx.config.requestClass` to whatever value you want
  - htmx will NOT append the inline `<style>` element to the head
  - htmx will use the custom defined request class as the class to add to indicator elements when a request is fired
  - it's up to the user to then provide some CSS to style the indicators

I may be missing something, but I don't see how those 2 ways (the one you're suggesting here, and the one already possible within the lib) differ in their respective result

terrablue:
The result is the same, but is friendlier to a secure CSP, that is the whole point. The difference is that with the PR, the default would be to use the CSS approach if the user (in my case the framework) includes the CSS file. The user might not want to have a different style at all, but use the default one that HTMX provides.

If kept as is, the framework would need to ship an overlay configuration of HTMX with `htmx.config.includeIndicatorStyles` set to `false` and  `htmx.config.requestClass` set to something just to be aligned with current default styles in JS -- that's somewhat too opinionated for me. I'd like to keep the frontend configuration entirely to the user.

It also means that if the in-JS styles ever change, I would need to keep track of that, essentially duplicating code. 

Note that I had originally wanted to remove the else path (the original code) completely, but opted to provide a path forward that is as compatible as possible with older setups. Perhaps that is what is causing the confusion here?

On a general note, it makes more sense to me that the CSS is extracted into its own file. On the long run and if you decide to go forward with this, I think the else path could be entirely removed.

If this isn't convincing, would you be better aligned with a solution that allows supplying nonces via a HTMX configuration variable?

Telroshan:
Ok I think I understand your point, but I'm still confused with some arguments
> the default would be to use the CSS approach if the user (in my case the framework) includes the CSS file.

> If kept as is, the framework would need to ship an overlay configuration of HTMX with htmx.config.includeIndicatorStyles set to false and htmx.config.requestClass set to something just to be aligned with current default styles in JS -- that's somewhat too opinionated for me

`htmx.config.requestClass` already has a default value set to `htmx-request` so you wouldn't need to define it if you want to rely on the default style.
Also, if your framework can ship a default CSS file along, why couldn't it ship a [meta tag](https://htmx.org/docs/#config) / a script to override some of the default htmx config?

> I'd like to keep the frontend configuration entirely to the user.

I'm not sure I understand this stance, as [htmx configuration](https://htmx.org/docs/#config) is already available to the user. You may either configure with a meta tag (so from your HTML) or override it from a JS script

> On a general note, it makes more sense to me that the CSS is extracted into its own file

While I agree theoretically, as in I don't ever embed CSS in my own JS scripts, the thing is that it would suddenly require people to not only include a single JS file to use htmx, but also a CSS file along. I must admit I'd be skeptical about the gains compared to the confusion it might bring to beginners that are getting started.

You're completely right about the retro compatibility, as htmx strives to always be retro-compatible unless a breaking change happens (htmx 2.0 will do that), so thanks for providing a if-else path to avoid breaking older setups

> If this isn't convincing, would you be better aligned with a solution that allows supplying nonces via a HTMX configuration variable?

I'm not familiar at all with nonces, so maybe is that where the confusion comes from and I'm completely missing the point? I thought that would be something you handle from the backend, not the frontend, so maybe am I off the mark here

terrablue:
> Also, if your framework can ship a default CSS file along, why couldn't it ship a [meta tag](https://htmx.org/docs/#config) / a script to override some of the default htmx config?

That's the 'opinionated' part. Shipping a default HTMX CSS file is one thing, but overriding the HTMX defaults to align it with CSP is another.

> I'm not sure I understand this stance, as [htmx configuration](https://htmx.org/docs/#config) is already available to the user. You may either configure with a meta tag (so from your HTML) or override it from a JS script

Sure, I was talking about the framework not needing to do this for the user, essentially leaving HTMX as is.

> While I agree theoretically, as in I don't ever embed CSS in my own JS scripts, the thing is that it would suddenly require people to not only include a single JS file to use htmx, but also a CSS file along. I must admit I'd be skeptical about the gains compared to the confusion it might bring to beginners that are getting started.

If we include the fallback, the user doesn't need to include the CSS file, especially if he doesn't care about a strict CSP.

> I'm not familiar at all with nonces, so maybe is that where the confusion comes from and I'm completely missing the point? I thought that would be something you handle from the backend, not the frontend, so maybe am I off the mark here

A nonce is basically a way to whitelist inline scripts. It is generated on the server and then included as `<style nonce="value"></style>`. It's also possible to generate a hash for an external CSS file, which is what the framework would do if the solution in the PR is accepted.

Telroshan:
Ok I understand, thanks!
So it seems we should just agree to disagree on this! 😆 Over all, this CSS variables trick seems too hacky to me _(i.e. CSS variables not used as they are intended to i.e. to define a style value, but here to define what in my opinion should be an attribute value / config value)_, and in your case I would rely on overriding the configuration as it's meant to be configured
But again, not pretending this is the truth, just my personal opinion on the matter.

As I've not even been a maintainer for 24 hours yet, I won't take the liberty of taking a decision about this PR, especially given that it seems to involve personal tastes in the equation

Thoughts on this @1cg @alexpetros ?

Telroshan:
Until we get other insights, could you please retarget your PR to the `dev` branch as per the [contributing guidelines](https://github.com/bigskysoftware/htmx/blob/master/CONTRIBUTING.md) ?

Issue #1040, opened 2022-09-10
HX-Reselect in addition to HX-Retarget and HX-Reswap


BoPeng:
Many thank for implementing `HX-Reswap` (#933 and #949). However, could you also add `HX-Reselect` to modify the original `hx-select`? I have in my oroginal html

```
hx-post='/url' hx-swap='innherHTML' hx-target='some-div' hx-select='.some-content'
```
when an error happens, the backend tries to trigger a differnt workflow using `HX-Retarget` and `HX-Reswap`. It puzzled me for quite some time when the new content disappears after swap, and then I realized that the old `hx-select` is still in effect, which removes the returned content since it does not have `.some-content`.

So `HX-Reselect` is last missing piece of the `HX-Re` series.

nwalters512:
I found this when I searched for `HX-Reselect` in the docs and didn't see anything - this is exactly what I've been looking for! Thanks for opening a PR 🚀

1cg:
coming in next release

PR #1385, opened 2023-04-17
Implementation of HX-Reselect

RoToRa:
See #1040


CleitonDeLima:
@1cg  Will this be included in a release?

PR #1427, opened 2023-05-07
docs: updating attribute table to use beforeend instead of beforeEnd

paganotoni:
I noticed in the table shown in the reference index page was using `beforeEnd` instead of `beforeend` this could lead to confusion among the developers. The hx-swap detailed view shows it correctly. This small PR changes it to `beforeend`.

BTW, I'm loving htmx: it's just amazing!


1cg:
:) glad you are enjoying it, thank you for your contribution!

PR #1428, opened 2023-05-08
Type definition for defaultFocusScroll

mozey:


mozey:
Assuming this is a *"docs fixes can be made directly against master"* as per [contributing](https://github.com/bigskysoftware/htmx#contributing)

PR #1436, opened 2023-05-12
response-targets extension with documentation, exposed 3 functions in internal API

siefca:
The response-targets extensions allows to specify different target elements to be swapped when
different HTTP response codes are received.

Example:

```html
<div hx-ext="response-targets">
    <div id="response-div"></div>
    <button hx-post="/register"
            hx-target="#response-div"
            hx-target-5*="#serious-errors"
            hx-target-404="#not-found">
        Register!
    </button>
    <div id="serious-errors"></div>
    <div id="not-found"></div>
</div>
```


RafaelZasas:
Has this been updated in the htmx documentation?
If this has been added officially, I would like to add this to the [HTMX LSP](https://github.com/ThePrimeagen/htmx-lsp/tree/master)  thats being worked on.

PR #1440, opened 2023-05-14
We're so back...

edezhic:
Ref - https://twitter.com/htmx_org/status/1657773466794471424?t=WDJT_HLfz1RSITsNdoKE_g&s=19


1cg:
it's so over

PR #1451, opened 2023-05-24
hx-boost and hx-verb do not consider attribute value changes at runtime

xhaggi:
At the moment htmx does not consider changes to the `href`, `action` and `method` attributes for boosted elements at runtime, nor does it consider changes to the hx-verb attributes (`hx-get`, `hx-post`, etc.). With these code changes, you can adjust the values of these attributes at runtime and htmx will take them into account during the request.

Furthermore, overriding attributes for form submission (`method` and `action` via `formmethod` and `formaction`) is now also supported. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attributes_for_form_submission.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1327,8 +1327,18 @@ return (function () {
                             path = elt.href; // DOM property gives the fully resolved href of a relative link
                         } else {
                             var method = getRawAttribute(elt, "method") || "get";
+                            var action = getRawAttribute(elt, "action");
+                            // support for overriding method and action
+                            if (evt.submitter) {
+                                if (evt.submitter.hasAttribute('formmethod')) {
+                                    method = evt.submitter.getAttribute('formmethod');
+                                }
+                                if (evt.submitter.hasAttribute('formaction')) {
+                                    action = evt.submitter.getAttribute('formaction');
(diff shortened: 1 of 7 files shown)


ekwoka:
I just made a PR that fixes an issue caused by the same kind of thing: #1820 

Basically, on every processed element it does a ton of upfront work instead of just lazily doing it when it's needed

src/htmx.js
```diff
@@ -1319,19 +1319,27 @@ return (function () {
...
-                    path = elt.href; // DOM property gives the fully resolved href of a relative link
-                } else {
-                    var rawAttribute = getRawAttribute(elt, "method");
-                    verb = rawAttribute ? rawAttribute.toLowerCase() : "get";
-                    if (verb === "get") {
-                    }
-                    path = getRawAttribute(elt, 'action');
-                }
                 triggerSpecs.forEach(function(triggerSpec) {
                     addEventListener(elt, function(elt, evt) {
```
ekwoka, line 1323: could this listener function now be removed from being inlined so it can be reused on all elements and not recreated for each? It doesn't seem to reference anything from the closure.

PR #1453, opened 2023-05-24
Support event names with "." character

xhaggi:
Unlike `hx-trigger`, `hx-on` does not support event names that contain dots.

```html
<div hx-get="/test" hx-trigger="my.custom.event" hx-on="my.custom.event: alert('Fired!')"></div>
```

Instead htmx throws a SyntaxError:
```
Uncaught SyntaxError: Unexpected token ':'
```


xhaggi:
Done.

1cg:
great change, thank you!

Issue #407, opened 2021-03-09
Scroll to top by default when navigating with hx-boost


jens:
When a link is boosted it would be nice to automatically be at the top of the page as one would be if it was a non boosted link. Currently the scroll position will be the same as it was on the page before navigating.

It is easy to configure with hx-swap="innerHTML show:top" but should probably be the default behaviour.

jens:
One disadvantage with using "show:top" is that all anchor links breaks. Instead of going to the anchor it will go to the top of the page.

1cg:
Thank you for contributing!

David-Guillot:
Hi there,

This is a nice default option but I see no way to get the original behavior back: what if I want no scroll at all?

David-Guillot:
Okay there is a (not perfect) solution: appending a `#` to the link URL. I think `show:noop` would be preferable, what do you think?

1cg:
agree

1cg:
looking at the code, it looks like

```
hx-swap="show:unset"
```
should work.  I will document this as a supported option.

David-Guillot:
With v1.4.1, `show:unset` doesn't work. I've seen a discussion [here](https://github.com/bigskysoftware/htmx/pull/444) about some `unset` behavior that would be similar, but unmerged yet.

boxed:
I found this issue while trying to look if there was a bug report on the scrolling to top. I think this idea was merged without thinking through the consequences fully. If you want it to scroll to the top, why not just use a normal link instead of boost?

1cg:
The idea is to have boosted links behave as much like "normal" links as possible, and normal links scroll to the top.  If you want custom scrolling behavior, you can use a `hx-get` pretty easily.

peterhry:
Possible to disable the smooth scrolling behaviour?

nwalters512:
This bit me today as I was getting started with HTMX for the first time. I was trying to make inline editing on a large table avoid reloading the whole page by adding `hx-boost="true"` to the form. It works perfectly _except_ for the fact that it scrolls to the top, which loses the context on the page, which is what I was trying to avoid by adopting HTMX in the first place! Adding `hx-swap="show:unset"` didn't help for me either.

[...]

PR #1455, opened 2023-05-25
Allow hx-swap options to be used without specifying modifier (innerHTML remains the default)

xhaggi:
This makes the swap style `innerHTML`, `outerHTML` etc. in `hx-swap` optional and improves the handling of unknown values.

@David-Guillot you have requested something like that in https://github.com/bigskysoftware/htmx/issues/407#issuecomment-876543782

### DIFF
--- a/www/content/attributes/hx-swap.md
+++ b/www/content/attributes/hx-swap.md
@@ -3,11 +3,12 @@ title = "hx-swap"
 +++
 
 The `hx-swap` attribute allows you to specify how the response will be swapped in relative to the
-[target](@/attributes/hx-target.md) of an AJAX request.
+[target](@/attributes/hx-target.md) of an AJAX request. If you do not specify the option, the default is
+`htmx.config.defaultSwapStyle` (`innerHTML`).
 
 The possible values of this attribute are:
 
-* `innerHTML` - The default, replace the inner html of the target element
+* `innerHTML` - Replace the inner html of the target element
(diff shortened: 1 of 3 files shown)


alexpetros:
@xhaggi if you're still interested, can you rebase and add some tests?

xhaggi:
Sure and done. Ah, and tests already exist, or do you mean more of them?

xhaggi:
>Am I correct that what this is essentially doing is making innerHTML the default swap style? If so, could you update the documentation to indicate that?

This is already part of this PR.
https://github.com/bigskysoftware/htmx/pull/1455/files#diff-33bf130564ad635cfa3ad8fa580ac93fa8c8bca856dfb656f5ab6682059c8a7bR6-R7

www/content/attributes/hx-swap.md
```diff
@@ -2,12 +2,13 @@
 title = "hx-swap"
 +++
 
-The `hx-swap` attribute allows you to specify how the response will be swapped in relative to the
-[target](@/attributes/hx-target.md) of an AJAX request.
+The `hx-swap` attribute allows you to specify optionally how the response will be swapped in relative to the
```
alexpetros, line 5: No need to say "specify optionally" here, I think just leaving it as "allows you to specify" is cleaner.

xhaggi: Done.

1cg:
great change, thank you!

PR #1456, opened 2023-05-25
Add support for event cancellation (preventDefault) in extensions

xhaggi:
As the title says, this PR adds support to cancel an event also with `preventDefault()` within `onEvent` of an extension.


alexpetros:
@xhaggi can you add a test for this? Would like to get it in the next release

xhaggi:
Sure. I have now added 2 tests that ensure you can use `return false` as well as `preventDefault` to cancel events in extensions.

PR #1459, opened 2023-05-26
Add config option to disable scroll into view for targets of boosted links and forms

xhaggi:
This adds a configuration option to enable/disable scrolling up after swapping a boosted link or form.

Related discussion https://github.com/bigskysoftware/htmx/issues/407

@David-Guillot you have requested something like that in https://github.com/bigskysoftware/htmx/issues/407#issuecomment-876543782

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -74,7 +74,8 @@ return (function () {
                 getCacheBusterParam: false,
                 globalViewTransitions: false,
                 methodsThatUseUrlParams: ["get"],
-                selfRequestsOnly: false
+                selfRequestsOnly: false,
+                scrollIntoViewOnBoost: true
             },
             parseInterval:parseInterval,
             _:internalEval,
@@ -2641,7 +2642,7 @@ return (function () {
                 "swapDelay" : htmx.config.defaultSwapDelay,
                 "settleDelay" : htmx.config.defaultSettleDelay
             }
-            if (getInternalData(elt).boosted && !isAnchorLink(elt)) {
+            if (htmx.config.scrollIntoViewOnBoost && getInternalData(elt).boosted && !isAnchorLink(elt)) {
               swapSpec["show"] = "top"
             }
             if (swapInfo) {
--- a/www/content/api.md
+++ b/www/content/api.md
@@ -121,6 +121,7 @@ Note that using a [meta tag](@/docs.md#config) is the preferred mechanism for se
 * `withCredentials:false` - boolean: allow cross-site Access-Control requests using credentials such as cookies, authorization headers or TLS client certificates
 * `wsReconnectDelay:full-jitter` - string/function: the default implementation of `getWebSocketReconnectDelay` for reconnecting after unexpected connection loss by the event code `Abnormal Closure`, `Service Restart` or `Try Again Later`
 * `scrollBehavior:smooth` - string: the behavior for a boosted link on page transitions. The allowed values are `auto` and `smooth`. Smooth will smoothscroll to the top of the page while auto will behave like a vanilla link.
+* `scrollIntoViewOnBoost:true` - boolean: whether or not the target of a boosted element is scrolled into the viewport. If `hx-target` is omitted on a boosted element, the target defaults to `body`, causing the page to scroll to the top.
 
 ##### Example
 
--- a/www/content/attributes/hx-swap.md
+++ b/www/content/attributes/hx-swap.md
@@ -112,6 +112,15 @@ You may also use `window:top` and `window:bottom` to scroll to the top and botto
   </div>
 ```
 
+For boosted links and forms the default behaviour is `show:top`. You can disable it globally with
+[htmx.config.scrollIntoViewOnBoost](@/api.md#config) or you can use `hx-swap="show:none"` on an element basis.
+
+```html
+<form action="/example" hx-swap="show:none">
+  ...
+</form>
+```
+
 #### Focus scroll
 
 htmx preserves focus between requests for inputs that have a defined id attribute. By default htmx prevents auto-scrolling to focused inputs between requests which can be unwanted behavior on longer requests when the user has already scrolled away. To enable focus scroll you can use `focus-scroll:true`.


Telroshan:
Hey it's me, the annoying semantic guy
I'm just wondering if `boostedShowTop` is the right choice of name here ; I know the swap spec uses the keyword "show" internally, but the user never gets to see that. I don't find that name super intuitive at the first glance to guess what it's doing.
Why not use the term scroll instead of show here, such as `boostedScrollTop` ?

Though looking at the existing config, maybe a name such as `scrollBoostedTop` or `scrollBoostedElementsTop` would be more consistent? (as we have `refreshOnHistoryMiss`, `includeIndicatorStyles`, `allowEval` for example, where the verb is the first word of the config property)
I'm suggesting this one here because `swapSpec["show"] = "top"` resolves to [a `scrollIntoView` call](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L2700)

Let me know your thoughts on this

sjc5:
I would love to see this get merged! One great thing about hx-boost is that you can, if you choose, basically just always target the full document, and get what you might call an "automatic" SPA user experience on top of a totally traditional MPA developer experience. Although not totally in line with HTMX's primary use case of using smaller, granular html snippets, I think this is also a really valid use case for a lot of sites. This is what we're doing by default with [Hwy](https://hwy.dev), for example.

At risk of overthinking and delaying merging, I'd love to see this take options:

```ts
scrollTopAfterBoosted: true | false | "anchor-only" | "form-only"
```

xhaggi:
I don't have a strong opinion on the name I've chosen, so I'll adapt it to your suggestion. Will do it tomorrow.

1cg:
should we allow for more than just a global override, or is global good enough?  (Thank you for your work on this @xhaggi )

sjc5:
> should we allow for more than just a global override, or is global good enough? (Thank you for your work on this @xhaggi )

I think global override is a good start. But yes, I think it would be great to also be able to override on a per-form / per-anchor basis. The main use case for me, just for the sake of being explicit, is to be able to return form validation results in what might be a long form, without scroll state being lost.

alexpetros:
I think we should use `scrollIntoView`, because that aligns with the [JS name](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView). I also don't know why we would limit this to boosted links. I'd suggest that we do `htmx.scrollIntoView` which by default scrolls the non-OOB element into view on all swaps.

We also probably will want a granular way to enable/disable that (and add it to OOB swaps), but that's no reason to hold up the global config.

sjc5:
> I think we should use `scrollIntoView`, because that aligns with the [JS name](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView). I also don't know why we would limit this to boosted links. I'd suggest that we do `htmx.scrollIntoView` which by default scrolls the non-OOB element into view on all swaps.
> 
> We also probably will want a granular way to enable/disable that (and add it to OOB swaps), but that's no reason to hold up the global config.

Based on some background discussions I read earlier (sorry don't have the links handy), the limitation to boosted forms and anchors is to mimic default browser behavior (though the default smooth scroll violates this). Personally I just want a way to set it to no scrolling at all, to avoid the UI slingshotting effect when you submit a form, so other than that have no strong opinion :)

alexpetros:
> I would love to see this get merged! One great thing about hx-boost is that you can, if you choose, basically just always target the full document, and get what you might call an "automatic" SPA user experience on top of a totally traditional MPA developer experience

Unrelated, but isn't this just a regular link?

sjc5:
> > I would love to see this get merged! One great thing about hx-boost is that you can, if you choose, basically just always target the full document, and get what you might call an "automatic" SPA user experience on top of a totally traditional MPA developer experience
> 
> Unrelated, but isn't this just a regular link?

Yep, regular anchor tag with hx-boost on for automatic client routing. Astro just recently released their client routing solution to get this same effect.

1cg:
ok, so https://htmx.org/attributes/hx-swap/ is where you can specify scrolling behavior

by default it does nothing, unless we are taking about a boosted link or form, so this global really only applies to those and the name should probably reflect that

you *can* override the behavior of a boosted form w/ something like this:

```html
<body hx-boost="true">
  ...
  <form hx-swap="innerHTML scroll:none" ... />
    ...
  </form>
  ...
</body>
```

(not tested)

sjc5:
> ok, so https://htmx.org/attributes/hx-swap/ is where you can specify scrolling behavior
> 
> by default it does nothing, unless we are taking about a boosted link or form, so this global really only applies to those and the name should probably reflect that
> 
> you _can_ override the behavior of a boosted form w/ something like this:
> 
> ```
> <body hx-boost="true">
>   ...
>   <form hx-swap="innerHTML scroll:none" ... />
>     ...
>   </form>
>   ...
> </body>
> ```
> 
> (not tested)

Unless I'm missing something, I think it's really important to inherit the global `hx-swap` setting. Sometimes that's going to be more complex, refer to extensions, etc.

If you've set hx-boost at a parent level, you can just use traditional form attributes.

`<form method="POST" action="/whatever" scrollTopOnResponse="false">`

Having to repeat something like `"morph:innerHTML"` everywhere isn't really necessary

1cg:
You can just modify the scrolling behavior:

```html
<form hx-swap="scroll:none" ... />
```

I'm not saying we don't need this global config, I'm just thinking out loud about how widely it matters with respect to the name.  

Maybe `htmx.config.scrollTopIntoViewOnBoost`?

alexpetros:
> by default it does nothing, unless we are taking about a boosted link or form, so this global really only applies to those and the name should probably reflect that

I understand now, and agree. I think `htmx.config.scrollTopIntoViewOnBoost` is also the right name.

sjc5:
> You can just modify the scrolling behavior:
> 
> ```
> <form hx-swap="scroll:none" ... />
> ```
> 
> I'm not saying we don't need this global config, I'm just thinking out loud about how widely it matters with respect to the name.
> 
> Maybe `htmx.config.scrollTopIntoViewOnBoost`?

There may be a bug because I've never been able to get that syntax to work. It still scrolls to top for me on a boosted form, with hx-boost set on the body.

1cg:
OK, that's a bug for sure. Do you mind debugging it?

The "scroll" should be determined by the swap spec.

Cheers,
Carson

/// big sky software ///

sjc5:
> OK, that's a bug for sure. Do you mind debugging it?

Yep, happy to. Thanks for hanging in there with me :)

alexpetros:
@sjc5 would you mind moving that investigation to a new issue?

@xhaggi are you available to change this to `scrollIntoViewOnBoost`?

xhaggi:
First of all, `scroll` and `show` do different things when used as modifiers in `hx-swap`.  `show` calls `element.scrollIntoView` and `scroll` calls `element.scrollTop`. For a boosted link or form, the swap modifier `show` is set to `top` by default.

>You can just modify the scrolling behavior:
>`<form hx-swap="scroll:none" ... />`

@1cg to change the default behavior, you have to use `hx-swap="show:none"` on boosted links and forms.

@alexpetros I don't really like `scrollTopIntoViewOnBoost` or `scrollIntoViewOnBoost` because neither explain what is really happening. What is being scrolled into the view? I like `scrollTopOnBoost` more.

xhaggi:
I have now decided to use `scrollTopOnBoost` and also added a note in the `hx-swap` documentation on how to disable the default behavior on a per-element basis

alexpetros:
> @alexpetros I don't really like scrollTopIntoViewOnBoost or scrollIntoViewOnBoost because neither explain what is really happening. What is being scrolled into the view? I like scrollTopOnBoost more.

The reason to use `scrollIntoViewOnBoost` is that it mirrors the DOM API, which makes it easier to search the documentation for. If I'm like "can htmx do scrollIntoView that I read about on MDN," this makes it easier to find. I don't feel especially strongly about it vs `scrollTopOnBoost`, which is fine, but I do prefer `scrollIntoViewOnBoost` for this reason.

@1cg we just need a final decision on `scrollTopOnBoost` or `scrollIntoViewOnBoost`.

xhaggi:
@alexpetros after a second thought the name `scrollIntoViewOnBoost` makes more sense, especially if you use `hx-target` on a boosted element. This will cause the target to scroll into the viewport. Without `hx-target` the body is used and it will scroll into the viewport, causing the page to scroll to the top. However, I think that probably very few users are aware of this. We should definitely make this clear in the documentation.

xhaggi:
Force pushed. @alexpetros let me know if the documentation is clear enough.

alexpetros:
@1cg beat me to punch and merged it, but the documentation is great :)

Issue #1189, opened 2023-01-03
hx-vals does not work together with delay


sigsegvat:
It seems that the event object is not passed to the jx-vals expression when using a delay modifier in hx-trigger (see code below).

```
<div contenteditable="true" hx-post="/test" contenteditable="true" hx-trigger="input delay:1s" 
     hx-vals="js:{text: convertEvent(event)}"
     >doesn't work</div>

<div contenteditable="true" hx-post="/test" contenteditable="true" hx-trigger="input" 
     hx-vals="js:{text: convertEvent(event)}"
     >works</div>
     
 window.convertEvent= function(event) {
  console.log(event)
  return event.target.innerHTML
}    
```

GlennWSo:
I have also noticed this problem

ahmed-com:
any update on this issue ? any work around ?

1cg:
fixed in 2.0.2

scrhartley:
I think it's a mistake saying that this was fixed in version 2.0.2.
PR #3196 has been merged into the dev branch (Apr 24, 2025), but hasn't made its way to master yet for a release.

PR #1460, opened 2023-05-27
Pass event data through to getValuesForElement

dararish:
Resolves issue #1189

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2649,7 +2649,7 @@ return (function () {
                 }
                 var varsValues;
                 if (evaluateValue) {
-                    varsValues = maybeEval(elt,function () {return Function("return (" + str + ")")();}, {});
+                    varsValues = maybeEval(elt,function () {return Function("event", "return (" + str + ")")(values.event);}, {});
                 } else {
                     varsValues = parseJSON(str);
                 }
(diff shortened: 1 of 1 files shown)


dararish:
I don't see how the evaluation in the `Function` call can access the `values` parameter at all. Does that need to be fixed too or am I missing something?

gulaschsippl:
If this change would be ready to merge, i.e. if it doesn't break any other behavior, I think it would be a good one, as relying on the global `window.event` is relying on a [deprecated property](https://developer.mozilla.org/en-US/docs/Web/API/Window/event).

janisint:
I would also appreciate if this got merged since I am affected by this issue

Telroshan:
Hey, following up very late on this sry, are you still interested in this change @dararish ?
If so, could you please
- Rebase to `dev` and resolve conflicts as the lib has been through quite some changes since then
- Add a test case for this issue (make sure it fails with the current lib, then succeeds with your fix)

y2kbugger:
@dararish I think this is still relevent

lamontfr:
This PR will be very useful to me also.

scrhartley:
I think this could be closed now that PR #3196 has been merged.

PR #1468, opened 2023-06-01
Fix WebSocket extension trigger handlers initialization

### LINKED ISSUE #1396: ws-send does not respect hx-trigger
Hi!

I am unable to send data over a websocket triggered by anything other than the "natural" event of an element. I am using the new websocket extension.

It seems that elements with `ws-send` do not support using `hx-trigger` to specify the trigger for sending a websocket message. The documentation at https://htmx.org/extensions/web-sockets/ indicates that using `hx-trigger` should be possible:
> ws-send - Sends a message to the nearest websocket based on the trigger value for the element (either the natural event of the event specified by [hx-trigger])

Am i missing something, or is the extension not working correctly?

**Example repro (fiddle at https://jsfiddle.net/q0z87uko/):**

```html

<script src="https://unpkg.com/htmx.org@1.9.0"></script>
<script src="https://unpkg.com/htmx.org/dist/ext/ws.js"></script>

<div hx-ext="ws" ws-connect="wss://ws.postman-echo.com/raw">
  <div ws-send>
    Click me to send WS (this works)
  </div>
  <div ws-send hx-trigger="click">
    Click me to send WS (this does not work)
  </div>
  <input ws-send hx-trigger="keyup changed" value="Edit me to send WS (this does not work)" />
</div>
```

Renerick:
Fix #1396 

"Naked hx-trigger" feature causes event handlers added after https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1938 to be ignored by https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1412. Not sure what to do with this right now, but moving extension initialization sooner seems to fix the problem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
-## [1.9.3] - 2023-04-28
+## [1.9.3] - 2023-06-??
+
+* Fixed bug w/ WebSocket extension initilization caused by "naked" `hx-trigger` feature
 
 ## [1.9.2] - 2023-04-28
 
--- a/src/ext/ws.js
+++ b/src/ext/ws.js
@@ -52,7 +52,7 @@ This extension adds support for WebSockets to htmx.  See /www/extensions/ws.md f
 					return;
 
 				// Try to create websockets when elements are processed
-				case "htmx:afterProcessNode":
+				case "htmx:beforeProcessNode":
 					var parent = evt.target;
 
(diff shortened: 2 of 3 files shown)


aral:
@1cg Is there any chance we could have a new release with this fix. I’m running into this currently (and need to use a release build as I’m working with [Kitten](https://codeberg.org/kitten/app) that includes first-class support for htmx via the latest release builds.)

PR #1478, opened 2023-06-05
[New feature] selector-less next and previous targets

Telroshan:
I'm very often using the following pattern :
```html
<div class="relative">
    <button hx-get="/some_popup" hx-target="nextElementSibling" hx-swap="outerHTML"></button>
    <div class="popup"></div>
</div>
```
clicking on the button brings in a popup, in place of a initially empty element. The relative parent allows positioning the popup next to the button with a position set to absolute. Ending up with something like:
```html
<div class="relative">
    <button hx-get="/some_popup" hx-target="nextElementSibling" hx-swap="outerHTML"></button>
    <div class="popup visible">
        <button>Action 1</button>
        <button>Action 2</button>
        Or whatever else...
    </div>
</div>
```

Since the JS API exposes [element.nextElementSibling](https://developer.mozilla.org/docs/Web/API/Element/nextElementSibling) and [element.previousElementSibling](https://developer.mozilla.org/docs/Web/API/Element/previousElementSibling), I think eponym values for `hx-target` would make sense and be handy.

The existing way of doing such a thing would be, in this example, to use a `hx-target="next div"` or `hx-target="next .popup"` for example. The current implementation under those next & previous elements, [are calling document.querySelectorAll](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L583) with the passed selector, then comparing positions.

Not that it's especially slow: I tried it out just to see, and for a `next div` selector on a page with ~30 000 divs, it takes from `1.5 ms` if the element is near the start of the results set, up to `75 ms` if it's rather to the end of the results set. Don't get me wrong there ; it's clearly negligible given such a huge page. Still though, in this situation, calling nextElementSibling / previousElementSibling is, in comparison, instant

I thought we might as well simply use the fastest path here, given that this change only adds 482 bytes to the code _(that is, before minification and gzip)_, and, in my opinion, makes for a clearer structure _(readability)_ when you know you're targeting the direct adjacent sibling

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -600,6 +600,10 @@ return (function () {
                 return [scanForwardQuery(elt, normalizeSelector(selector.substr(5)))];
             } else if (selector.indexOf("previous ") === 0) {
                 return [scanBackwardsQuery(elt, normalizeSelector(selector.substr(9)))];
+            } else if (selector === "nextElementSibling") {
+                return [elt.nextElementSibling]
+            } else if (selector === "previousElementSibling") {
+                return [elt.previousElementSibling]
             } else if (selector === 'document') {
                 return [document];
             } else if (selector === 'window') {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -596,14 +596,14 @@ return (function () {
                 return [closest(elt, normalizeSelector(selector.substr(8)))];
             } else if (selector.indexOf("find ") === 0) {
                 return [find(elt, normalizeSelector(selector.substr(5)))];
+            } else if (selector === "next") {
+                return [elt.nextElementSibling]
             } else if (selector.indexOf("next ") === 0) {
                 return [scanForwardQuery(elt, normalizeSelector(selector.substr(5)))];
+            } else if (selector === "previous") {
+                return [elt.previousElementSibling]
             } else if (selector.indexOf("previous ") === 0) {
                 return [scanBackwardsQuery(elt, normalizeSelector(selector.substr(9)))];
-            } else if (selector === "nextElementSibling") {
-                return [elt.nextElementSibling]
-            } else if (selector === "previousElementSibling") {
-                return [elt.previousElementSibling]
             } else if (selector === 'document') {
                 return [document];
             } else if (selector === 'window') {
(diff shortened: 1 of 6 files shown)


alexpetros:
Re: our previous discussion about the duplication of this syntax, would this make it available to `hx-trigger` too?

Telroshan:
Sure! They would be available everywhere
If you want, I can add the mentions to those new selectors in the other attributes' documentations right away as well

Though I was more thinking about working on that shared/common extended selectors documentation after this PR, and refactor all docs by the way

Let me know what would be the preferred way, I hope we can include all those changes before the next release

alexpetros:
> If you want, I can add the mentions to those new selectors in the other attributes' documentations right away as well
> 
> Though I was more thinking about working on that shared/common extended selectors documentation after this PR, and refactor all docs by the way

I'd like to suggest both—add it in both places for this PR, which will make the process of refactoring it easier later.

Telroshan:
Should be fine @alexpetros !

Since it is a 1-word selector, the code is already working as is for hx-trigger's `from`, I added test cases for that one too

I had to update to update the tests' `make` util function as it wouldn't behave as a standard scenario ; it was appending & processing HTML elements one at a time (making the `nextElementSibling` selector parsing fail as the next sibling wouldn't be in the DOM yet), whereas in a real scenario all works fine (as the DOM is ready when htmx processes nodes)

Please note that I'm not adding mentions in the following documentations:
- `hx-include`, as #1509 hasn't been merged thus its documentation page doesn't mention any extended selector
- `hx-indicator` as its doc doesn't details the extended selectors syntax (except `closest`)

alexpetros:
Yup, very reasonable, I know there's a lot of moving parts here. I'll tag this as "ready for review" once 1.9.4 goes out.

Would you like to create the issue for consolidating the expanded selector syntax (that references the other places we've had this discussion)? I'm happy to do it as well.

xhaggi:
Sorry for joining this discussion so late. What if we just use `next` and `previous` with no CSS selector following, instead of `nextElementSibling` and `priviousElementSibling`?

Telroshan:
Oh I didn't think about that, you're completely right @xhaggi, it's much more concise + would reuse an existing keyword, and makes perfect sense, thanks!

xhaggi:
Please rebase against dev instead of merge.

Telroshan:
You're right, forgot how merge pollutes the history, thanks @xhaggi

PR #1479, opened 2023-06-05
[New feature] Comma separated list syntax for HX-Trigger response header

Telroshan:
Adds support for comma separated list of event names to the `HX-Trigger` response header.
Currently, to trigger multiple events with that header, you have to use the JSON syntax, even if you don't have any additional data to send.

This change lets you specify a list of event names, separated by commas, without additional details, like so:
`HX-Trigger: event1, event2`

The syntax is similar to the one of the [`hx-target`](https://htmx.org/attributes/hx-trigger/#multiple-triggers) attribute, that also supports a comma separated list syntax

### DIFF
--- a/www/content/headers/hx-trigger.md
+++ b/www/content/headers/hx-trigger.md
@@ -60,9 +60,15 @@ document.body.addEventListener("showMessage", function(evt){
 
 Each property of the JSON object on the right hand side will be copied onto the details object for the event.
 
-Finally, if you wish to invoke multiple events, you can simply add additional properties to the top level JSON
+### Multiple Triggers
+
+If you wish to invoke multiple events, you can simply add additional properties to the top level JSON
 object:
 
 `HX-Trigger: {"event1":"A message", "event2":"Another message"}`
+
+You may as well trigger multiple events with no additional details by sending event names separated by commas, like so:
+
+`HX-Trigger: event1, event2`
 
 Using events gives you a lot of flexibility to add functionality to normal htmx responses.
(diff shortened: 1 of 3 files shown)


www/content/headers/hx-trigger.md
```diff
@@ -60,9 +60,15 @@ document.body.addEventListener("showMessage", function(evt){
...
 
-Finally, if you wish to invoke multiple events, you can simply add additional properties to the top level JSON
+### Multiple Triggers
+
+If you wish to invoke multiple events, you can simply add additional properties to the top level JSON
 object:
 
 `HX-Trigger: {"event1":"A message", "event2":"Another message"}`
 
+You may as well trigger multiple events with no additional details by sending event names separated by commas, like so:
```
1cg, line 70: Can you change to "You may also"?

Change looks great!

1cg (changes requested):

www/content/headers/hx-trigger.md
Telroshan: Thanks for the correction! I learned [something new](https://dictionary.cambridge.org/grammar/british-grammar/also-as-well-or-too)

Telroshan: Just in case you had missed it, allow me to ping you @1cg as I know you're overwhelmed with notifications/email

alexpetros:
@Telroshan is this one ready to go?

Telroshan:
@alexpetros yep, the requested wording change was pushed !

1cg (approved):

Issue #1469, opened 2023-06-01
Using extensions with Vite


matteocontrini:
Hello, I'm using Vite as a bundler and I'm unable to make htmx extensions work.

If I write:

```
import 'htmx.org';
import 'htmx.org/dist/ext/preload.js';
```

it compiles correctly but I then get this error in the console:

```
Uncaught ReferenceError: htmx is not defined
```

I've also tried making sure that htmx is available globally:

```
import * as htmx from 'htmx.org';
window.htmx = htmx;
```

But it doesn't seem to make a difference.

So is there a way to use extensions with a bundler like Vite?

Thanks.

xhaggi:
At the moment all extensions are missing a UMD wrapper. Therefore, you need to use the [@rollup/plugin-inject
](https://www.npmjs.com/package/@rollup/plugin-inject/v/5.0.3) to inject htmx into the extensions.

```js
// vite.config.js
import inject from '@rollup/plugin-inject'
import { defineConfig } from 'vite'

export default defineConfig({
  plugins: [
    inject({
       htmx: 'htmx.org'
    }),
  ],
})
```

After configuring the plugin in your `vite.config.js`, you should be able to use the extensions.

```js
import 'htmx.org/dist/ext/preload.js';
```

matteocontrini:
@xhaggi that's helpful, thank you!

However I think there's still something else missing. For example, the `preload` extension is loaded correctly but doesn't seem to work in practice.

It appears that the `htmx:afterProcessNode` event (on which the extension relies) is never called. If I enable debug mode I only see an `htmx:load` event being called and nothing else.

Any ideas? I think I can provide a sample project if it's helpful.

xhaggi:
Thank you. I looked deeper into this issue and the reason why it does not work, is the fact that htmx initialize itself on DOM ready. A `<script type="module">` is async by default and the DOM is ready at the time of evaluating the `import from 'htmx.org'`, which then initializes htmx without the extension. You can achieve the same behavior by adding the `async` attribute to the `<script>` tags in the `htmx.html` file. At the moment I think there is a no workaround for this problem, but I am working on a fix and will create a pull request in the next few days.

matteocontrini:
I imagined it was something like that but I'm not very familiar with htmx internals. Thanks a lot for investigating and working on this!

Sciumo:
Is this issue still open and is there an example of using Vite with htmx?

davidjr82:
[...]

Issue #151, opened 2020-07-26
Built Toolchain?


benpate:
So I submitted my first pull request (yay!) and am working on another.  I've noticed myself missing a number of pieces of infrastructure that I'm used to having around on other projects.  I'm **usually not in favor** of big, complicated toolchains for building Javascript projects.  Less is more.  But here is a short list of some tools that have really helped me in the past.  

Are these being considered for the future 1.0 release?

**Minification** 
I like [rollup](https://rollupjs.org/) because it's a little more modern than [WebPack](https://webpack.js.org/).

**Linting Tool**
I like [tslint](https://github.com/palantir/tslint) because of Typescript (see below) but any solid linter would do.

**Typescript**
I know this is a big one, and controversial with many Javascript developers.  Typescript has helped me catch so many errors before they even get saved.  It's been great for self-documenting code, and automatically identifying edge cases (like unhandled null values) that are easy for humans to miss.  I could go on, but I'd strongly encourage making a branch to explore building/transpiling the core HTMX files with Typescript.

--
Rolling out things like this has consequences for every person who touches the code, so it should be much more of a discussion than a pull request from a newbie.  But, I'll throw out a few suggestions to see what makes sense to the powers that be :) And, if there's a decision/consensus from you all who've actually been doing this for a while, I'd be happy to volunteer to do a bunch of the work.

1cg:
Hi @benpate,

I'm happy to have someone improve the build chain.  Broadly speaking I would prefer to have as simple a build chain as possible for htmx.  I don't like the javascript UMD stuff and I would prefer to write javascript in the idiosyncratic manner that I have been, as a global object.  I don't mind small transformations for generating a UMD or alternative final product, as long as it is small.  The closer the /src version is to the /dist version, the better IMO.

Typescript is an interesting idea, but with the test suite I don't think matters as much.  I like type systems mainly for auto-complete and WebStorm does a pretty good job in javascript.  In general I would prefer a typed language to work in, but since we end up producing javascript anyway, I lean towards just dealing with the many, many warts of that language.

All that being said, the hyperscript project seems like it would be much more amenable to a more elaborate build system and potentially using TypeScript:

https://github.com/bigskysoftware/_hyperscript

Any interest in working on that?

benpate:
Hey, thanks for responding (esp when you're swamped).

[...]

PR #1482, opened 2023-06-06
Add gulp as build system to generate UMD wrapper also for extensions

xhaggi:
This adds a minimal build system based on [gulp.js](https://gulpjs.com/) to add a UMD wrapper, minify and gzip the source files, including all extensions.

Related issue #1469


matsp:
I think the project would really benefit from this PR - I was already on the way to do a PR myself after discovering that extensions are not minified. When gulp is not minimal enough, we could run uglify over all distributed files?

alexpetros:
Incorporating a build system into htmx has been discussed extensively, and then ultimately rejected. See here for some of that discussion:
* https://github.com/bigskysoftware/htmx/issues/151
* https://github.com/bigskysoftware/htmx/issues/606
* https://github.com/bigskysoftware/htmx/pull/1250

It looks like there's a reasonable workaround too? https://github.com/bigskysoftware/htmx/issues/1469#issuecomment-1577414184

The question of minifying/UMDing the extensions is a fair one. I would suggest opening an issue dedicated to it and then we can discuss the best way to handle that. I know there's some possible willingness to at least automating the UMD situation, but it's a slightly touchy subject, so a PR can't really happen until we have some consensus on that.

gnat:
image

lol

matsp:
@xhaggi maybe we should provide a fork which code is identical but with your PR content?

xhaggi:
@matsp That would certainly not be the right way to go, nor is it my intention.

matsp:
Sure it's not the right way. This library is getting more and more attraction and thus people will ask about this PR. Providing just a pull solution to import in their projects will lead to the right amount of pressure in the maintainer team to think about it.
I don't really want to fork away, it's just about a alternative solution till this team is ready to change their minds.

andryyy:
Imo it's still the correct decision.

PR #1489, opened 2023-06-12
Add support for hx-on-* attribute

### LINKED ISSUE #1480: Proposal: `on:` listener attributes
## Problem
HTML does not support attribute listeners for custom DOM events. For example, HTML supports `onclick` for the [click](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) event, but nothing for HTMX custom events like [htmx:beforeSwap](https://htmx.org/events/#htmx:beforeSwap):

```html
<!-- Legal -->
<button onclick="doSomething()">

<!-- Not legal -->
<button onhtmx:beforeSwap="doSomething()">
```

The `hx-on` [attribute](https://htmx.org/attributes/hx-on/) was added to rectify this problem, but the introduction of an additional mini-language is somewhat out of step how the rest of HTML works:

```html
<!-- DOM built-in -->
<button onclick="alert('Clicking!')" onmouseover="alert('Hovering!')">

<!-- HTMX -->
<button hx-get="/info" hx-on="htmx:beforeRequest: alert('Making a request!')
                              htmx:afterRequest: alert('Done making a request!')">
```

## Solution
I think the ideal interface is a more HTML-style interface that looks something like this:
```html
<!-- DOM built-in -->
<button onclick="alert('Clicking!')" onmouseover="alert('Hovering!')">

<!-- HTMX -->
<button hx-get="/info" 
        on:htmx:beforeRequest="alert('Making a request!')" 
        on:htmx:afterRequest="alert('Done making a request!')"
>
```
Because you cannot `querySelector` for attribute wildcards (i.e. `querySelectorAll('button[on*]')`), HTMX needs another way to identify elements to which it should attach custom events. For elements that have an `hx-METHOD` attribute already, that isn't a problem, but you might like to add listeners to elements that do not make requests. I propose re-using the `hx-on` attribute, empty, for this purpose:

```html
<button hx-on on:htmx:afterSwap="alert('I was swapped in!')" >
```

When composed with triggers, this completes the idea I was getting at in https://github.com/bigskysoftware/htmx/pull/1327, allowing you to trigger actions based on `hx-trigger` syntax without having to use any javascript (outside the element):

```html
<button hx-on 
        hx-trigger="load delay:1s"
        on:htmx:trigger="this.innerText='Back to normal!'"
>Just Swapped!</button>
```

In addition to being already part of the spec, re-using `hx-on` allows you to write its documentation in such a way that makes it clear this new way is the preferred way of using it, while preserving backwards compatibility with all existing HTMX code that made use of `hx-on`. The new way of using `hx-on` (with `on:` attributes) would _only_ work if its value were empty; otherwise it would default to the deprecated functionality.

I also believe that even though it's not strictly necessary, `hx-on` should be required even if the element has an `hx-METHOD` method already, because it prevents some subtle bugs where a seemingly unrelated event listener stops working because you removed the `hx-get` method.

You can also include the existing DOM events in the new style, under a `dom` namespace:

```html
<!-- DOM events HTMX-style -->
<button hx-on on:dom:click="alert('Clicking!')" on:dom:mouseover="alert('Hovering!')">
```
which is obviously strictly worse than just using the actual `onclick` attribute, but you get it for "free" as part of the interface and demonstrates the naming convention nicely.

_Thanks to @dz4k for bringing this back up in the Discord; I incorporated a lot of his suggestions._

alexpetros:
## Description
Resolves #1480, after some feedback. Adds an additional `hx-on:*` attribute so you can specify the event to listen to directly in the attribute name:

```html
<button hx-get="/info" hx-on:htmx:before-request="alert('Making a request!')">
    Get Info!
</button>
```

And multiple attributes:

```html
<button hx-get="/info"
        hx-on:htmx:beforeRequest="alert('Making a request!'")
        hx-on:htmx:afterRequest="alert('Done making a request!')">
    Get Info!
</button>
```

## Testing
Copied the test suite from `hx-on` and modified it for the new attribute to ensure feature parity. Also added a new test for `data-hx-on`, and updated the `hx-on` documentation.

Thanks in particular to @dz4k for the XPath snippet, which I modified to include `data-hx-on` as well.


dz4k (commented):
Sorry to keep proposing changes to this feature but: should we add a shorthand for htmx events, maybe `hx-on::before-request`?

src/htmx.js
```diff
@@ -1823,6 +1829,16 @@ return (function () {
             return document.querySelector("[hx-boost], [data-hx-boost]");
         }
 
+        function findHxOnWildcardElements(elt) {
+            if (!document.evaluate) return []
+
+            let node = null
+            const elements = []
+            const iter = document.evaluate('//*[@*[starts-with(name(), "hx-on:")]] | //*[@*[starts-with(name(), "data-hx-on:")]]', elt)
```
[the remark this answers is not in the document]
1cg: 🤣 🤣 🤣

test/attributes/hx-on-wildcard.js
```diff
@@ -0,0 +1,122 @@
...
+        this.server.respond();
+
+        // click button again
+        byId('foo').click();
+        window.tempCount.should.equal(2);
+
+        delete window.tempCount;
+    });
+
+});
```
1cg, line 122: 😍 thank you for these tests!

alexpetros: Thank you for writing 80% of them!

alexpetros:
> Sorry to keep proposing changes to this feature but: should we add a shorthand for htmx events, maybe `hx-on::before-request`?

Good suggestion, happy to do it

src/htmx.js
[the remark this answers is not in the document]
alexpetros: I was trying to figure out how to `OR` the attribute condition but I couldn't figure it out and was too embarrassed to ask

dz4k (commented):
LGTM!

dz4k (approved):

www/content/attributes/hx-on.md
```diff
@@ -6,40 +6,81 @@ The `hx-on` attribute allows you to embed scripts inline to respond to events di
...
+    Get Info!
+</button>
+
 ```
 
-### Symbols
+Adding multiple handlers is easy, you just specify additional attributes:
+```html
+<button hx-get="/info"
+        hx-on::beforeRequest="alert('Making a request!'")
```
andryyy, line 42: This should be `"alert('Making a request!')"`

It was exactly that line I copied for a test. 😄

PR #1493, opened 2023-06-13
Preload extension: Include preloaded node as 'source' for preload so hx-sync/queue works properly

checketts:
I found that when using the `preload="preload:init"` attribute some requests were being dropped. Without the source element, the `body` is used for queuing and `hx-sync` can't be used to override the element.

FYI @benpate


1cg:
@benpate can you approve this?  lgtm

checketts:
Sorry for my mistake I had the PR going against master. I rebased and pointed it to dev now.

benpate:
Hey.. I'm looking at this now.  In principle, this is an awesome addition.  Thank you @checketts !  I'll just try to see if it breaks anything else unintentionally.

PR #1511, opened 2023-06-22
Disable hx-indicator while performing preload ajax requests

checketts:
Currently preload ajax requests trigger the configured 'hx-indicator', however when preloading that doesn't necessarily make sense. This change ensures that no hx-indicator is triggered during preload, while preserving any configured hx-indicator for regular htmx activities.

### DIFF
--- a/src/ext/preload.js
+++ b/src/ext/preload.js
@@ -48,12 +48,20 @@ htmx.defineExtension("preload", {
 				// in the future
 				var hxGet = node.getAttribute("hx-get") || node.getAttribute("data-hx-get")
 				if (hxGet) {
+					var previousIndicator = node.getAttribute("hx-indicator");
+					node.setAttribute("hx-indicator", "")
+
 					htmx.ajax("GET", hxGet, {
 						source: node,
 						handler:function(elt, info) {
 							done(info.xhr.responseText);
 						}
 					});
+					if(previousIndicator) {
+						node.setAttribute("hx-indicator", previousIndicator)
+					} else {
+						node.removeAttribute("hx-indicator")
+					}
 					return;
 				}
 


alexpetros:
@checketts can you please update the description describing what this PR solves? I read the issue that it references but a clear description would make this easier to review. I don't think tests or docs are necessary—what you're fixing is basically a bug IMO.

checketts:
@alexpetros Good idea. I've updated the description now.

1cg:
Good idea, but this probably needs to be implemented internally in htmx, since the indicator can be found elsewhere (e.g. a parent).  Also dislike mutating the DOM, just gives me the heebs.  (Not that I don't do it elsewhere... :/)

checketts:
@1cg So would you like a proposal that enhances the htmx.ajax call the adds an indicator as a parameter and even adjusts the typical calls to leverage it?

src/ext/preload.js
```diff
@@ -48,12 +48,20 @@ htmx.defineExtension("preload", {
 				// in the future
 				var hxGet = node.getAttribute("hx-get") || node.getAttribute("data-hx-get")
 				if (hxGet) {
+					var previousIndicator = node.getAttribute("hx-indicator");
+					node.setAttribute("hx-indicator", "")
```
1cg, line 52: OK, have thought about this more and I think you are doing the right thing here.

The only thing I would change is that here, rather than setting the value to "", set it to "unset" so that if there is a parent hx-indicator, that is also ignored.

Thinking about it, the ajax request will fire off and be handled asyn, but the code will keep on executing so the resetting of the attribute in the next set of if statements will be run before any other htmx requests get fired off.

1cg (changes requested):

alexpetros:
@checketts are you available to rebase and make the changes described above?

alexpetros:
Feel free to re-open if you want to make the requested changes.

PR #1513, opened 2023-06-23
[Bugfix] Encode DELETE payload as URL params like GET, instead of using body

Telroshan:
Currently, htmx sends values in the request body for DELETE requests.
While this seems to be _allowed_, the [RFC](https://www.rfc-editor.org/rfc/rfc7231#section-4.3.5) says:
>  A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.

And this is exactly what happens with a Golang backend. As [its doc](https://pkg.go.dev/net/http#Request.ParseForm) says: 
> For POST, PUT, and PATCH requests, it also reads the request body, parses it as a form [...]
> For other HTTP methods, or when the Content-Type is not application/x-www-form-urlencoded, the request Body is not read

A simple fix here is to handle DELETE requests like GET requests in the core, so the parameters get URL encoded instead, then any backend would handle those DELETE requests as expected

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3001,7 +3001,7 @@ return (function () {
             var pathNoAnchor = splitPath[0];
             var anchor = splitPath[1];
             var finalPathForGet = null;
-            if (verb === 'get') {
+            if (verb === 'get' || verb === 'delete') {
                 finalPathForGet = pathNoAnchor;
                 var values = Object.keys(filteredParameters).length !== 0;
                 if (values) {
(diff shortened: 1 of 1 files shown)


1cg:
excellent, ty!

alexpetros:
This PR breaks this test: https://github.com/bigskysoftware/htmx/blob/master/test/ext/json-enc.js#L112

I'm not exactly sure what the best way to handle this is—I think probably the best is to have the `json-enc` extension override this PR's change and send its payload in the body (since that's almost certainly what the user expects).

PR #1518, opened 2023-06-26
Add error messages

checketts:
These error messages will be included in the logError call now, so target selector details and similar will be exposed to the user.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -790,7 +790,8 @@ return (function () {
                 oobElement.parentNode.removeChild(oobElement);
             } else {
                 oobElement.parentNode.removeChild(oobElement);
-                triggerErrorEvent(getDocument().body, "htmx:oobErrorNoTarget", {content: oobElement});
+                triggerErrorEvent(getDocument().body, "htmx:oobErrorNoTarget",
+                    {message:"Unable locate target for out-of-band swap. selector:"+selector, content: oobElement, selector: selector});
             }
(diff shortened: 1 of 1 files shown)


alexpetros:
I don't know... obviously more logging is good but on some level I think these are things you should only log if the developer explicitly asks for them. It makes sense to log the `historyCache` if you have a `historyCache` error, but not necessarily as standard behavior? That's sort of what the debugger is for.

@1cg thoughts?

checketts:
This PR doesn't add any logging that wasn't already being logged. It merely ensures the details are present so the log message is useful.

alexpetros:
Yeah I get that, but I think the expectation from users is that, by default, exceptions print their stack trace and that's it. I don't think it's really the library's place to decide to log a bunch of other runtime variables.

1cg:
We need to think hard about this: string constants blow out the size of the library.  Maybe we can make this an extension, e.g. `dev-messages` or something like that?

checketts:
I agree with the concern around string constants. I'll remove the 'message' attributes and move them into an extension. 

The extension will merely be an event listener on the body that gives the detail log message. So it won't even need to leverage the `hx-ext` mechanism.

alexpetros:
@checketts that sounds good to me. I'm going to close this PR, and feel free to open a PR on the core that hangs relevant data on the error object (but doesn't log it by default) as needed.

Issue #1509, opened 2023-06-22
hx-include including all matching elements in the page, not only from children


atelierJVA:
Not sure if this is a bug or if I am reading the doc wrong.

Inside a div, `.filter-container` in the example below, I have multiple inputs that I want to include in a request, so I added `hx-include="[input]"` to it, but for some reason the input in the outside div is also being included.

```
<div>
  <label for="outside_input">This input is not supposed to be included</label>
  <input id="outside_input" name="not_this_one" type="text">
</div>
<div class="filter-container" hx-include="input">
  <div class="filter-name">
    Include these:
  </div>
  <div class="filter-range">
    <label for="from">From</label>
    <input name="from" type="text">
    <label for="to">To</label>
    <input name="to" type="text">
  </div>
  <div class="buttons-group">
    <button hx-get="/path">Apply</button>
  </div>
</div>
```
When I fill the inputs and click in the `Apply` button I get `path/?not_this_one=99&from=1&to=12`, the query parameter `not_this_one=99` is the one that should not be there.

From on the doc for hx-include  :
"Notes: hx-include is inherited and can be placed on a parent element" and the one for inheritance:
"Most attributes in htmx are inherited: they apply to the element they are on as well as any children elements. " 

So my assumption is that anything outside the element with hx-include should not be selected, is this wrong or is it a bug?

Telroshan:
Hey, the query selector passed to [hx-include](https://htmx.org/attributes/hx-include/) simply resolves to [document.querySelectorAll](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll).
A standard CSS selector will not take the element defining the hx-include's position in the hierarchy

Take a look at [hx-include documentation](https://htmx.org/attributes/hx-include/), where the example code is
```html
<div>
    <button hx-post="/register" hx-include="[name='email']">
        Register!
    </button>
    Enter email: <input name="email" type="email"/>
</div>
```
You can see here, how the included input is not a child of the button, yet is included thanks to the `[name='email']` selector.

So what can you do here?
You may use the [htmx extended selectors](https://htmx.org/docs/#extended-css-selectors), so with your example code, you could do:
1. ```<button hx-get="/path" hx-include="closest .filter-container">Apply</button>``` which will include the closest parent element that has the `filter-container` class, and thus, will include any input that is a child of that container
This will **not** include the outside input, as it's not a child of that container
Remember that, as [the doc](https://htmx.org/attributes/hx-include/) says:
[...]

PR #1522, opened 2023-06-27
List extended selectors + caveats in hx-include doc

Telroshan:
This addresses issue #1509
- `hx-include` supports the extended selectors, which were not mentioned in its documentation
- The behaviour of `find` for example, can be confusing when using an inherited hx-include attribute: added a note to explain that
- Extended selectors only return a single element, which could be confusing compared to a standard CSS selector that returns all elements matching that selector: added a note to explain that

### DIFF
--- a/www/content/attributes/hx-include.md
+++ b/www/content/attributes/hx-include.md
@@ -2,8 +2,19 @@
 title = "hx-include"
 +++
 
-The `hx-include` attribute allows you to include additional element values in an AJAX request.  The value of
- this attribute is a CSS query selector of the element or elements to include in the query.
+The `hx-include` attribute allows you to include additional element values in an AJAX request. The value of this
+attribute can be:
+
+* A CSS query selector of the elements to include.
+* `this` which will include the descendants of the element.
+* `closest <CSS selector>` which will find the [closest](https://developer.mozilla.org/docs/Web/API/Element/closest)
+  ancestor element or itself, that matches the given CSS selector
+  (e.g. `closest tr` will target the closest table row to the element).
+* `find <CSS selector>` which will find the first child descendant element that matches the given CSS selector.
+* `next <CSS selector>` which will scan the DOM forward for the first element that matches the given CSS selector.
+  (e.g. `next .error` will target the closest following sibling element with `error` class)
+* `previous <CSS selector>` which will scan the DOM backwards for the first element that matches the given CSS selector.
+  (e.g `previous .error` will target the closest previous sibling with `error` class)
 
 Here is an example that includes a separate input value:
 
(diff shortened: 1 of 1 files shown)


checketts (approved):
Good clarifications!

Telroshan:
That's the next thing I was going to suggest! 😄
Since they are used for hx-include, hx-indicator, hx-target, hx-trigger's from modifier, and probably others that I'm forgetting, I think we should give the htmx's extended selectors syntax its own page and link to that page in every attribute doc supporting the syntax.
I'm afraid we would inevitably have differences and outdated descriptions in each attribute's doc otherwise

alexpetros:
My only concern is that I think the `trigger` syntax and the other ones are parsed separately, which doesn't necessarily mean they should be separate in the documentation, but it's possible they could diverge

Telroshan:
I hadn't noticed this, indeed others use [`querySelectorAllExt`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L583) while [`getTriggerSpecs`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1217) does its own parsing.
Though, looking at the code, `getTriggerSpecs`'s [logic](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1260) is just "is the keyword one of the following predefined extended selectors ? If so, parse one more token", but then this parsed selector is also [resolved through `querySelectorAllExt`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1399)

- I can't think of any reason hx-trigger's `from` would need to use a different selectors set than, say, hx-include
- if you're concerned about the duplication here @alexpetros  (technically, the extended selectors are hardcoded at 2 places in the code), maybe we could do an array, a map or something else there, to factor the logic and define the selectors only once?
Probably a dumb idea but what comes to the top of my head is an `{ "selectorName": function }`, such as
```js
var extendedSelectors = {
    "closest": (elt, selector) => [closest(elt, normalizeSelector(selector.substr(8)))],
    "find": (elt, selector) => [find(elt, normalizeSelector(selector.substr(5)))],
    ...
}
```
_(Using the short syntax on purpose here but ofc this isn't IE11 compatible)_
Basically [`querySelectorAllExt`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L583)'s code, but as a map instead of a if-else series, thus something that can be iterated from another place in the code, to loop through all defined extended selectors without hardcoding them once again

alexpetros:
> I wonder if copy-pasting that bit about the `find`, `closest`, etc. mini-language everywhere is the way to go, or if we shouldn't standardize it somehow.

After letting this sit for too long, I've decided that just because we *should* do this eventually isn't a reason to hold up this PR.

alexpetros (approved):

Issue #1514, opened 2023-06-25
Fix json-enc DELETE implementation


alexpetros:
> This PR breaks this test: https://github.com/bigskysoftware/htmx/blob/master/test/ext/json-enc.js#L112
>
> I'm not exactly sure what the best way to handle this is—I think it's probably to have the `json-enc` extension override this PR's change and send its payload in the body (since that's almost certainly what the user expects). That probably requires an additional core change though.

_Originally posted by @alexpetros in https://github.com/bigskysoftware/htmx/issues/1513#issuecomment-1605837195_

Telroshan:
Probably a naive question, but should've the test been there in the first place?
I see there is no "GET" version of those tests for the json-enc extension, adding one quickly also raises the same error. 
Which isn't surprising, because the [RFC for the GET method](https://www.rfc-editor.org/rfc/rfc7231#section-4.3.1) says the same thing as for the DELETE method ( see #1513 )
> A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request.

So if you're going to add support for JSON body anyway for DELETE, you might as well want to add it for GET (which also doesn't seem to work atm)

Or am I missing something there?

alexpetros:
No it's a fair question! That's why I made the issue for it. 

My take on this is that while it's *generally* considered terrible practice to add payload bodies to GET requests, I don't know if the same cultural norm applies to DELETE and I expect that people making use of the `json-enc` extension will expect that to work.

As you correctly point out, the semantics of both are undefined, so there's no reason to allow DELETE bodies but not GET bodies. Since the `json-enc` extension is purely additive (and applied selectively), I would cautiously be in favor of also adding it to GET, since, if you want to define that way for your application, why not?

Telroshan:
Yeah I would be in favor of adding to both of them in that case, for the sake of consistency and homogeneity

andryyy:
I never really thought about that. But I seem to misuse DELETE with data bodies. I did it recently when I wanted to drop a bunch of ids from a db while not looping every record.

1cg:
yeah, seems like we should make it possible to override including things in the body via the `htmx:configRequest` event to me

Telroshan:
While I understand the retro-compatibility issue, I would still argue that not following the RFC looks more like a bug than a feature to me, and a "bugged state" shouldn't be the default behaviour but an optional one, letting the user move away from the RFC recommandations if they want to

[...]

PR #1525, opened 2023-06-29
Add configs to choose between ecoding parameters in the URL or request body

alexpetros:
## Summary
Even though the semantics of request bodies for GET are undefined, it is legal to set request bodies for them. By default, GET form parameters should be encoded as URLs. If, however, an extension defines `encodeParameters`, that should override the default and set the request bodies for GET methods, just like it does for all the other HTTP methods.

Resolves: #1514, #1539

## Testing
Wrote some new unit tests for `GET` requests that have `json-enc` enabled, and this PR fixes the broken `DELETE` one.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2973,6 +2973,7 @@ return (function () {
             var eltIsBoosted = getInternalData(elt).boosted;
             var requestConfig = {
                 boosted: eltIsBoosted,
+                useUrlParams: verb === 'get', // For GET requests, default to params in the URL
                 parameters: filteredParameters,
                 unfilteredParameters: allParameters,
                 headers:headers,
@@ -3009,18 +3010,16 @@ return (function () {
             var pathNoAnchor = splitPath[0];
             var anchor = splitPath[1];
 
-            // Encode body within the URL if:
-            // 1. The HTTP method for the request is GET
-            // 2. an extension on the element defines the `encodeParameters` function
-            var shouldEncodeBodyInUrl = verb === 'get'
+            // Override the useUrlParams config if an extension defines encodeParameters
+            var useUrlParams = requestConfig.useUrlParams
             withExtensions(elt, function(extension) {
                 if (typeof extension.encodeParameters === 'function') {
-                    shouldEncodeBodyInUrl = false
+                    useUrlParams = false
                 }
             })
 
             var finalPath = path
-            if (shouldEncodeBodyInUrl) {
+            if (useUrlParams) {
                 finalPath = pathNoAnchor;
                 var values = Object.keys(filteredParameters).length !== 0;
                 if (values) {
@@ -3132,7 +3131,8 @@ return (function () {
                 });
             });
             triggerEvent(elt, 'htmx:beforeSend', responseInfo);
-            xhr.send(shouldEncodeBodyInUrl ? null : encodeParamsForBody(xhr, elt, filteredParameters));
+            var params = useUrlParams ? null : encodeParamsForBody(xhr, elt, filteredParameters)
+            xhr.send(params);
             return promise;
         }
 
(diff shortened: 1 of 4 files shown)


src/htmx.js
```diff
@@ -3008,26 +3008,35 @@ return (function () {
...
+            var shouldEncodeBodyInUrl = verb === 'get'
+            withExtensions(elt, function(extension) {
+                if (typeof extension.encodeParameters === 'function') {
+                    shouldEncodeBodyInUrl = false
+                }
+            })
+
+            var finalPath = path
+            if (shouldEncodeBodyInUrl) {
+                finalPath = pathNoAnchor;
```
1cg, line 3024: what do you think about handling this via a new field on `requestConfig` above?  Something like `requestConfig.useBody`?  We could default that value to `verb == 'get'` and then let folks override it when necessary?

for the default we could also introduce a config variable, something like `htmx.config.paramRequestTypes` that defaults to `['get']` and test membership in that, which would allow folks who want correct delete behavior to write this:

```js
htmx.config.paramRequestTypes.push('delete');
```

1cg (changes requested):

src/htmx.js
alexpetros: Agreed, except I did `requestConfig.useUrlParams` instead since I think it's a little more descriptive.

alexpetros: Likewise I think `htmx.config.httpMethodsThatDefaultToUrlParams` instead of `htmx.config.paramRequestTypes`. I know it's a mouthful but it's not going to be a very common config (soon) and I'd rather the name be too long than too short.

alexpetros: If you disagree on either of those names let me know and I'll change them.

alexpetros: I have settled on `htmx.config.methodsThatUseUrlParams`

alexpetros:
@1cg updated and ready for review

src/htmx.js
```diff
@@ -3008,26 +3014,32 @@ return (function () {
...
-            if (verb === 'get' || verb === 'delete') {
-                finalPathForGet = pathNoAnchor;
+
+            // Override the useUrlParams config if an extension defines encodeParameters
+            withExtensions(elt, function(extension) {
+                if (typeof extension.encodeParameters === 'function') {
+                    useUrlParams = false
+                }
+            })
+
```
1cg, line 3024: think we can drop this now?  I don't like signaling this via the presence of a function, would rather the extension do so via the `htmx:configRequest` event if necessary...

1cg (changes requested):

src/htmx.js
alexpetros: Yup that's smart, done.

1cg:
thx @alexpetros!

jantari:
So I found this PR (and related issues) after searching for why my Go backend didn't accept the form data that HTMX sends in a DELETE request.

According to this stackoverflow post: https://stackoverflow.com/a/59011736

Sending a body in a DELETE request is non-standard and cannot be expected to work. Indeed the body is ignored by Golangs `request.ParseForm()` method.

It looks like this PR didn't just make the behavior of DELETE requests configurable, which is fine, but also changed their default behavior from the normal and preferred way of encoding parameters in the URL to sending them in the body. This is now an issue for Golang users such as myself, and surely for other languages or libraries as well.

Am i misinterpreting something or should the default behavior for DELETEs be reverted back to encoding in URL query params?

alexpetros:
> It looks like this PR didn't just make the behavior of DELETE requests configurable, which is fine, but also changed their default behavior from the normal and preferred way of encoding parameters in the URL to sending them in the body. This is now an issue for Golang users such as myself, and surely for other languages or libraries as well.
>
> Am i misinterpreting something or should the default behavior for DELETEs be reverted back to encoding in URL query params?

Yes, you're wrong about the second part—htmx has always sent DELETE bodies by default, that's what inspired the PR in the first place. This default is being changed in 2.0 to the behavior that you're requesting, but in the meantime, you can configure it using the config that was implemented here.

PR #1528, opened 2023-06-30
Add support for from: modifier on intersect events

JanRomero:
This would allow you to do infinite scroll with better locality of behaviour, by putting the attribs on the list itself:

```HTML
<ul id="tweets"
    hx-get="/api/moretweetspls/"
    hx-swap="beforeend"
    hx-include="#tweets li:last-child"
    hx-trigger="intersect from:footer"
>

    <div class="tweet">
        <input hidden name="timestamp" value="123456789"/>
    </div>

    <div class="tweet">
        <input hidden name="timestamp" value="123456788"/>
    </div>

</ul>

<footer>mAdE wItH &#10084; In dUdErStaDt</footer>
```

thx

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1766,6 +1766,7 @@ return (function () {
                 maybeReveal(elt);
             } else if (triggerSpec.trigger === "intersect") {
                 var observerOptions = {};
+                const from = triggerSpec.from ? querySelectorExt(elt, triggerSpec.from) : elt;
                 if (triggerSpec.root) {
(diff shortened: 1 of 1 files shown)


alexpetros:
I like it, can you update the docs and add tests?

alexpetros:
Feel free to re-open if you make the above changes.

JanRomero:
@alexpetros Hey thanks for having a look! I've added a test and docs to the branch, could you reopen the PR or should I make a new one?

Telroshan:
@JanRomero seems you'll have to open a new PR unfortunately
image

PR #1530, opened 2023-06-30
Update podcasts.csv

srcerer:
grug make rick roll, grug want share


alexpetros:
This seems great to me, but I do wonder if someone needs to actually listen to what's in these podcasts before they go on the website

srcerer:
> 

Yes, I've listened to all of these. Only including links to podcasts Carson was guest of.

alexpetros:
Okay I'm taking your word for it 👍

alexpetros (approved):

PR #1542, opened 2023-07-05
Update alpine template example in docs.md

jdiaz5513:
Using `$watch` with AlpineJS in this manner ends up creating a race condition (which I've observed in live code) where sometimes the templated elements are not yet rendered by the time the `$watch` callback is invoked.

Using `x-init` like this instead has proven to be 100% reliable and less verbose in my use case - this technique should be recommended to newcomers to avoid the race condition I personally lost time to.


alexpetros:
I honestly don't know anything about alpine, so I need to find someone who does to evaluate this.

alexpetros:
I guess if you also have some evidence that this your approach is preferred in the Alpine community, that would help too.

alexpetros:
Closing this for the above reasons. We're not really here to provide up-to-date Alpine examples, more to demonstrate a trivial example of how the two might be integrated. If you someone feels strongly that our example on the page isn't good enough for that purpose, feel free to comment it and we can revisit this.

Issue #1541, opened 2023-07-05
Using the same `name` attribute on multiple HTML form elements results in only first value being sent


vrzh:
Greatly enjoying using htmx!

I have an issue with HTML forms. I have multiple form elements that share the same value for `name`. If an ordinary HTML `POST` is done, I receive all values. But, if htmx is used, then I receive only the first value.

I provide a contrived minimal example below, where I want to send a default "action" value that can be "overridden" by different submit buttons. A similar trick can be used to send a "default" value when a checkbox is not checked. Or, if you just want to send a list of values to the server.

If you click on button "C" of the "HTML" form, the server receives something like `[('action', 'A'), ('action', 'C')]`, as expected.

However, if you click on button "C" of the "htmx" form, the server receives only `[('action', 'A')]`.

Printing out `evt.detail.parameters` when the `htmx:configRequest` event occurs also shows a single value.

Is this a bug? Or is there some way to hook into htmx to get the same behaviour?

Example:

```html
<!doctype html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>

    <h3>HTML</h3>
    <form action="/submit-html" method="POST">
        <input type="hidden" name="action" value="A">
        <button type="submit">A</button>
        <button type="submit">A</button>
        <button type="submit" name="action" value="B">B</button>
        <button type="submit" name="action" value="C">C</button>
    </form>

    <h3>htmx</h3>
    <form hx-post="/submit-htmx" method="POST">
        <input type="hidden" name="action" value="A">
        <button type="submit">A</button>
        <button type="submit">A</button>
        <button type="submit" name="action" value="B">B</button>
        <button type="submit" name="action" value="C">C</button>
    </form>

    <script src="https://unpkg.com/htmx.org@1.9.2"></script>

    <script>
        document.body.addEventListener('htmx:configRequest', function(evt) {
            console.log(evt.detail.parameters);
        });
    </script>

  </body>
</html>
```

Telroshan:
I would say that not behaving as a form element does by default, is a bug "by definition", but I may be wrong.
I think it's due to [this line](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L2260) that explicitly ignores any value attached to a button
The comment here says
> `// ignore "submitter" types (see jQuery src/serialize.js)`

So it seems to be voluntary?
May I invoke you @1cg there? As it could be related to the "philosophy" of the lib maybe?

vrzh:
[...]

Issue #1506, opened 2023-06-21
Submit button values are not sent


azmeuk:
On a simple form, the name and value of a `<button type="submit">` button are not sent.

```html
<form>
    <input type="text" name="text-field" value="input OK">
    <button type="submit" name="submit-button" value="foobar" hx-target="#results"
  hx-post="/echo/json">button KO</button>
</form>

<code id="results"></code>
```

You can check this fiddle, and inspect the requests to observe this behavior: https://jsfiddle.net/azmeuk/7ztxd820/54/

Using `hx-include` on the button does not send the data either so it cannot be used as a workaround, but I could manage to achieve what I want with `hx-vals`.

Might be related to #970 #971 #1120 #1337

Telroshan:
Submitted a PR that addresses this issue.

In the meantime, for current & older versions, even though it may not be what you want, it turns out that putting the `hx-post` (and `hx-trigger`) on the form itself, correctly includes the clicked button's value, if of course such workaround is possible in your situation

azmeuk:
Fixed since 1.9.4
Thank you!

Issue #1337, opened 2023-03-23
ws-send does not send submit button name and value


medihack:
This issue is very similar to #1132, but when using the [HTMX WebSockets extension](https://htmx.org/extensions/web-sockets/). When a form is sent to the server using WebSocket extension by pressing a submit button (regardless of inside or outside the form), its name and value are not transmitted.

medihack:
I investigated a bit further and think the problem is that `initButtonTracking` is only called when an [htmx attribute](https://github.com/bigskysoftware/htmx/blob/09565cde1bad12b28a552a998391b63fd7194a6e/src/htmx.js#L1771) is on the [form itself](https://github.com/bigskysoftware/htmx/blob/09565cde1bad12b28a552a998391b63fd7194a6e/src/htmx.js#LL1829C21-L1829C39). But according to the [documentation](https://htmx.org/extensions/web-sockets/) `hx-ext="ws"` should be placed outside the form.

Renerick:
WS extensions uses htmx's method for gathering input data. Buttons with type submit are not included. I don't think that changing this on the extension level is a good idea.

As for initButtonTracking and ws-ext workaround, I'll investigate further, what's going on there

medihack:
It should somehow (on the server) be distinguished which button was pressed. The workaround works because [`initButtonTracking`](https://github.com/bigskysoftware/htmx/blob/09565cde1bad12b28a552a998391b63fd7194a6e/src/htmx.js#L1782) is called in [`initNode`](https://github.com/bigskysoftware/htmx/blob/09565cde1bad12b28a552a998391b63fd7194a6e/src/htmx.js#L1803) which is [only called on nodes](https://github.com/bigskysoftware/htmx/blob/09565cde1bad12b28a552a998391b63fd7194a6e/src/htmx.js#L1848) that have an attribute of kind [[hx-sse], [data-hx-sse], [hx-ws], [data-hx-ws], [hx-ext], [data-hx-ext]](https://github.com/bigskysoftware/htmx/blob/09565cde1bad12b28a552a998391b63fd7194a6e/src/htmx.js#LL1774C17-L1775C63). I hope this helps with the investigation.

1cg:
fixed

Issue #1120, opened 2022-11-04
hx-post with external submit button excludes the button value


wooooodward:
Submitting a form with the submit button outside of the `<form>` tag does not include the button's `name` attribute as a parameter or its value (`'lol': 'cats'`) in the request which is inconsistent with the default expected behaviour.

```
<form id="my-form" hx-post="/example">
  <input type="text" name="name" />
</form>

<input type="submit" form="my-form" name="lol" value="cats"/>
```

mihalikv:
Same issue here with the newest version 1.8.5.

drowe-official:
Also ran into this issue. It popped up when I was trying to use form buttons inside a Bootstrap modal footer. For anyone in the same boat- I ended up having to wrap the entire `modal-dialog` div in the `form` tag, which isn't ideal.

alexpetros:
Resolved by #1559

PR #1559, opened 2023-07-10
[Bugfix] Fix submit buttons/inputs handling

Telroshan:
The suggested changes should fix the following issues:
- #1506
=> values of clicked submit buttons defining a name/value pair, are now correctly included in the request payload
- #1120 
=> Submit buttons/inputs using the [`form attribute`](https://developer.mozilla.org/docs/Web/HTML/Element/button) and placed outside a form, are now correctly included in the request payload when clicked
- #971 / #970 / #1337 
=> `form`, `a` and `input type="submit"` are now processed by htmx not only when the element has `chanceOfBeingBoosted`. This fixes the issue where you had to put `hx-ext` on the form itself, otherwise the button's own value wouldn't be in the payload.
Also, this doesn't seem to do any harm since it doesn't impact the default behaviour when no htmx logic should happen.
    - `htmx:beforeProcessNode` and `htmx:afterProcessNode` are going to be dispatched
    - `processVerbs` won't do anything if the element doesn't declare hx-verb attribute
    - `hasExplicitHttpAction` being false, nothing happens in the following if either
    - `initButtonTracking` will be called for any form OR submit button/input that declares the "form" attribute (i.e. outside a form). `initButtonTracking` is what fixes the issue in the end, as it's going to properly set the form's `internalData.lastButtonClicked` when clicking a button that it outside that form
- #1541
=> values of clicked submit button/input now correctly add to regular input values with the same name, [as it does in a standars form](https://jsfiddle.net/o1qxec4g/)

**Notes**
- Added test cases inspired from all the mentioned issues above
- All tests pass, though let me know if I missed any usecase

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2316,6 +2316,29 @@ return (function () {
             return true;
         }
 
+        function addValueToValues(name, value, values) {
+            // This is a little ugly because both the current value of the named value in the form
+            // and the new value could be arrays, so we have to handle all four cases :/
+            if (name != null && value != null) {
+                var current = values[name];
+                if (current === undefined) {
+                    values[name] = value;
+                } else if (Array.isArray(current)) {
+                    if (Array.isArray(value)) {
+                        values[name] = current.concat(value);
+                    } else {
+                        current.push(value);
+                    }
+                } else {
+                    if (Array.isArray(value)) {
+                        values[name] = [current].concat(value);
+                    } else {
+                        values[name] = [current, value];
+                    }
+                }
+            }
+        }
+
         function processInputValue(processed, values, errors, elt, validate) {
             if (elt == null || haveSeenNode(processed, elt)) {
                 return;
@@ -2332,28 +2355,7 @@ return (function () {
                 if (elt.files) {
                     value = toArray(elt.files);
                 }
-                // This is a little ugly because both the current value of the named value in the form
-                // and the new value could be arrays, so we have to handle all four cases :/
-                if (name != null && value != null) {
-                    var current = values[name];
-                    if (current !== undefined) {
-                        if (Array.isArray(current)) {
-                            if (Array.isArray(value)) {
-                                values[name] = current.concat(value);
-                            } else {
-                                current.push(value);
-                            }
-                        } else {
-                            if (Array.isArray(value)) {
-                                values[name] = [current].concat(value);
-                            } else {
-                                values[name] = [current, value];
-                            }
-                        }
-                    } else {
-                        values[name] = value;
-                    }
-                }
+                addValueToValues(name, value, values);
                 if (validate) {
                     validateElement(elt, errors);
                 }
@@ -2422,17 +2424,8 @@ return (function () {
             if (internalData.lastButtonClicked || elt.tagName === "BUTTON" ||
                 (elt.tagName === "INPUT" && getRawAttribute(elt, "type") === "submit")) {
                 var button = internalData.lastButtonClicked || elt
-                var name = getRawAttribute(button, "name");
-                if (name) {
-                    var currentValue = values[name]
-                    if (typeof currentValue === "undefined") {
-                        values[name] = button.value
-                    } else if (Array.isArray(currentValue)) {
-                        values[name].push(button.value)
-                    } else {
-                        values[name] = [currentValue, button.value];
-                    }
-                }
+                var name = getRawAttribute(button, "name")
+                addValueToValues(name, button.value, values)
             }
 
             return {errors:errors, values:values};
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1293,7 +1293,7 @@ return (function () {
                 return triggerSpecs;
             } else if (matches(elt, 'form')) {
                 return [{trigger: 'submit'}];
-            } else if (matches(elt, 'input[type="button"]')){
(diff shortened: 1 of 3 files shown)


src/htmx.js
```diff
@@ -1849,17 +1845,23 @@ return (function () {
 
         function findElementsToProcess(elt) {
             if (elt.querySelectorAll) {
-                var boostedElts = hasChanceOfBeingBoosted() ? ", a, form" : "";
-                var results = elt.querySelectorAll(VERB_SELECTOR + boostedElts + ", [hx-sse], [data-hx-sse], [hx-ws]," +
+                var results = elt.querySelectorAll(VERB_SELECTOR + ", a, form, [type='submit']" + ", [hx-sse], [data-hx-sse], [hx-ws]," +
```
xhaggi, line 1848: To avoid processing each `a` and `form` element without a `hx-boost` attribute, I would suggest changing the selector from `a, form` to
```js 
a[hx-boost], a[data-hx-boost], [hx-boost] a, [data-hx-boost] a,
form[hx-boost], form[data-hx-boost], [hx-boost] form, [data-hx-boost] form
```

The old implementation already did this, but not really strictly.
And there is no need to concat the strings `..[type='submit']" + ", [hx-sse]...`.

Telroshan: We actually want htmx to process non-boosted `a` and `form` elements here since some extensions can have behaviors depending on them, that are outside htmx's initial scope.
For example, one of the linked issues was related to `ws-send` that can be placed on a child form, which shouldn't require to be boosted, see [the websocket extensions tests](https://github.com/bigskysoftware/htmx/pull/1559/files#diff-5d44e06c69b1e00b7645e380407a00efeb408254485e448e6262aa4237b69445R437)

Htmx processing non-htmx elements will, in the worst case, simply result in those elements having an empty internal data for links, and the `lastButtonClicked` property in internal data for forms (since that element won't declare hx-on, hx-trigger & so on, htmx won't process anything on it)

Telroshan: You're right about the useless concatenation though, let me fix that

xhaggi: >We actually want htmx to process non-boosted a and form elements here since some extensions can have...

Then we could also add `[hx-ext] a, [hx-ext] form` to solve these problems, couldn`t we?

>Htmx processing non-htmx elements will, in the worst case, simply result in those elements having an empty

Apart from what you said, htmx will also trigger events, register event listeners (initButtonTracking), but most of the time you do not use `hx-boost` or an extension.

Telroshan: I'm not familiar with the extensions system, will an element always be a descendant of a `hx-ext` element ?

xhaggi: I would say yes, as described in the documentation:
>You will need to include the javascript file for the extension and then install it using the `hx-ext` attributes.

>For example, one of the linked issues was related to ws-send that can be placed on a child form, which shouldn't require to be boosted, see [the websocket extensions tests](https://github.com/bigskysoftware/htmx/pull/1559/files#diff-5d44e06c69b1e00b7645e380407a00efeb408254485e448e6262aa4237b69445R437)

I had a look at the websocket extension code and IMO an extension needs to take care of handling special attributes like `ws-send` itself. So I am not sure why we should initialize links or forms without `hx-boost` at all.

BTW: I adjust my code to check if all my tests are green, even without the additional selector `[hx-ext] a, [hx-ext] form`. And yep, all tests are fine.

xhaggi: After reading the issue https://github.com/bigskysoftware/htmx/issues/1337 I'm sure we need the additional selector `[hx-ext] form` because of the button tracking stuff for form elements.

xhaggi: Ah and don't forget `[data-hx-ext] form`.

Telroshan: Now that I think about it, I think there's another issue with this selectors set.
Htmx runs them relatively to the root element that's being initialized.
Which means, if htmx were to process a `form` that's dynamically appended to a `hx-ext` ascendant (whether it be through standard htmx attributes, or manually + a `htmx.process` call), that element would be ignored, since it wouldn't match the selector

Something like
```html
<div hx-ext="whatever">
    <div id="target"></div>
</div>

<button hx-post="/url" hx-target="#target">Click me</button>
```
If `/url` responds with a form, that gets swapped inside target, with your suggested selectors, that form won't be processed since it wouldn't match any of those selectors (as it'd be run on the `target` element, not the document root)

I'll probably add some tests to ensure htmx processes correctly dynamically added elements as well

Telroshan: I was skeptical at first about the real gain of this, given that it adds "a lot" (subjective ofc) of code just to avoid processing elements and firing 2 events, that would ultimately cause no issue anyway (since htmx doesn't get in the way of standard / expected behaviours for those `a` and `form` elements),

Decided to benchmark it with [this JSFiddle](https://jsfiddle.net/26kshucz/). I was honestly even expecting your code to be slightly slower
```
Finding out if form should be processed : 30 ms
Firing events : 140 ms
```
_(10 000 iterations, average results)_

Turns out firing events is actually freaking slow in comparison.

So now I'm fully convinced 😄 thanks for your help! I'll add those changes

Telroshan: Pushed, all tests pass, great! Thanks again, closing this conversation

1cg (approved):

PR #1562, opened 2023-07-11
[Documentation] Fix beforeProcessNode link in event reference doc

Telroshan:


alexpetros (approved):
good catch, thank you

PR #1567, opened 2023-07-13
[Bugfix] Handle dynamically added hx-disable

### LINKED ISSUE #1176: Dynamically adding hx-disable to element doesn't disable htmx actions on element

Telroshan:
This fixes #1176 
When dynamically adding hx-disable to an element from your custom code:
- If any trigger handler is triggered on an element, it first checks if the element is currently disabled or not. If it is, cleans up the element _(which removes all htmx listeners from it)_ & aborts the handler
- You can also call `htmx.process` right after adding the attribute, to force the clean up immediately

When **removing** hx-disable from an element from your custom code:
- You will have to explicitly call `htmx.process` to have htmx initialize that element and bind handlers again

Added test cases for `hx-disable` dynamic addition / removal

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -937,6 +937,9 @@ return (function () {
                     }
                 });
             }
+            if (internalData.initHash) {
+                internalData.initHash = null
+            }
             deInitOnHandlers(element);
         }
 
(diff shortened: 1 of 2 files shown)


alexpetros:
These tests are 👍👍

Issue #721, opened 2021-12-16
HTMX breaks command + click


thekashifmalik:
I've noticed that HTMX-fied links no lo longer let you command click to open in a new tab. This seems surprising.

Does it makes sense to manually handle that for links by default?

waldyrious:
I just came here to report that the logo in the htmx website doesn't allow Ctrl+click to open in a new tab.

However, I'm not sure if that one counts as an "HTMX-fied link":

https://github.com/bigskysoftware/htmx/blob/e3c0dadbe7de09fde4df32fb7d730ab8385379b4/www/_includes/layout.njk#L30

Happy to open a separate issue if it is indeed unrelated.

waldyrious:
For the record, I am able to Ctrl+click [hx-boosted](https://htmx.org/attributes/hx-boost/) links, such as the ones in the website's navbar: 

https://github.com/bigskysoftware/htmx/blob/e3c0dadbe7de09fde4df32fb7d730ab8385379b4/www/_includes/layout.njk#L40-L44

...and have them open in a new tab. If it helps, I'm using Firefox 95.0 on Ubuntu 20.04.

gmso:
Hi,

I believe the problem is that DOM elements with an `hx-get` attribute are not considered links by the browser. Therefore, the _Ctrl+Click_ command does not open the new route in a new tab, as it usually does with regular links using `<a>` tags.

Related to that, what @waldyrious mentions seems to show that even boosted `<a>` tags have this behavior (like in the website's navbar) but not other HTML tags with an `hx-get` attribute (like in the website's logo). Interestingly enough, if you change the website's logo element from a `<span>` to a normal `<a>` tag pointing to the root, the _Ctrl+Click_ behaviour works as expected. So from:
```html
<span hx-get="/" hx-target="body" hx-push-url="true" class="logo light"> ... </span>
```
To:
```html
<a class="logo light" href="/"> .... </a>
```

With this change I can _Ctrl+Click_ and get a new tab with the root page. This seems to work with Firefox 95.0.2 and Chrome  96.0.4664.110. I haven't tested any other browsers.

For me, the rule of thumb would be to use boosted `<a>` tags when you want to point to a different page/route within your site, and `hx-get` attributes when you want to dynamically fetch data from the server, but you want to stay inside the same page/route

benpate:
[...]

PR #1572, opened 2023-07-14
Don't preventDefault on events that are filtered

HeyHugo:
Run the `maybeFilterEvent()` check before `evt.preventDefault()` so that events excluded via `hx-trigger` filters are ignored and not prevented.

Prior to this change `<a href="/some/url" hx-get="/some/url" hx-trigger="click[!ctrlKey && !metaKey]">` would not let the browser use default ctrl + click behaviour due to the event being prevented.

This PR makes [the suggestion here](https://github.com/bigskysoftware/htmx/issues/721#issuecomment-1003263995) viable without adding another event listener on your own.

For a bit of context I have a use case where it makes sense to use `hx-get` on a `<a>` tag. It's a modal/overlay thing using `hx-push-url` and `hx-history-elt`. Direct browsing to the url makes sense and I'd really like to keep `<a>` ctrl/meta + click behaviour for the specific link. My current solution is to add another event listener specifically for click + modifier and then do `window.open("/some/url", "_blank")`.

My workaround is to add something like this:
```javascript
  const link = document.querySelector("#some-container a");
  const url = link.getAttribute('href');
  link.addEventListener("click", function (e) {
    if (e.metaKey || e.ctrlKey) {
      e.preventDefault();
      window.open(url, "_blank");
    }
  });
```

It works but requires me to reimplement the default browser behaviour with another event listener. I think it'd make sense in the general case that if you filter an event out explicitly it keeps whatever default there is.

### DIFF
--- a/test/core/regressions.js
+++ b/test/core/regressions.js
@@ -99,7 +99,7 @@ describe("Core htmx Regression Tests", function(){
         byId("d1").innerText.should.equal('Replaced');
     });
 
-    it('does not submit with a false condition on a form', function() {
+    it('submission is aborted but event default is not prevented when event is excluded via filter', function() {
         this.server.respondWith("POST", "/test", "Submitted");
         var defaultPrevented = false;
         htmx.on("click", function(evt) {
@@ -108,7 +108,8 @@ describe("Core htmx Regression Tests", function(){
         var form = make('<form hx-post="/test" hx-trigger="click[false]"></form>');
         form.click()
         this.server.respond();
-        defaultPrevented.should.equal(true);
+        defaultPrevented.should.equal(false);
+        form.innerHTML.should.equal("");
     })
 
     it('two elements can listen for the same event on another element', function() {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1413,11 +1413,11 @@ return (function () {
                     if (ignoreBoostedAnchorCtrlClick(elt, evt)) {
                         return;
                     }
-                    if (explicitCancel || shouldCancel(evt, elt)) {
-                        evt.preventDefault();
-                    }
                     if (maybeFilterEvent(triggerSpec, elt, evt)) {
                         return;
+                    }
+                    if (explicitCancel || shouldCancel(evt, elt)) {
+                        evt.preventDefault();
                     }
                     var eventData = getInternalData(evt);
                     eventData.triggerSpec = triggerSpec;


HeyHugo:
@alexpetros Aha a test had to be updated. Did so now, can you trigger workflow again 🙏

judyburgett3 (approved):

judyburgett3:
if it can be done do it. thank you.

judyburgett3 (commented):
thank you

judyburgett3:
please unblocke it

judyburgett3 (approved):

judyburgett3:
unblock it and go. thank you.

judyburgett3:
please unblock it and and it going

judyburgett3:
thank you

judyburgett3:
please keep going

judyburgett3 (approved):

HeyHugo:
@judyburgett3 Spamming won't help. 

Meanwhile you can use the workaround I posted above.

alexpetros:
Thanks Hugo. This is going to be looked at for 1.9.6 btw

alexpetros:
Thanks Hugo. This is going to be looked at for 1.9.6 btw

judyburgett3:
thank you for fixing it.

judyburgett3 (commented):
make the changes

judyburgett3:
unblock it

judyburgett3:
thank you

test/core/regressions.js
```diff
@@ -99,7 +99,7 @@ describe("Core htmx Regression Tests", function(){
         byId("d1").innerText.should.equal('Replaced');
     });
 
-    it('does not submit with a false condition on a form', function() {
+    it('submission is aborted but event default is not prevented when event is excluded via filter', function() {
```
alexpetros, line 102: Is this a chance in behavior that could cause possible regressions?

alexpetros:
Related: https://github.com/bigskysoftware/htmx/pull/1456

test/core/regressions.js
HeyHugo: I would say no. If by any chance you rely on the event excluded by the filter preventing the default of that event you rely on a "bug" behavior.

There could be a case where someone is using a filter on a <a> element and relying on the behavior that any clicks even those excluded by filter does not navigate the user away. In this case one could say it's a regression, but I'd argue:

Why are you using an <a> element with href attr if you don't intend to navigate the user? Perhaps replace it with a button.

It'd be an extremely rare regression worth the change to allow ctrl/meta + click links by only adding a filter.

1cg: what about someone who has an event filter on a form that has an hx-post on it?

if this change is accepted, the form would submit if the filter was false, no?

i see what you are saying here, but I think that this change might be a little too aggressive.

1cg:
OK, closing this out for now although we should probably make this use case addressable some way, either via additional syntax in the trigger, events, etc.  Thank you for your efforts here!

PR #1580, opened 2023-07-15
Make elementSelector, elementSwapStyle in multi-swap.js cleaner and m…

newturok:
In this updated code, the variable declarations for split, elementSelector, and elementSwapStyle are combined into a single line. Array destructuring is used to assign the values from element.split(/\s*:\s*/). The first element is assigned to elementSelector, and the second element, if present, is assigned to elementSwapStyle. If the second element is null or undefined, the default value of "innerHTML" is used.

This change reduces the code length and improves readability by eliminating the need for the typeof check and using a more concise syntax for variable declaration and assignment.

### DIFF
--- a/src/ext/multi-swap.js
+++ b/src/ext/multi-swap.js
@@ -16,10 +16,8 @@
                 var elements = swapStyle.replace(/^multi\s*:\s*/, '').split(/\s*,\s*/);
 
                 elements.map(function (element) {
-                    var split = element.split(/\s*:\s*/);
-                    var elementSelector = split[0];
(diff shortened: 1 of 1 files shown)


Telroshan:
Unfortunately, [the destructuring syntax isn't supported on IE 11](https://caniuse.com/?search=destructuring). and htmx 1 strives to support IE 11.

IE 11 support is planned to be dropped for htmx 2 though, so I guess (but I can't speak for @1cg there) that your change could be accepted when htmx 2 happens :)

PR #1582, opened 2023-07-15
[Documentation] Make it clear that `from` with a standard selector gets all matching elements

Telroshan:
Just to make it clear that
`hx-trigger="keyup from:input"` will retrieve all inputs on the page and listen to keyup on each of them, while
`hx-trigger="keyup from:find input"` will only retrieve the first input child (as well as all the extended selectors described on the reference page)


alexpetros:
You can merge doc fixes against `master`; if you do that, it will update the website immediately instead of when the release is cut.

alexpetros (approved):

Telroshan:
Oh right, I'll do that for future documentation PRs, thanks !

PR #1586, opened 2023-07-16
Fix: `hx-on` event listener clean-up

JackWithOneEye:
Event listeners added to an element via `hx-on` will not be removed in `deInitOnHandlers` as intended.

The reasons are as follows:

- in `addHxOnEventHandler`, the return value of `addEventListener`, which is `undefined`, was set as `listener` in the object pushed to `onHandlers`
- in `deInitOnHandlers`, the wrong keys were used to access event name and listener from an object contained in `onHandlers`

This PR fixes both these issues.


1cg (approved):
thank you!

PR #1579, opened 2023-07-15
Fix typos in www/

szepeviktor:
Found some misspellings with `typos`.
Excluded paths:
- dist/
- test/lib/
- test/realtime/static/data.json
- www/static/test/

Please regenerate `dist` files after this PR. I am totally lost in this repo!

PR #1590, opened 2023-07-17
Fix typos

szepeviktor:
Continuation of #1579


src/ext/response-targets.js
```diff
@@ -83,7 +83,7 @@
                 evt.detail.xhr.status !== 200) {
                 if (evt.detail.target) {
                     if (htmx.config.responseTargetPrefersExisting) {
-                        evt.detail.shoudSwap = true;
+                        evt.detail.shouldSwap = true;
```
alexpetros, line 86: This is actually a change to the extension, and seems to be a bug, so it should be in its own PR with a test to catch whatever isn't being caught here.

alexpetros (changes requested):

src/ext/response-targets.js
alexpetros: Everything else looks good

szepeviktor: Okay.
Could you open an issue for this bug?

alexpetros: If you don't, I probably will, but since you're the one who seems to want it done, you'll definitely get to it sooner than I will ;)

szepeviktor: I am not a htmx user. This is theoretical PR.

szepeviktor: Reverted.

alexpetros:
🤖

Issue #1537, opened 2023-07-04
OOB does not escape query selector


eduardvercaemer:
I have a very simple HTML with some elements such as
```html
<div id="foo-/bar/">
```

I want to execute an OOB swap, but sending the following HTML
```hyml
<div id="foo-/bar/" hx-swap-oob="true">
```
results in the following error in the console
```
Failed to execute 'querySelectorAll' on 'Document': '#foo-/bar/' is not a valid selector.
```

The query selector can be escaped like '#foo-\\/bar\\/' and should be done automatically by htmx.

alexpetros:
Is that a valid `id`?

matiboy:
@alexpetros [According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id), it is:

> Technically, the value for an id attribute may contain any character, except [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) characters.

The [documentation for querySelector](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#escaping_special_characters) actually mentions escaping special characters though it surprisingly does not mention using [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static) which would be the easiest way to handle the above.

I'll try and create a PR for this.

alexpetros:
I still think this is a reasonable request if you want to write your own escape function for just wait for htmx 2.

danielo515:
what about using getElementByID when it is detected to be an ID? that method is able to work with any id string

Encephala:
This problem doesn't just apply to `hx-swap-oob`, it applies to any selector HTMX uses. The following snippet yields an error similar to the one in this issue:

```html
<button hx-get="/clicked" hx-swap="beforeend" hx-target="#target!">Click me</button>
<div id="target!"></div>
```

That makes the solution more complicated, as we can't simply apply some escape function inside f.i. [`normalizeSelector`](https://github.com/bigskysoftware/htmx/blob/4dc97a400f81f3d7b8e2b4d6cb18c3c4cb7cec5e/src/htmx.js#L587-L594), as characters like `,` have meaning in attributes outside of CSS selectors.

[...]

PR #1602, opened 2023-07-18
OOB Swap to accept IDs with special characters

matiboy:
Refer to #1537 

Includes test with some special characters for a single id and multiple.

An alternative to escaping would be to use `getElementById` in the relevant case.

This PR does not cover the case where the selector is in the form `beforeend:#some/bad/id` though we may want to add that too?

Note that this will need to be merged with #1563 as it affects the same code

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -760,7 +760,7 @@ return (function () {
          * @returns
          */
         function oobSwap(oobValue, oobElement, settleInfo) {
-            var selector = "#" + oobElement.id;
+            var selector = "#" + CSS.escape(oobElement.id);
             var swapStyle = "outerHTML";
             if (oobValue === "true") {
                 // do nothing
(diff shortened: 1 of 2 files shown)


src/htmx.js
```diff
@@ -760,7 +760,7 @@ return (function () {
          * @returns
          */
         function oobSwap(oobValue, oobElement, settleInfo) {
-            var selector = "#" + oobElement.id;
+            var selector = "#" + CSS.escape(oobElement.id);
```
xhaggi, line 763: `escape()` is not supported by IE11.

https://caniuse.com/mdn-api_css_escape_static

matiboy: Well that settles that then. I'll close the PR. Though could you please point to the docs regarding browser support which HTMX guarantees so I won't make the same mistake next time, thanks!

Telroshan: Couldn't it be an option to write our own escaping function here? To avoid relying on a IE11 incompatible API while still supporting special ids like those?

alexpetros:
> Well that settles that then. I'll close the PR. Though could you please point to the docs regarding browser support which HTMX guarantees so I won't make the same mistake next time, thanks!

htmx supports IE11, which is the limiting factor. htmx 2.0 will not though, at which point we'll have to figure out exactly what guarantees that means.

PR #1603, opened 2023-07-18
Polishing findElementsToProcess function

xhaggi:
Some code cleanups and optimizations.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -119,9 +119,6 @@ return (function () {
         }
 
         var VERBS = ['get', 'post', 'put', 'delete', 'patch'];
-        var VERB_SELECTOR = VERBS.map(function(verb){
-            return "[hx-" + verb + "], [data-hx-" + verb + "]"
-        }).join(", ");
 
         //====================================================================
         // Utilities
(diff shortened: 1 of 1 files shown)


alexpetros:
This breaks the feature it refactors and also it's not clear what the value add for this PR is. Again, there's no contribution guidelines so I promise to get some hard rules about this at some point. There's lots of issues that we have open (and you've opened PRs for some of them!).

xhaggi:
>This breaks the feature it refactors.

I have fixed the bug. I thought that's what the code reviews and CI pipeline were for, or isn't it?

>and also it's not clear what the value add for this PR is.

These changes clearly improve the code by moving code only used in that function to the function and also improve the boosted link + form selection. Cleaner code leads to better maintainability and understanding for contributors. And I know that this is a matter of taste.

>Again, there's no contribution guidelines so I promise to get some hard rules about this at some point.
Does this mean that code cleanups are evil and PRs are closed immediately 😄 I hope we'll find a better way to handle PRs with code cleanups.

alexpetros:
> And I know that this is a matter of taste.

> Does this mean that code cleanups are evil and PRs are closed immediately

With the caveat that this isn't official yet, yes, that's what I'm going for. We don't have the ability to handle everyone "cleaning" the code to their taste. If you're making a performance or practical improvement, please explain it, but just refactoring because you think it makes the project more maintainable is not helpful at this time—especially as we're developing htmx 2 in parallel.

PR #1604, opened 2023-07-18
[cleanups] Get rid of the getDocument() function + reduce dupe code

xhaggi:
Since both `document` and `getDocument()` are used wildly, I decided to use `document` because it is not necessary to wrap `document` in a getter. The second commit reduces the code that has to be written to access document.body.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -177,13 +177,6 @@ return (function () {
         }
 
         /**
-         * @returns {Document}
-         */
-        function getDocument() {
-            return document;
-        }
-
-        /**
          * @param {HTMLElement} elt
          * @param {(e:HTMLElement) => boolean} condition
          * @returns {HTMLElement | null}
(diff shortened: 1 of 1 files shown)


Telroshan:
Probably nitpicking, but the argument I'd see in favor of using the `getDocument` wrapper, is that it's minified to a way more concise name when building htmx.
If you look at [htmx.min.js](https://github.com/bigskysoftware/htmx/blob/master/dist/htmx.min.js), you'll see the function becomes `function K(){return document}`, while the `document` instruction itself cannot be minified because it's [part of the standard API](https://developer.mozilla.org/docs/Web/API/Document).

Your `body` wrapper is in the same vein helping minifying the code even more as the `.body` instruction couldn't be minified either, while your wrapper can

I fully agree that we should homogeneize the document calls and either pick one way or the other, though the decision isn't mine anyway!

alexpetros:
Especially with htmx 2 coming up, I'm not sure that this kind of refactor is a good idea, and I also don't know what value it adds. That's not a criticism @xhaggi, you've put up a bunch of good PRs that we're going to get to, but I generally don't think refactor PRs are a good use of our time at this moment.

alexpetros:
Discussed in the discord but we're going to close refactor PRs. I promise this will make getting to your other PRs faster—I'm tagging them for intake/review in 1.9.5

Issue #1569, opened 2023-07-14
Active-search example partially incompatible with Firefox


xSke:
The [active-search](https://htmx.org/examples/active-search/) example includes the following paragraph:

> Since we use a `search` type input we will get an `x` in the input field to clear the input. To make this trigger a new `POST` we have to specify another trigger. We specify another trigger by using a comma to separate them. The `search` trigger will be run when the field is cleared but it also makes it possible to override the 500 ms delay on `keyup` by just pressing enter.

However, both the `x` in the input field, and the ability to override the delay by pressing enter, seem to be Chrome-specific behavior.

While `type="search"` is standardized, it's supposed to behave equivalently to a `type="text"`. And the `search` event is, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/search_event), also non-standard and should not be relied on.

It may be worth either redoing this example in a browser-compatible way, or at least adding a line to the page saying a lot of the functionality only works in Chrome.

alexpetros:
> It may be worth either redoing this example in a browser-compatible way, or at least adding a line to the page saying a lot of the functionality only works in Chrome.

Both of these are reasonable, feel free to submit a PR.

alexpetros:
On second thought, let's just fix the example.

PR #1608, opened 2023-07-18
Extend documentation for active search example

Djursing:
A quick fix for #1569


alexpetros:
Appreciate you jumping on this, but we're going to work with the opener of the original issue to get a better solution before we resort to this one.

brablc:
I would suggest accepting this. It really is confusing to Firefox users.

PR #1610, opened 2023-07-18
Fix `confirmed` being ignored in htmx:confirm event

### LINKED ISSUE #1588: Custom "htmx:confirm" also triggers window.confirm
Using htmx 1.9.3

Reproduction:

```html
<button hx-get="" hx-confirm="Something here">Click me</button>
<script>document.body.addEventListener("htmx:confirm", (event) => {
  event.preventDefault()
  // Some logic to decide whether to trigger request
  if(true) {
    event.detail.issueRequest()
  }
})</script>
```

The above will trigger the `window.confirm` regardless of prevent default etc.
This is caused by [failing to check the value of confirmed](https://github.com/bigskysoftware/htmx/blob/8766ad7db4d7578ac2178daf5e95a2ba90a4b03c/src/htmx.js#L2940C13-L2940C13) , even though `issueRequest` specifically [sets `confirmed` to `true`](https://github.com/bigskysoftware/htmx/blob/8766ad7db4d7578ac2178daf5e95a2ba90a4b03c/src/htmx.js#L2839)

A workaround is to set `hx-confirm=""` (empty string) (then confirmQuestion is falsy), but my code wants to use that value in a Sweetalert title for example.

Additionally, the value of the question is not available on [the event details](https://github.com/bigskysoftware/htmx/blob/8766ad7db4d7578ac2178daf5e95a2ba90a4b03c/src/htmx.js#L2841)

My proposal would be to 

1. add the question to the details:

```js
var confirmDetails = {question: getClosestAttributeValue(elt, "hx-confirm"), target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest};
```

2. modify `issueRequest` to take `skip_confirmation` as argument in case the custom `htmx:confirm` wishes to continue with `window.confirm` (should be backwards compatible)

```js
var issueRequest = function(skip_confirmation) {
  return issueAjaxRequest(verb, path, elt, event, etc, !skip_confirmation)
}
```

3. add the check for confirmed:

```js
if(confirmQuestion && !confirmed) {
  if(!confirm(confirmQuestion)) {
```

I can come up with a PR if this behavior is accepted

matiboy:
Referring to #1588 

- Adds test for existing `hx-confirm` behavior by stubbing window.confirm; tests were added to a new test file `hx-confirm.js`
- Enhance `htmx:confirm` event detail to include the value found in `hx-confirm` (set to `detail.question`)
- Allow passing `skipConfirmation` to `detail.issueRequest`; when true, this skips window.confirm and triggers the request right away
- Fix the missing check for `confirmed` in original code (that was the original issue)
- Fix typo in [original `htmx:config` test ](https://github.com/bigskysoftware/htmx/blob/2534a970b6d606cb590edd8f6097e4a6440d8d06/test/core/events.js#L646) (incorrect handler would be removed in case of test failure)

Backwards compatibility: if `skipConfirmation` is not passed, the behaviour remains the same as it was previously, meaning it still calls window.confirm if `hx-confirm` has a value; I think it's incorrect, but it is the old behavior, so this should be backwards compatible

Closes #1588

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2871,8 +2871,8 @@ return (function () {
             var confirmQuestion = getClosestAttributeValue(elt, "hx-confirm");
             // allow event-based confirmation w/ a callback
             if (!confirmed) {
-                var issueRequest = function() {
-                    return issueAjaxRequest(verb, path, elt, event, etc, true);
+                var issueRequest = function (skipConfirmation) {
+                    return issueAjaxRequest(verb, path, elt, event, etc, skipConfirmation);
                 }
                 var confirmDetails = { target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest, question: confirmQuestion };
                 if (triggerEvent(elt, 'htmx:confirm', confirmDetails) === false) {
@@ -2972,8 +2972,7 @@ return (function () {
                 }
             }
 
-
-            if (confirmQuestion) {
+            if (confirmQuestion && !confirmed) {
                 if(!confirm(confirmQuestion)) {
                     maybeCall(resolve);
                     endRequestLock()
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2870,9 +2870,9 @@ return (function () {
 
             var confirmQuestion = getClosestAttributeValue(elt, "hx-confirm");
             // allow event-based confirmation w/ a callback
-            if (!confirmed) {
+            if (confirmed === undefined) {
                 var issueRequest = function(skipConfirmation) {
-                    return issueAjaxRequest(verb, path, elt, event, etc, skipConfirmation);
+                    return issueAjaxRequest(verb, path, elt, event, etc, !!skipConfirmation);
                 }
                 var confirmDetails = {target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest, question: confirmQuestion};
                 if (triggerEvent(elt, 'htmx:confirm', confirmDetails) === false) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2868,12 +2868,13 @@ return (function () {
                 return;
             }
 
+            var confirmQuestion = getClosestAttributeValue(elt, "hx-confirm");
             // allow event-based confirmation w/ a callback
             if (!confirmed) {
                 var issueRequest = function() {
                     return issueAjaxRequest(verb, path, elt, event, etc, true);
                 }
-                var confirmDetails = {target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest};
+                var confirmDetails = { target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest, question: confirmQuestion };
                 if (triggerEvent(elt, 'htmx:confirm', confirmDetails) === false) {
                     return;
                 }
@@ -2971,7 +2972,7 @@ return (function () {
                 }
             }
 
-            var confirmQuestion = getClosestAttributeValue(elt, "hx-confirm");
+
             if (confirmQuestion) {
                 if(!confirm(confirmQuestion)) {
                     maybeCall(resolve);
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2871,10 +2871,10 @@ return (function () {
             var confirmQuestion = getClosestAttributeValue(elt, "hx-confirm");
             // allow event-based confirmation w/ a callback
             if (!confirmed) {
-                var issueRequest = function (skipConfirmation) {
+                var issueRequest = function(skipConfirmation) {
                     return issueAjaxRequest(verb, path, elt, event, etc, skipConfirmation);
                 }
-                var confirmDetails = { target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest, question: confirmQuestion };
+                var confirmDetails = {target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest, question: confirmQuestion};
                 if (triggerEvent(elt, 'htmx:confirm', confirmDetails) === false) {
                     return;
                 }
--- a/test/attributes/hx-confirm.js
+++ b/test/attributes/hx-confirm.js
@@ -53,5 +53,33 @@ describe("hx-confirm attribute", function () {
         }
     })
 
+    it('should prompt when htmx:confirm handler calls issueRequest', function () {
+        var confirm = sinon.stub(window, "confirm");
+        try {
+            var btn = make('<button hx-get="/test" hx-confirm="Surely?">Click Me!</button>')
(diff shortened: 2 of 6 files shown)


alexpetros (commented):
Cleverly done. I loove the test-to-code ratio and appreciate the attention to backwards compat. Could you update the docs to indicate the presence of `detail.question` and the `skipConfirmation` argument?

I'm actually a little torn on how the `skipConfirmation` should work (I'm not sure we need that be backwards compatible, but it's always better to be on the safe side).

matiboy:
@alexpetros I changed the example slightly because the example given as text didn't match the one that was demo-ed at the bottom. So I kept the bottom version (uses click+custom event) as hyperscript and moved the version using `invokeRequest` to vanilla Javascript. Hope that's ok

faizanakram99:
Not directly related to this pr, but why doesn't `hx-on::confirm` on a parent element work (read would it work after this PR)?

I'd like to avoid any custom JabbaScript and still use a different confirmation than window.confirm.

alexpetros:
> Not directly related to this pr, but why doesn't hx-on::confirm on a parent element work (read would it work after this PR)?

`hx-on` is not inherited, if that's what you mean

faizanakram99:
> > Not directly related to this pr, but why doesn't hx-on::confirm on a parent element work (read would it work after this PR)?
> 
> `hx-on` is not inherited, if that's what you mean

it doesn't even work on the element that triggers event, in short how to use hx-on::confirm with hx-get/post etc to use sweetalert for example?

matiboy:
> > Not directly related to this pr, but why doesn't hx-on::confirm on a parent element work (read would it work after this PR)?
> 
> `hx-on` is not inherited, if that's what you mean

Note that `hx-confirm` on a parent element does work since we use `getClosestAttributeValue` in the code

matiboy:
> > > Not directly related to this pr, but why doesn't hx-on::confirm on a parent element work (read would it work after this PR)?
> > 
> > 
> > `hx-on` is not inherited, if that's what you mean
> 
> it doesn't even work on the element that triggers event, in short how to use hx-on::confirm with hx-get/post etc to use sweetalert for example?

Hi @faizanakram99 I've created a few examples [in this Codepen](https://codepen.io/matiboy/pen/oNQOdRp) it shows that you can use `hx-on=` or `hx-on::confirm` successfully. 

It also shows an example of how to do Sweetalert without hyperscript. However this is before my PR so you have to use the same trick as the [example available on the docs](https://htmx.org/examples/confirm/) (look at the code on the demo button, not the code in text) which is to react on click, then trigger the `hx-trigger` event of the button.

1cg:
Thank you @matiboy can I ask you to add some docs on the hx-confirm.md page as well?

matiboy:
> Thank you @matiboy can I ask you to add some docs on the hx-confirm.md page as well?

Hi @1cg the PR already contains the edits to `hx-confirm.md` and a revamp of the `examples/confirm.md` to reflect the changes.

Please let me know if you need me to add anything else.

hakosl:
The vanilla js example at: [htmx.org/examples/confirm](https://htmx.org/examples/confirm/) seems to trigger both the native alert(), as well as the Swal alert, and i seems to also be having the same issue locally.


image

Additionally(I don't know if it's related to this PR) I am having issues with the "htmx:confirm" event being triggered on other regular events, like clicking a hx-boost link

1cg:
Dang it. I will roll this change back and cut a new release tonight. Sorry about that.

Telroshan:
Ok so it seems we just have a problem about our release script, as the website's htmx file seems to be in version 1.9.5, i.e. before that PR was merged (thus before the fix), as it came in the latest release (1.9.7).
If you run it locally, you should get the expected behavior
Investigating on the release process, but the good news is we shouldn't need to revert those changes 😄

matiboy:
As the author of the PR, please let me know if there is indeed an issue with the change and whether you need me to investigate further please; thanks

matiboy:
> Additionally(I don't know if it's related to this PR) I am having issues with the "htmx:confirm" event being triggered on other regular events, like clicking a hx-boost link

Hi @hakosl could you give a quick codepen repro or describe a bit more so I can investigate this issue please? Thanks!

matiboy:
Oh I think I see what you're referring to @hakosl Yes `htmx:confirm` actually triggers on everything (somewhat surprisingly) but that behavior existed before the PR. When I fixed this bug and proposed to enhance the `details` prop, I did not suggest to change that behavior since it would be breaking.

1cg:
no worries it was a bug in our web gen script that got the version of htmx on the website out of sync with the /arc directory

Nothing to do with your change sorry for the alarm!

hakosl:
Agreed, with htmx 1.9.7, it works as expected the. "htmx:confirmed" event is still being triggered on all elements, but thats a different issue. 
[https://codepen.io/h-kon-osland/pen/JjxEawK](https://codepen.io/h-kon-osland/pen/JjxEawK)

abe-101:
> Oh I think I see what you're referring to @hakosl Yes `htmx:confirm` actually triggers on everything (somewhat surprisingly) but that behavior existed before the PR.

Any workaround for this bug?

abe-101:
I came up with this solution:

```js
document.addEventListener("htmx:confirm", function(e) {
    // Check if the question is null or empty
    if (!e.detail.question) {
        // If it is, issue the request directly
        e.detail.issueRequest(true); // use true to skip window.confirm
    } else {
        // If not, show the Swal confirmation dialog
        e.preventDefault();
        Swal.fire({
            title: "Proceed?",
            text: `${e.detail.question}`,
            icon: "warning",
            showCancelButton: true,
            confirmButtonText: "Yes",
            cancelButtonText: "No",
            reverseButtons: true
        }).then(function(result) {
            if(result.isConfirmed) e.detail.issueRequest(true); // use true to skip window.confirm
        });
    }
});
```

c-nv-s:
so is there a ticket anywhere for the fact ``htmx:confirm`` is triggering on everything? which is buggy behavior, especially with the documented sweetalert example if you have other htmx elements on the page that are triggered ``hx-get="/something" hx-trigger="load"`` then it triggers the sweet alert confirm message on every page load.

1cg:
the `htmx:confirm` event is always triggered on every request.  this is by design: it provides a scripting mechanism to cancel or delay a request that is integrated with the queueing mechanism in htmx.  it has nothing to do with the hx-confirm attribute, which is a simple way to use a browser-based `confirm()` dialog.

the current sweet alert example that I'm familiar with uses a custom event:

https://htmx.org/examples/confirm/

and doesn't use the `htmx:confirm` event

abe-101:
> the `htmx:confirm` event is always triggered on every request. this is by design: it provides a scripting mechanism to cancel or delay a request that is integrated with the queueing mechanism in htmx. it has nothing to do with the hx-confirm attribute, which is a simple way to use a browser-based `confirm()` dialog.
> 
> the current sweet alert example that I'm familiar with uses a custom event:
> 
> https://htmx.org/examples/confirm/
> 
> and doesn't use the `htmx:confirm` event

Can you clarify where on the page you are reffering to?

the example shown at https://htmx.org/examples/confirm/#vanilla-js-hx-confirm
does use the `htmx:confirm`, am I missing something?

PR #1611, opened 2023-07-18
Fix hx-on:* examples to use kebab-case event names, add clarification

ElnuDev:
Currently the examples in the documentation for `hx-on:*` include camelCase event names, which as the documentation says itself won't work due to DOM attributes not preserving casing. I've updated the events in the examples from camelCase to kebab-case and added extra clarification to the documentation.


alexpetros (approved):
Great catch, thank you!

PR #1613, opened 2023-07-19
[docs] separate documentation for the extended CSS selector syntax

### LINKED ISSUE #1601: [Documentation] Common extended selectors doc
As discussed in #1522 and #1478 with @alexpetros 

Several attributes use the extended selectors set, such as `hx-target`, `hx-include`, `hx-indicator`, and `hx-trigger`'s `from` clause.

As of today though, there's no common documentation for that selectors syntax, it's up to each attribute's doc to give details about it or not.
As such, docs like [`hx-include`](https://htmx.org/attributes/hx-include/) don't mention them at all, while a doc like [`hx-indicator` ](https://htmx.org/attributes/hx-indicator/) only mentions `closest`.

All of those attributes, even `hx-trigger`'s `from` clause, result in a call to [`querySelectorAllExt`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L583).

It would be nice to provide a separate documentation for that extended CSS selectors syntax (especially as we'll probably add more features to it in the future, such as #1597), and link to that doc in every attribute using a selector.

xhaggi:
This adds a separate documentation for the extended CSS selector syntax. 

In the first commit, I moved the handling of the keywords `window` and `document` from the `querySelectorAllExt` function to the `hx-trigger` handling, as they are both only valid for `hx-trigger="<event> from:<document|window>"`. This is also clarified in the documentation.

Fixes #1601

### DIFF
--- a/www/content/attributes/hx-include.md
+++ b/www/content/attributes/hx-include.md
@@ -2,8 +2,10 @@
 title = "hx-include"
 +++
 
-The `hx-include` attribute allows you to include additional element values in an AJAX request.  The value of
- this attribute is a CSS query selector of the element or elements to include in the query.
+The `hx-include` attribute allows you to include additional element values in an AJAX request.
+
+The value of this attribute is an extended CSS selector that matches the elements whose values should
+be included in the request. See [Extended CSS Syntax](@/extended-css.md) for more details.
 
 Here is an example that includes a separate input value:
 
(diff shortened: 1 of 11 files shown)


alexpetros:
Really appreciate the enthusiasm, but we're going to hold off on this one for now because there are a couple PRs in flight (referenced in the original issue) that need to happen before we can address this.

Issue #1619, opened 2023-07-21
Add error swapping attributes to the core


Telroshan:
As discussed in [the Discord](https://discord.com/channels/725789699527933952/1046573806547910677/1131673935453360221)

I'd like to suggest adding the possibility to swap content on errors, to the core of htmx. Not as fine-grained as the [response-targets extension](https://htmx.org/extensions/response-targets/) that lets you swap for specific HTTP codes, but a more generic "swap this there in case of error" approach.
Btw, correct me if I'm wrong, but there doesn't seem to be a way with response-targets to swap "in case of any error" ? From what I understand, you'd have to define a 4x wildcard as well as a 5x wildcard. So maybe this generic "error target/swap" could be in the core, and coexist with response targets that really lets you specify behaviors for precise response codes ?

Imo (for what it's worth), htmx should encourage its users to use proper HTTP response codes, yet using a 200 response to render an "invalid credentials" or "email already taken" error message, or basically any field value validation _(which is a very common need)_ seems wrong to me

Something like
```html
<div id="content">Initial content</div>
...
<form hx-post="/do/something" hx-target="#content" hx-swap="innerHTML" hx-error-target="find .feedback" hx-error-swap="outerHTML">
	<input type="text" name="X"/>
	...
	<div class="feedback"></div>
</form>
``` 
Where, on error, the server would respond with
```html
<div class="feedback error">Invalid value for parameter X</div>
```
Or any internal error really, such as
```html
<div class="feedback error">File upload failed. Please try again</div>
```
Hope you get the idea!

To make a rough summary of the first opinions expressed on the Discord:
- @1cg
> i always figured people would handle those situations via scripting but maybe that's unreasonable

> i can be convinced
- @Renerick
> aside from 204 disabling swapping, every other response code should swap as specified by default. This too is up for discussion

Also suggested `htmx.config.defaultErrorSwap` and `htmx.config.defaultErrorTarget` config parameters

- @alexpetros 
> "what belongs in scripting" is a super delicate balance but I think htmx right now is just a hair too reliant on hyperscript for common UI patterns

> this is a default that causes people a lot of headaches (I've actually updated the docs to signpost it more) but I think changing it would be hard

-------
Along with this issue, I submitted [a PR](https://github.com/bigskysoftware/htmx/pull/1620) that addresses the issue, here are the highlights of the suggested changes:
- 2 new attributes: `hx-error-target` and `hx-error-swap` attributes
[...]

PR #1620, opened 2023-07-21
hx-error-swap and hx-error-target attributes #1619

Telroshan:
This PR addresses #1619, see issue for the context

here are the highlights of the suggested changes:
- 2 new attributes: `hx-error-target` and `hx-error-swap` attributes
They both use the same syntax as, respectively, [hx-target](https://htmx.org/attributes/hx-target/) and [hx-swap](https://htmx.org/attributes/hx-swap/)
- They both support an additional value, that I named `mirror` _(feel free to suggest names, it's already hard enough to name things in my mother tongue, send help)_, which indicates htmx to swap errors _(that would otherwise be ignored by default)_ using the same specifications as their standard attribute counterpart
    - Thus, `hx-error-target="mirror"` would use the same target as defined in `hx-target`, or, if no `hx-target` was specified in the hierarchy, fallback to the default target (being the element itself)
    - `hx-error-swap="mirror"` would use the same swap spec as defined in hx-target, or, if no `hx-swap` was specified in the hierarchy, fallback to [config.defaultSwapStyle](https://htmx.org/api/#properties).
        - Note here that it defaults to `defaultSwapStyle` and not the new config parameter `defaultErrorSwapStyle`, for which you would just have to set no value at all, or an invalid one
- New config param: `htmx.config.defaultErrorSwapStyle`, that defaults to "none". This default value allows htmx to keep the behaviour it had, thus avoiding breaking changes, if we were to add that feature to htmx1
- New config param: `htmx.config.defaultErrorTarget`, that defaults to "mirror". This default value allows "lazy errors handling" so that by default, the errors would be swapped at the same target you had specified with the standard `hx-target` _(given that you have defined an error swapping strategy other than "none" ofc, whether it's globally or on the element triggering the request)_
- This way, if you were to set both those 2 new config parameters to "mirror", you would get content swapped like you defined using `hx-target` and `hx-swap` everywhere in your code, no matter if the request was successful, and resulted in an error
- Added a bunch of tests, most of them are `hx-target` and `hx-swap`'s tests adapted to their error counterpart. Also added tests for these special `mirror` values. All tests pass, though let me know if you think of any edge case that isn't included there

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -738,11 +738,12 @@ return (function () {
         /**
          * @param elt
          * @param initialTarget
+         * @param swapOverride
          * @returns {{shouldSwap: boolean, target: (HTMLElement|null), targetStr: string}}
          */
-        function getErrorTargetSpec(elt, initialTarget) {
+        function getErrorTargetSpec(elt, initialTarget, swapOverride) {
             var targetStr = getClosestAttributeValue(elt, "hx-error-target") || htmx.config.defaultErrorTarget
-            var swapStr = getClosestAttributeValue(elt, "hx-error-swap") || htmx.config.defaultErrorSwapStyle
+            var swapStr = swapOverride || getClosestAttributeValue(elt, "hx-error-swap") || htmx.config.defaultErrorSwapStyle
             if (!swapStr || swapStr === "none") {
                 return {
                     shouldSwap: false,
@@ -1133,7 +1134,7 @@ return (function () {
                         swapInnerHTML(target, fragment, settleInfo);
                     } else {
                         var defaultSwapStyle = htmx.config.defaultSwapStyle
-                        if (isStandardError && htmx.config.defaultErrorSwapStyle !== "mirror") {
+                        if (isStandardError && swapStyle !== "mirror" && htmx.config.defaultErrorSwapStyle !== "mirror") {
                             defaultSwapStyle = htmx.config.defaultErrorSwapStyle
                         }
                         swap(defaultSwapStyle, elt, target, fragment, settleInfo, isStandardError);
@@ -3495,7 +3496,10 @@ return (function () {
             // interfere with hx-error-target & hx-error-swap logic by relying solely on isError
             var isStandardErrorSwap = false
             if (isError && (htmx.config.httpErrorCodesToSwap.length === 0 || htmx.config.httpErrorCodesToSwap.indexOf(xhr.status) >= 0)) {
-                var errorSwapSpec = getErrorTargetSpec(elt, target);
+                if (etc.errorTargetOverride) {
+                    responseInfo.target = etc.errorTargetOverride
+                }
+                var errorSwapSpec = getErrorTargetSpec(elt, responseInfo.target, etc.errorSwapOverride);
                 if (errorSwapSpec.shouldSwap) {
                     shouldSwap = true;
                     isStandardErrorSwap = true
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -74,9 +74,7 @@ return (function () {
                 getCacheBusterParam: false,
                 globalViewTransitions: false,
                 methodsThatUseUrlParams: ["get"],
-                selfRequestsOnly: false,
-                defaultErrorSwapStyle: "none",
-                defaultErrorTarget: "mirror",
+                selfRequestsOnly: false
             },
             parseInterval:parseInterval,
             _:internalEval,
@@ -740,9 +738,9 @@ return (function () {
          * @returns {{shouldSwap: boolean, target: (HTMLElement|null), targetStr: string}}
          */
         function getErrorTargetSpec(elt, initialTarget, swapOverride) {
-            var targetStr = getClosestAttributeValue(elt, "hx-error-target") || htmx.config.defaultErrorTarget
-            var swapStr = swapOverride || getClosestAttributeValue(elt, "hx-error-swap") || htmx.config.defaultErrorSwapStyle
-            if (!swapStr || swapStr === "none") {
+            var targetStr = getClosestAttributeValue(elt, "hx-error-target")
+            var swapStr = swapOverride || getClosestAttributeValue(elt, "hx-error-swap")
+            if ((!targetStr && !swapStr) || swapStr === "none") {
                 return {
                     shouldSwap: false,
                     target: null,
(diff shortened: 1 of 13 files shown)


src/htmx.js
```diff
@@ -1025,7 +1065,7 @@ return (function () {
             return fragment;
         }
 
-        function swap(swapStyle, elt, target, fragment, settleInfo) {
+        function swap(swapStyle, elt, target, fragment, settleInfo, isStandardError) {
```
Renerick, line 1068: I'm a bit suspicious about passing error flag to `swap` method. Ideally, the decision to use the default value should come from the use site (which would be the `doSwap` method), but I suppose this calls for larger refactoring

src/htmx.js
```diff
@@ -73,6 +73,8 @@ return (function () {
                 getCacheBusterParam: false,
                 globalViewTransitions: false,
                 methodsThatUseUrlParams: ["get"],
+                defaultErrorSwapStyle: "none",
+                defaultErrorTarget: "mirror",
```
Renerick, line 77: Controversial idea - allow functions as configuration values. Function is evaluated at runtime and the returned value is used as corresponding configuration value. Removes the need for these reserved magic values, but maybe a bit of an overkill for what is essentially a single use-case?

Renerick (commented):
Just some thoughts out loud. Great PR @Telroshan, helps a lot!

src/htmx.js
Telroshan: Yeah that's because the swap function gets `htmx.config.defaultSwapStyle` in the fallback case, and I needed to differentiate from the error counterpart when relevant here
Couldn't find an elegant solution to work around this issue, other than passing that boolean around, which isn't that elegant either I know 😆

src/htmx.js
Telroshan: I mean, why not adding function on top of that, but I'm not convinced by them **replacing** the magic value.

My goal with those values was to provide an easy, quick & lazy way for the user to setup a global errors swapping strategy, so that it's easy to toggle them. Just set double `mirror` and you'll have errors everywhere, or let `defaultErrorTarget` to mirror then you'll only need to specify `hx-error-swap` and the target will follow `hx-target` everywhere, that kind of thing.

To reproduce the `mirror` logic with a function and without that magic value, you'd have to code yourself the retrieval of standard swap/target strategy (if applicable) in your project, so that doesn't seem to remove the need in my opinion

Also, wouldn't the preferred way for that kind of thing, be to maybe refactor the `htmx:beforeSwap` and add more parameters to it, so that the dev can override the error swap parameters if they want for specific requests, using that even instead of a function global config?

Config is also supposed to be preferably set through meta tags, though maybe is that not relevant for specific use cases where you'd want your users to write some custom JS to set up their own thing anyway?

Renerick: Very valid points, I especially forgot about meta tags, don't think there is a good way to mix them with functions.

Telroshan:
Added the idea we talked about yesterday @alexpetros 
Only adds 2 lines of code !

src/htmx.js
Telroshan: @Renerick see my [latest commit](https://github.com/bigskysoftware/htmx/pull/1620/commits/665f3de8f2dd580184f83f7345c66d179b5703e9), I prefer that new version, though it still requires passing a parameter down the way, but now it's the default swap style instead of a boolean, let me know what you think about that

alexpetros:
> Added the idea we talked about yesterday @alexpetros (whitelist config param of error status codes to swap) Only adds 2 lines of code !

Love this—would suggest you do it in a separate PR, as this one has a lot of new features and configs and will probably go through some revisions.

Telroshan:
Wouldn't it make more sense to have them all in the same PR ? Since the error codes whitelist cannot exist without this PR, and really adds 2 lines of code to it.
If it's just about patience, no worries I can wait 😆

I would say, if this goes through some revisions, I think it'd better be looked at as a whole, as any required change might impact that whitelist system, so I'd suggest revisions of the whole thing instead of 95% + 5% apart in a second phase, if that makes sense

alexpetros:
> Since the error codes whitelist cannot exist without this PR, and really adds 2 lines of code to it.

Why is this true? We could totally do this separately. 

In a broader sense, it lets us put the proper attention on each subset of the interface. A 2-source-line PR that introduces a new config name, documentation, and tests is already a good-sized PR! Once something is in our interface it is in there _forever_ so even the small API changes need scrutiny; such is the nature of maintaining a library.

Telroshan:
Well, this adds a whitelist to the error swapping system, so if the error swapping system were to be rejected, this whitelist wouldn't make any sense on its own

As you want then, I'll remove the whitelist from this PR
Should I wait until this PR eventually gets merged (if not closed) before making the PR for the whitelist then? Or would you like me to fire it as a second PR right away?

alexpetros:
I can push a branch and show you what I mean. Also there's no rush because this PR isn't going to be in the upcoming release.

Telroshan:
If you have the time for it, I'd be glad to have an example, I must admit I'm a bit confused about what you expect me to do!

Yeah I knew that, no worries!

Telroshan:
Forgot about this, removed the whitelist from the PR @alexpetros

ethanniser:
is there a reason this ran out of steam?

I have been using the `response-targets` extension, and found it difficult being forced to use the same `hx-swap` value for error and success swaps

But this pr looks even better, most people dont need the fine grainedness of `response-targets`, just success or error

would be happy to help to get this merged if others are still interested

Telroshan:
It's just that there were (and still are) a lot of PRs to process, so PRs with lots of changes always take some time
Though now that 1.9.6 is out of the way, I think this can be looked at for the next release, thanks for the ping!

alexpetros:
This PR turned out to be a bit more than we can chew in a single PR. After much discussion, including with @Telroshan, this is how we're going to break it down:

* We're going to merge #1929. In addition to being a much, much smaller change, I also prefer `hx-target-error` to `hx-error-target`, because it extends the existing namespace, and aligns with the `hx-target-5xx` schema of the response targets extension.
* We're going to discuss merging the response targets extension into the core, since it's one of the most commonly requested features. I personally support
* Evaluate `htmx.config.defaultErrorSwapStyle` as a config option in a separate PR
* Evaluate `htmx.config.defaultErrorTarget`

I want to reiterate that we are much interested in getting better error swapping into the core. It's a pain point for myself, and may others, as demonstrated by the interest in this PR and the parent issue. Core changes are *irreversible*; barring extreme cases, if an attribute or config options is merged into the core, we're supporting it forever. So having to make this many irreversible changes on PR overwhelmed our bandwidth. I think this path forward will unblock the issue and get us some better error swapping soon.

Issue #1564, opened 2023-07-11
Support `5xx` syntax for response-targets


spiffytech:
It appears that Rust's [`Maud`](https://maud.lambda.xyz/) HTML templating tool doesn't support asterisks in HTML attribute names, as used in the `response-targets` extension.

I wouldn't be surprised if there are other tools that don't support the syntax, given I can't recall seeing it used anywhere else.

I considered asking Maud to add support, but it looks like the asterisks might violate the HTML spec: while HTML built-in attributes [seem allowed](https://stackoverflow.com/questions/925994/what-characters-are-allowed-in-an-html-attribute-name) to use them, `data-` attributes and attributes on custom elements can't.

Given that, I request that the `response-targets` extension support an alternate wildcard syntax, `hx-target-5xx`, that's unambiguously HTML-compliant.

alexpetros:
> data- attributes and attributes on custom elements can't.

Do you have a citation for that? [Looks like it's allowed](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2) to me.

spiffytech:
Here you go!

[Data attributes](https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes):

> 3.2.6.6 Embedding custom non-visible data with the data-* attributes
> 
> A custom data attribute is an attribute in no namespace whose name starts with the string "data-", has at least one character after the hyphen, is XML-compatible, and contains no ASCII upper alphas.

[Custom elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-core-concepts):

> Any namespace-less attribute that is relevant to the element's functioning, as determined by the element's author, may be specified on an autonomous custom element, so long as the attribute name is XML-compatible and contains no ASCII upper alphas.

The relevant part of [the XML spec](https://www.w3.org/TR/xml/#NT-Name) is the `NameStartChar` and `NameChar` patterns (note that `*` is `0x002A`):

```
[...]

PR #1625, opened 2023-07-22
5xx-style wildcard support for response-targets extension

siefca:
See https://github.com/bigskysoftware/htmx/issues/1564

With this change it is possible to use custom wildards, so for example `hx-target-4xx` will match 404 response status code.


alexpetros:
Two concerns about this PR:

* I would like to give @spiffytech the opportunity to contribute since they are the one that originally raise the issue (if they're interested)
* We do not need custom wild cards. They add a lot of complexity to the extension for questionable value. I'd rather just see `x` made available as an alternative to `*` i.e. `5xx` instead of `5**`.

spiffytech:
> I would like to give @spiffytech the opportunity to contribute since they are the one that originally raised the issue, and they said they'd be interested. Obviously it's open source, and they may not have the time, but I would like to give them more than 2 days.

Thank you :heart: I plan to tackle this tomorrow.

siefca:
@alexpetros Ah, sorry, I misread the response and thought @spiffytech wants this feature badly, not wants to add the feature – my mind didn't associate "PRing" with "making a pull request" and I thought it meant that someone is very interested. :) lol. That's probably because I was beaten with hypermedia hrefs when I was little.

alexpetros:
Very much appreciate you opening it @siefca! From the maintainer side, we're going to keep getting better at signposting what issues people should jump on. Closing now in favor a much smaller-scoped PR.

PR #1627, opened 2023-07-23
[htmx 2.0] The overhaul of htmx's extensibility and feature separation

Renerick:
This PR is a proof of concept of brand new way to organize internal architecture and extensions in htmx 2.0. Drawing inspiration from _hyperscript, I introduce concepts of "features" and "feature sets" to group together related functionality of the core and provide clear, unambiguous extensibility points.

## But why?

Current implementation of extensions was, evidently, a later addition to htmx 1. While it did serve the purpose fairly well, the design of it left quite a bit to be desired. Have a look at this extension contract

```ts
export interface HtmxExtension {
    init(api): void;
    onEvent(name: string, evt: CustomEvent): boolean | void;
    transformResponse(text: string, xhr: XMLHttpRequest, elt: HTMLElement): any;
    isInlineSwap(swapStyle: string): boolean;
    handleSwap(swapStyle: string, target: HTMLElement, fragment: HTMLElement, settleInfo: any): HTMLElement[] | boolean;
    encodeParameters(xhr: XMLHttpRequest, parameters: object, elt: HTMLElement): XMLHttpRequestBodyInit;
}
```

Personally, here is a list of things I don't like:

1. Inconsistent extension points. This contract features:
   - `transformResponse`, `handleSwap`, `encodeParameters`, which are pretty obvious what their purpose are
   - `isInlineSwap`, which is called literally once when performing OOB swaps
   - `init` and `onEvent`, which can do literally anything (also fun fact, `onEvent` will always be called *last*, after every other event handler)
2. Inconsistent interactions with other htmx features
   - `handleSwap` will have the value of `hx-swap` attribute passed to the extension to match against (see morphdom and alpine-morph extensions). Extension is activated by a custom value of standard htmx attribute
   - `encodeParameters` will encode *any* request, ignoring values of `hx-encoding` and `enctype` (it has no obligations to read the value of hx-encoding, see json-enc extension).  Extension is activated by the presence of `hx-ext` attribute and has the power to override both default behavior and explicitly specified behavior
   - `tranformResponse` doesn't even have a corresponding htmx attribute or config value (see client-side-template extension).      Again, extension is activated by the presence of `hx-ext` attribute and has the power to do anything
3. Boilerplate extension invocation syntax. Every time htmx (or extensions) pass through an extension point, there has to be a call to `withExtensions` method which needs a lambda which actually calls the extension and performs a side effect of setting the result, and then there has to be a check if `withExtensions` did anything and if not, then core htmx will be called

## Goals

As such, I formulated this list of goals with this design

1. Extensions as first class citizens with clear place in htmx world
2. No boilerplate for extension invocation
3. No way to implicitly override defaults and core features
4. Provide straightforward extension points to create swaps, encodings, headers, response transformers and, ideally, attributes

## The outcome

**Disclaimer**: This is not final and not complete implementation. At this moment, the goal is to showcase the idea and evaluate if it's worth to consider moving forward. I converted only swaps and encodings to showcase how it would look and work in the core and in the extensions

Due to formatting and moving code around, the diff may be a bit messy (make sure to disable whitespace diff). Here is an overview of what's going on:

1. There is a new internal entity called `FeatureSet`, which carries within itself a `FeatureCollection`, which stores the `Feature`s. A `Feature` represents, well, a feature of htmx (in this implementation this includes swaps and encoding) and is itself a collection of named implementations of given feature (i.e., specific swap strategies or encodings)
2. Core htmx features (agains, swaps and encodings) are collected in the `coreFeatureSet` 
3. Extensions are implemented by registering their own feature collections.
5. Calls to `withExtensions` when passing through extension points were replaced by single `getFeature` method, which provides access to both core and extension features
      - if the feature implementation is defined in the core, it always has a priority over extensions. It's not possible to override `innerHTML` or `multipart/form-data`
      - otherwise, the method recursively collects extensions defined by hx-ext (similar to how `withExtenstions` does it) and from them retrieves the required implementation

There are some shortcomings of course:

1. It's not covered by tests at all (although everything else works (mostly))
2. There is no way to dynamically match attribute value (multi-swap extension relies on mini-syntax in `hx-swap`)
3. You probably will find other things

## But why not events?

In the discord, there was an idea of reworking extensibility to completely rely on events. While I like this idea, events carry additional semantics, such as bubbling, cancellation and execution order, which I find undesirable for extensions, especially if the extension wants to integrate with htmx's request pipeline. Also, they contradict my goals 1 and 2. Finally, htmx 2 includes exposing more public methods (such as swaps) in public API, and if the user wants to call `swap` with a swap provided by extension, events make it super non-trivial.

## This is too much

Soo... As I was preparing this I looked through the extensions. And there's a lot of them which really simple. They include a single and `onEvent` definition and that's it. And you know what? They are perfect. 

Like, how would you provide an extension point for [event-header](https://github.com/bigskysoftware/htmx/blob/master/src/ext/event-header.js), or [debug](https://github.com/bigskysoftware/htmx/blob/master/src/ext/debug.js), or [ajax-header](https://github.com/bigskysoftware/htmx/blob/master/src/ext/ajax-header.js), or [disable-element](https://github.com/bigskysoftware/htmx/blob/master/src/ext/disable-element.js)? Do you need an extension point for them? But then, how would you implement them? As attribute? Is there really a need to make a custom attribute for debug? Is it worth making user to both define `hx-ext` AND some `hx-debug` attribute to have it doing basically exactly the same? Maybe we could allow extensions to define some "root" "pseudo-attribute" which will be automatically set up on `hx-ext` but idk

And also, this is a big change. Like, complete overhaul of all internals and all extensions. While I believe it will be change for the best, it's still a lot of work within completely new paradigm. Breaking changes? Oh yes. Can it be adapted to the old extension syntax? Unlikely, but I haven't tried.

At least the HTML attributes should work the same

So, these thoughts above are the reason why I mentioned that I want to scratch everything and completely rework it. But as it stands now, I have no idea where to move. This PR is a product of about 1.5 weeks of thinking and prototyping, so I have relatively clear vision how this could be done, and I don't have an alternative with same level of clarity.

## So what if yes?

If we agreed that this is a worthy endeavor, then me and anyone who wishes to assist would work on migrating htmx features and extension to new framework, ensuring that all tests remain green with minimal (ideally zero) modifications. Ideally, we also adapt the simple extensions, mentioned above, to work with no modifications at all.

The drawback is that we would probably need a temporary feature freeze for 2.0, so that all existing features can be migrated and no new features will be lost

As I see it, such a massive refactoring would result in much more clear separation of different aspects of htmx, so then we can move forward with exposing those aspects as public API methods

## But what if maybe?

Of course, this is all subject to discussion and modifications. I fully recognize how radical this change is and will have no problems to tone it down a bit

## And what if no?

Back to the drawing board

## Why did you add `registerExtension`

So that `defineExtesion` keeps working while this is in progress, I'm not attached to the name

## Acknowledgments

- hyperscript.org for inspiration
- https://discord.com/channels/725789699527933952/1046573806547910677/1119478583493152870 for voicing this idea

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -8,7 +8,9 @@
         "lib": [ "dom" ]
     },
     "include": [
-        "./src/htmx.js"
+        "./src/htmx.js",
+        "./src/ext/*.js",
+        "./test/**/*.js",
     ],
     "verbose": true
 }
--- a/src/ext/alpine-morph.js
+++ b/src/ext/alpine-morph.js
@@ -1,16 +1,14 @@
-htmx.defineExtension('alpine-morph', {
-    isInlineSwap: function (swapStyle) {
-        return swapStyle === 'morph';
-    },
-    handleSwap: function (swapStyle, target, fragment) {
-        if (swapStyle === 'morph') {
+htmx.registerExtension('alpine-morph', (features) => {
+    features.addSwap('alpine-morph', {
+        isInlineSwap: true,
+        handleSwap: (target, fragment) => {
             if (fragment.nodeType === Node.DOCUMENT_FRAGMENT_NODE) {
                 Alpine.morph(target, fragment.firstElementChild);
-                return [target];
+                return { newElements: [target] };
             } else {
                 Alpine.morph(target, fragment.outerHTML);
-                return [target];
+                return { newElements: [target] };
             }
-        }
-    }
-});
+        },
+    })
+})
--- a/src/ext/alpine-morph.js
+++ b/src/ext/alpine-morph.js
@@ -1,14 +1,16 @@
-htmx.registerExtension('alpine-morph', (features) => {
-    features.addSwap('alpine-morph', {
-        isInlineSwap: true,
-        handleSwap: (target, fragment) => {
-            if (fragment.nodeType === Node.DOCUMENT_FRAGMENT_NODE) {
-                Alpine.morph(target, fragment.firstElementChild);
-                return { newElements: [target] };
-            } else {
-                Alpine.morph(target, fragment.outerHTML);
-                return { newElements: [target] };
-            }
-        },
-    })
+htmx.registerExtension('alpine-morph', {
+    swaps: {
+        'apline-morph': {
+            isInlineSwap: true,
+            handleSwap: (target, fragment) => {
+                if (fragment.nodeType === Node.DOCUMENT_FRAGMENT_NODE) {
+                    Alpine.morph(target, fragment.firstElementChild);
+                    return { newElements: [target] };
+                } else {
+                    Alpine.morph(target, fragment.outerHTML);
+                    return { newElements: [target] };
+                }
+            },
+        }
+    }
 })
--- a/src/ext/json-enc.js
+++ b/src/ext/json-enc.js
@@ -1,8 +1,8 @@
-htmx.registerExtension('json-enc', features => {
-    features.addEncoding('application/json', parameters => {
-        return {
+htmx.registerExtension('json-enc', {
+    encodings: {
+        'application/json': parameters => ({
             contentType: 'application/json',
             body: JSON.stringify(parameters)
-        }
-    })
+        })
+    }
 })
--- a/src/ext/json-enc.js
+++ b/src/ext/json-enc.js
@@ -1,12 +1,8 @@
-htmx.defineExtension('json-enc', {
-    onEvent: function (name, evt) {
-        if (name === "htmx:configRequest") {
-            evt.detail.headers['Content-Type'] = "application/json";
+htmx.registerExtension('json-enc', features => {
+    features.addEncoding('application/json', parameters => {
+        return {
+            contentType: 'application/json',
+            body: JSON.stringify(parameters)
         }
-    },
-    
-    encodeParameters : function(xhr, parameters, elt) {
-        xhr.overrideMimeType('text/json');
-        return (JSON.stringify(parameters));
-    }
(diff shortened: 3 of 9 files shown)


Telroshan:
I like the idea, makes it nice & easy to add new keywords for existing features through extensions or even in the core.
I'm having a hard time imagining it right now but, how would you concretely handle say the [response-targets](https://htmx.org/extensions/response-targets/) extension as an example, with such a system? Or any extension that needs to define new attributes really. How would that work with this feature set?

Renerick:
As I see it, a the `FeatureCollection` type defines `attributes` feature set. On load event htmx retrieves all registered attributes, generates a query selector and runs some hypothetical `init` function on matching elements. Attribute implementation then sets up appropriate event listeners and other things it needs to do. Allowing attributes to define `onEvent` method to simplify implementation is also an option (it would work exactly like current onEvent does, just attached to an attribute)

As such, implementation of response-target will likely stay the same. It would get wrapped in a feature collection like

```js
{
    attributes: {
        'response-target': { // TODO: dynamic attribute matching with functions/regexes
            onEvent: () => {}
        }
    }
}
```
but other than that, not much should change

I may look into prototyping this, if there is an iterest

Renerick:
Given that htmx 2.0 plans have been reviewed in favor of less drastic changes and even potentially not doing breaking updates at all, I'm going to close this PR for the time being. I don't want to clutter the PR listing with this, but my proposal still stands, so we may get back to it in the future.

Issue #1380, opened 2023-04-16
[proposal] hx-reset option to reset form after triggered


tdb-alcorn:
This is a feature proposal to add a new hx-reset field available on <form>s that would call form.reset() after the form's htmx behavior has been triggered (presumably, submit). In the use case where multiple form submissions are expected in sequence, this would obviate the need to send back an empty version of the form after receiving each submission. It would also preserve the active element and feel snappier in general.

The default behavior would be the same as now, that is without the option present, no reset would occur.

Thoughts?

andryyy:
Hm, what about something along hx-on "htmx:afterRequest if e.detail.successful then reset next form"?

Since you are sending the form and await an answer, you could also just resend the plain form, that’s not much overhead. You could also send a status message out-of-band with that response.

An extension to handle the reset by using a new attribute should be easy to create though.

jakespracher:
+1, sending the whole form back over the wire seems like overkill and having to implement your own extension as an alternative is not ideal. If I can find some time I may be able to work on this.

electroae4:
hx-on::after-request="this.reset()"

This way you can reset the form after submitting it.

Example for Django (Python):
<form hx-post="{% url 'comment_add' post.id %}"
    hx-target="#tab-contents"
    hx-swap="afterbegin"
    hx-on::after-request="this.reset()">
    {% csrf_token %}
    {{ form }}
    <button type="submit">Отправить</button>
</form>

dwasyl:
The only problem with these solutions is that when the `reset()` is called is that it doesn't necessarily change the form's state so if it previously displayed any errors, those same errors would still display with empty fields. Taking server-side validation into account and if anything did happen, there probably needs to be care to remove any error classes or hide any error messages from a previously failed submission.

calebbrgr:
I'm using `hx-post` on a form tag and on form.is_valid() I return a HttpResponse.
If you `hx-on::after-request="this.reset()"` the form tag the original form isn't cleared on page refresh.
With this method use `hx-on::before-request="this.reset()"`. The post request will validate first, then .rest(), finally the HttpResponse will be submitted back.

calebbrgr:
This should defiantly be a feature. However, I can see the complications for it to be implemented correctly.

This worked for me:

`<form hx-post="/contact/" hx-swap="outerHTML" hx-on:htmx:before-swap="if (event.detail.shouldSwap) this.reset(); else alert('Message Failed!')">`

[...]

PR #1636, opened 2023-07-26
[enhancement] hx-reset attribute

matiboy:
Referring to #1380 

If the proposed behavior is approved, this PR has the following feature and corresponding tests:

1. No change in current behavior
2. if `[data-]hx-reset` is set on an element which triggers a request:
   1. without value: `hx-reset` or `hx-reset=""` 
      a) if the element is a form, reset it on successful request
      b) if element has a form as parent (checked via `element.form`), reset that form on successful request
   2. with value: use querySelector with said value and reset that element. if no element or element is not a form, do nothing
   3. If request is unsuccessful, do not reset form
3. if `hx-reset` is set on a form and one of its children triggers the request, reset the form on successful request

If this is approved, I'll help write the docs

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3440,6 +3440,22 @@ return (function () {
                             maybeCall(settleResolve);
                         }
 
+                        // hx-reset attribute
+                        var formElt
+                        if(hasAttribute(elt, "hx-reset")) {
+                            var resetTarget = getAttributeValue(elt, "hx-reset");
+                            if(!resetTarget) {
+                                if(elt.tagName === "FORM") formElt = elt
+                                else if(elt.form) formElt = elt.form
+                            } else {
+                                formElt = document.querySelector(resetTarget)
+                            }
+                        } else if(elt.form && elt.form.hasAttribute("hx-reset")) {
+                            formElt = elt.form
+                        }
+                        
+                        if(formElt && formElt.reset) formElt.reset()
+
                         if (swapSpec.settleDelay > 0) {
                             setTimeout(doSettle, swapSpec.settleDelay)
                         } else {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3450,7 +3450,7 @@ return (function () {
                             } else {
                                 formElt = document.querySelector(resetTarget)
                             }
-                        } else if(elt.form && elt.form.hasAttribute("hx-reset")) {
+                        } else if(elt.form && hasAttribute(elt.form, "hx-reset")) {
                             formElt = elt.form
                         }
                         
--- a/test/attributes/hx-reset.js
+++ b/test/attributes/hx-reset.js
@@ -1,0 +1,168 @@
+describe("hx-reset attribute", function() {
+    beforeEach(function () {
+        this.server = makeServer();
+        clearWorkArea();
+    });
+    afterEach(function () {
+        this.server.restore();
+        clearWorkArea();
+    });
+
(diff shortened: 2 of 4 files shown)


Telroshan:
Omg I didn't even know [form.reset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reset) was a thing, very nice!
I love the idea, I'm often using this pattern of resetting a form after a successful submit, it'd be a great addition!

alexpetros:
Is this a use-case that would be solved by the following:
```html
<form hx-put="/user" hx-on::after-request="this.reset()">
...
</form>
```

jakespracher:
@alexpetros Oh that does look nice. For what it’s worth: I was not able to figure this out as a succinct alternative. Most suggestions on the internet are to either send a whole new form over the wire or to write more js than that. For example:

https://stackoverflow.com/questions/70043237/how-do-we-make-a-htmx-response-trigger-a-form-reset

If you choose to reject this PR on the grounds that the above is sufficient, it might be worth calling that out in the docs somewhere as this seems like a common pattern. It was the very first dilemma I ran into when trying out HTMX.

alexpetros:
@jakespracher you are totally correct, and I also wasn't aware of `reset()` until I read this PR so I wouldn't have been able to come up with it either. I'll test it and then answer that stackoverflow question.

We're at a place with htmx where we have to be pretty careful not to add a bunch more attributes without fully exploring whether the existing design surface adequately covers them. With htmx in particular, the existing surface also includes bunch of HTML and JS features that are rarely used because they are not declarative, and learning how to write good htmx involves making use of them. So there's a huge missing space for an "htmx cookbook" to make patterns like that clear, and we're still (hopefully) in the early stages of documenting and onboarding people to them.

matiboy:
@alexpetros Please feel free to reject/close the PR if you feel it bloats htmx rather than adding value; no issue at all.

Yes a cookbook would be really good actually.

FrenchTechLead:
is this ever gonna be merged ?

andryyy:
I guess @alexpetros alternative approach is pretty solid coming with no changes to the code, while the PR remains a nice contribution anyway. Don't want to play it down. 🙃

alexpetros:
This is probably not going to be merged, for the reasons discussed above—an equally ergonomic solution exists with existing attributes. The reason I've left it open is to remind myself to put an example somewhere on the documentation, like @jakespracher suggested.

matiboy:
@alexpetros As mentioned above, I don't mind this being dropped if the team feels it can easily be achieved some other way, however I want to point out that your solution above triggers this.reset regardless of successful or failed request which is probably not desired. You can see [this reproduced here](https://codepen.io/matiboy/pen/GRPRZJr)

Your solution can likely be extended a bit to achieving the "hx-reset" pattern; if you'd like to assign #1698 I can look into that, things like not reacting on error, targeting another form etc?

alexpetros:
Feel free to submit a PR for it :)

tedyoung:
For those coming here for a concise solution for clear-only-on-success, issue #1698 has the solution:

`<form hx-on::after-request="if (event.detail.successful) this.reset()">`

mihouami:
hx-on::after-request="this.reset()"

Is there a way to use this only with post requests ? Not all requests.

calls9-amirbraham:
Is there a way to achieve this while respecting CSP constraints ?

PR #1639, opened 2023-07-28
Update disable-element.js to allow for multiple elements to be selected

imaperson1060:
Allow selection of multiple elements for toggling

### DIFF
--- a/src/ext/disable-element.js
+++ b/src/ext/disable-element.js
@@ -5,12 +5,14 @@ htmx.defineExtension('disable-element', {
     onEvent: function (name, evt) {
         let elt = evt.detail.elt;
         let target = elt.getAttribute("hx-disable-element");
-        let targetElement = (target == "self") ? elt : document.querySelector(target);
+        let targetElements = (target == "self") ? elt : document.querySelectorAll(target);
 
-        if (name === "htmx:beforeRequest" && targetElement) {
-            targetElement.disabled = true;
-        } else if (name == "htmx:afterRequest" && targetElement) {
-            targetElement.disabled = false;
-        }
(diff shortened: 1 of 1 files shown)


src/ext/disable-element.js
```diff
@@ -5,12 +5,14 @@ htmx.defineExtension('disable-element', {
...
         let target = elt.getAttribute("hx-disable-element");
-        let targetElement = (target == "self") ? elt : document.querySelector(target);
+        let targetElements = (target == "self") ? elt : document.querySelectorAll(target);
 
-        if (name === "htmx:beforeRequest" && targetElement) {
-            targetElement.disabled = true;
-        } else if (name == "htmx:afterRequest" && targetElement) {
-            targetElement.disabled = false;
-        }
+        targetElements.forEach(targetElement => {
```
alexpetros, line 10: We can't use forEach because it's not IE11 compatible.

alexpetros (commented):
Could you add a multiple-element test please? Otherwise looks looks good

alexpetros:
Ah it looks like it broke the existing test too. You can run `npm run test` locally, using node 15

imaperson1060:
> Ah it looks like it broke the existing test too. You can run `npm run test` locally, using node 15

I'll make a new PR following the contributing guide (which I may have forgotten to read last time), but the first test seems to already be broken with the current extension (https://github.com/bigskysoftware/htmx/blob/be85c2c/test/ext/disable-element.js#L25). In the console it is saying that `targetElement` is not defined (https://github.com/bigskysoftware/htmx/blob/be85c2c/src/ext/disable-element.js#L7) - the only changes I made to the file started at line 8.

Issue #1640, opened 2023-07-28
Bug: Response Header `HX-Retarget` cause classList error, when `hx-target` attribute works fine


AjaniBilby:
I have a normal `a` tag link:
```jsx
<a
	href={"/server"}
	hx-get={"/server"}
	hx-push-url={"/server"}
	hx-headers='{"hx-headless": "true"}'
>{contents}</a>
```

This request's response has the following headers:
![image](https://github.com/bigskysoftware/htmx/assets/11359344/0e0e6511-6e05-402c-bab4-7c19a6949509)

When this request is recieved I get the following error:
![image](https://github.com/bigskysoftware/htmx/assets/11359344/c0819da1-c477-44a8-b7f4-3f18caac5b77)

If I then remove the `HX-Retarget` header from my response and add the exact same value as an attribute of my `<a>` for `hx-target`, I however get no errors, and it behaves correctly - however I want to be able to use `HX-Retarget` for some clever server side dynamic swapping.

Interestingly if I intentionally give an invalid id as the retarget (as in an ID that does not exist in the current HTML) I get this error - which is very interesting that it's a completely different from the error I'm actually seeing.
![image](https://github.com/bigskysoftware/htmx/assets/11359344/67ae179d-9bf8-434c-838c-cecbce21f38f)

I'm not sure of the best way to debug for diagnose this issue, nor if it is a fault with htmX or something I'm doing wrong.

matiboy:
This seems to happen if HX-Retarget contains "closest" but not if it is just the ID as seen[ in this reproduction code pen](https://codepen.io/matiboy/pen/MWzzvPQ)

Will investigate further on other "keywords"

matiboy:
Actually the documentation doesn't mention that HX-Retarget allows for the use the modifiers that hx-target attribute does. It says:

> A CSS selector that updates the target of the content update to a different element on the page

and the code is [indeed straight CSS selector;](https://github.com/bigskysoftware/htmx/blob/61945431aa72d44f244ecf69b5d188796bc0b7e5/src/htmx.js#L3298) so "closest" won't work on the headers.

This should probably be relabeled from Bug to Feature Request if you feel it should be implemented

AjaniBilby:
Ahh, okay my bad - somewhere along the way I assumed it would behave the same as `HX-Reswap` compared to `hx-swap`.
And since I am using a unique id anyway I can just remove closest and get the functionality I need.

[...]

PR #1641, opened 2023-07-29
[Draft] Proposed HX-Retarget-Root header

matiboy:
Referring to #1640 , it does feel as though HX-Retarget could behave like `hx-target` in the same type of relationship as HX-Reswap/hx-swap

Specifically, hx-target has modifiers `closest`, `find` etc, but `HX-Retarget` is a direct `querySelector` from `document`

For backwards compatibility's sake, this PR would introduce a `HX-Retarget-Root` header:

1. if unset, behavior is unchanged, HX-Retarget is a simple querySelector from `document`
2. if set, the value of `HX-Retarget` can use the modifiers e.g. `find div` and it is triggered not from `document` but from an element determined as follows:
    1. the element which initially triggered the request if `HX-Retarget-Root: this`
    2. an element found by `document.querySelector(xhr.getHeaderValue('HX-Retarget-Root'))` otherwise

This PR is not ready for merge at all, tests are not written and there is a manual test which would be removed.; it is meant to showcase the behavior 

I'll write tests/documentation if the  maintainers feel that this is of interest to the project.

Thanks!

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3294,8 +3294,20 @@ return (function () {
                 }
             }
 
-            if (hasHeader(xhr,/HX-Retarget:/i)) {
-                responseInfo.target = getDocument().querySelector(xhr.getResponseHeader("HX-Retarget"));
+            if (hasHeader(xhr, /HX-Retarget:/i)) {
+                var retargetSelector = xhr.getResponseHeader("HX-Retarget")
+                if (hasHeader(xhr, /HX-Retarget-Root:/i)) {
+                    var retargetRootSelector = xhr.getResponseHeader("HX-Retarget-Root")
+                    var retargetRoot
+                    if (retargetRootSelector === "this") {
+                        retargetRoot = elt;
+                    } else {
+                        retargetRoot = getDocument().querySelector(xhr.getResponseHeader("HX-Retarget-Root"))
+                    }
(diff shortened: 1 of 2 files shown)


matiboy:
@alexpetros I noticed you reviewed a series of PR but not this one; I set it to "Draft" to highlight that it's not at all ready for merge (no tests etc) but would still like your opinion on the concept and whether this would be useful, in which case I'd write tests/docs. Thanks!

alexpetros:
I understand! This is a tricky one. Not ready to comment because I need to give it more thought. I'll say that I don't particularly love the header basically being a switch flip for the behavior (it feels clunky) and that the `HX-Retarget` header having access to the entire document is a little more intuitive than you're giving it credit for. Could maybe be signposted in the documentation better but I'm not immediately convinced that this improvement is obvious enough that it warrants the backward compatibility mechanism.

matiboy:
Perfectly understandable; same as with my PR for `hx-reset` this was in response to an issue posted and just a proposal if needed. I personally have no need for this one, and even the OP said he wouldn't even really need it.

alexpetros:
> I personally have no need for this one, and even the OP said he wouldn't even really need it.

If that's the case I think we can close it for now! Appreciate you proposing it though.

PR #1650, opened 2023-07-30
Allow disable-element extension to disable multiple elements

imaperson1060:
Allow multiple elements to be selected in the query

Better version of #1639 - this time I actually read the contribution guide

### DIFF
--- a/src/ext/disable-element.js
+++ b/src/ext/disable-element.js
@@ -7,7 +7,7 @@ htmx.defineExtension('disable-element', {
         let target = elt.getAttribute("hx-disable-element");
         let targetElements = (target == "self") ? [ elt ] : document.querySelectorAll(target);
 
-		for (var i = 0; i < targetElements.length; i++) {
+        for (var i = 0; i < targetElements.length; i++) {
             if (name === "htmx:beforeRequest" && targetElements[i]) {
                 targetElements[i].disabled = true;
             } else if (name == "htmx:afterRequest" && targetElements[i]) {
--- a/src/ext/disable-element.js
+++ b/src/ext/disable-element.js
@@ -5,12 +5,14 @@ htmx.defineExtension('disable-element', {
     onEvent: function (name, evt) {
         let elt = evt.detail.elt;
         let target = elt.getAttribute("hx-disable-element");
-        let targetElement = (target == "self") ? elt : document.querySelector(target);
+        let targetElements = (target == "self") ? [ elt ] : document.querySelectorAll(target);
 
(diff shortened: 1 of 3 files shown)


imaperson1060:
For some reason, when I run `npm test`, the hx-boost attribute fails. It's not like this in the browser and I didn't change anything related to hx-boost.

image
image

test/ext/disable-element.js
```diff
@@ -59,4 +59,37 @@ describe("disable-element extension", function() {
         btn2.disabled.should.equal(false);
     });
 
+    it('disables multiple elements during htmx request', function () {
+		// GIVEN:
+		// - A button triggering an htmx request with disable-element extension
+		// - Another button that needs to be disabled during the htmx request
+		// - A control button that won't be disabled during the htmx request
+		// - All buttons are enabled
```
alexpetros, line 67: I appreciate the comments but I think the test is self-explanatory and it would be better to remove them :)

alexpetros (commented):
Great PR; strict, intuitive enhancement and I can't think of any reason why this would cause a regression.

test/ext/disable-element.js
imaperson1060: da89d61

src/ext/disable-element.js
```diff
@@ -5,12 +5,14 @@ htmx.defineExtension('disable-element', {
...
         let elt = evt.detail.elt;
         let target = elt.getAttribute("hx-disable-element");
-        let targetElement = (target == "self") ? elt : document.querySelector(target);
+        let targetElements = (target == "self") ? [ elt ] : document.querySelectorAll(target);
 
-        if (name === "htmx:beforeRequest" && targetElement) {
-            targetElement.disabled = true;
-        } else if (name == "htmx:afterRequest" && targetElement) {
-            targetElement.disabled = false;
+		for (var i = 0; i < targetElements.length; i++) {
```
1cg, line 10: indentation looks a little funky here.  tabs?

1cg (changes requested):

src/ext/disable-element.js
imaperson1060: Yeah, I noticed that in the GitHub editor. The other lines use spaces but I have my IDE set to use tabs instead. In most places the tab size is 4 spaces so it's not noticeable, but for some reason the GitHub editor has like 8.

alexpetros: Can you re-indent to match the spaces scheme of the project please?

PR #1656, opened 2023-08-01
Update same site security to use origin

adamckay:
As discussed in Discord, use `origin` to compare whether it's the same site to have same behaviour as CSP check with protocol, hostname and port.

Updated `sameHost` boolean to be true if it's the same site, too.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2848,10 +2848,10 @@ return (function () {
 
         function verifyPath(elt, path, requestConfig) {
             var url = new URL(path, document.location.href);
-            var hostname = document.location.hostname;
-            var sameHost = hostname !== url.hostname;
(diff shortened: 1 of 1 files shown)


src/htmx.js
```diff
@@ -2848,10 +2848,10 @@ return (function () {
 
         function verifyPath(elt, path, requestConfig) {
             var url = new URL(path, document.location.href);
-            var hostname = document.location.hostname;
-            var sameHost = hostname !== url.hostname;
+            var origin = document.location.origin;
+            var sameHost = origin === url.origin;
```
alexpetros, line 2852: nitpick but I think this would be even simpler as:
```js
var sameHost = document.location.origin === url.origin
```
then it's a little clearer what you're comparing

PR #1659, opened 2023-08-02
Add interactive check to ready event handler

dsyer:
If HTMX is imported in a module the readyState is "interactive" so the extension processing happens too late. This would fix it.

See #1655


alexpetros:
This seems good but could you explain the problem and your solution a little more thoroughly?

dsyer:
There's a sample here: https://github.com/scratches/htmx-demo. If you prefer Python there's a simpler version in the "django" branch. They both work as they are, but they use a hacked version of `htmx.js` with the change in this PR. If you switch to the regular HTMX (e.g. from unpkg.com) it will break because the extension isn't installed before the ready listener fires.

alexpetros:
Got it, thank you. You're right, and the async issue has come up before (#1485). I also think it make sense to initialize htmx last (after the extensions) anyway.

Just to make sure I understand fully, since there are only [three readyStates](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState), this is logically equivalent to following?

```js
if (getDocument().readyState === 'complete')
```

dsyer:
I guess it must be. You want to make that change?

guilleiguaran:
I've tested this with Vite/Rollup (as in https://github.com/bigskysoftware/htmx/issues/1469) and it's working as expected.

alexpetros:
Thank you both for verifying, this looks great. I'm not entirely sure why the CI is failing, because it only does so intermittently on my local machine. 

@1cg is it possible this timeout just needs to be longer?
https://github.com/bigskysoftware/htmx/blob/66387c04221b257b9cfa858db2636e6dd3c6fda7/test/core/api.js#L18-L28

1cg:
yeah, I'm going to merge this and I'll take a look at that test today

1cg:
actually, @dsyer can you make this PR against `dev` for me?  

thank you!

dsyer:
I don't think I know how to do that. I could close this one and open a new PR? Or you could walk me through it if it's possible otherwise and I'm just too stupid.

1cg:
lmao i have no idea how to do it either I just ask other people to do it

dsyer:
Sorry, I didn’t check if I needed to rebase before I made those PRs. I’m AFK now but I can do it tomorrow.

alexpetros:
> actually, @dsyer can you make this PR against `dev` for me?
> 
> thank you!

@dsyer what he means is that this PR is set to merge into the "master" branch and there's a setting (should be at the top, when you edit the PR) to chose to merge it against the "dev" branch instead (that's the one we use for commits that will be in the next release).

dsyer:
I know what he meant, but I don’t think there’s a way to change the base branch once the PR is submitted.

alexpetros:
Tried to change the branch but unfortunately the PR has a bunch of commits from `master` in it now, so I cherry-picked your commit and put it in PR #1672 (should still say you authored it).

alexpetros:
You're in: https://github.com/bigskysoftware/htmx/commit/347ce7a19f5c15a91a4fca166505e0b397b160ed

Sorry about the weirdness

PR #1662, opened 2023-08-02
Update splitting-your-apis.md

Zearin:
Add `abbr` element around “YAGNI”.

Since HTMX is a celebration of HTML, we might as well make use of the many tools for semantic enrichment that are available! 🤓


alexpetros:
Appreciate the enthusiasm (and we are about celebrating HTML!) but if we put an `abbr` tag on abbreviation we would have a huge task ahead of us (API, HTML, etc) so I don't think it makes sense.

Relevant link to the docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr#typical_use_cases

Zearin:
☹️ K.

PR #1672, opened 2023-08-04
Add interactive check to ready event handler

alexpetros:
If HTMX is imported in a module the readyState is "interactive" so the extension processing happens too late. This would fix it.

Rebased version of: #1659

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3605,7 +3605,7 @@ return (function () {
         //====================================================================
 
         function ready(fn) {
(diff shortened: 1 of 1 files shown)


1cg:
awesome, thank you @alexpetros and @dsyer!

xhaggi:
Does this make https://github.com/bigskysoftware/htmx/pull/1485 obsolete?

alexpetros:
I believe so, yes

PR #1674, opened 2023-08-05
[websocket] pass hx-target id in the HEADERS

Renerick:
Previously, ws.js used `ws-connect` element as target when collecting request headers. Now it correctly takes `hx-target` attribute into account


alexpetros:
@Renerick if you rebase against `dev` the tests should work again

Renerick:
@alexpetros I think it's somehow worse now lol

alexpetros:
@Renerick re-ran it, good to go.

Renerick:
Thank you!

PR #1676, opened 2023-08-05
[websocket] bug fix: convert blob and arraybuffer to string for processing

alittlebrighter:
The websocket connection was failing to swap because websocket messages coming from the server are either [blobs](https://developer.mozilla.org/en-US/docs/Web/API/Blob) or [arraybuffers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) which don't have the same methods as strings. This converts incoming messages to strings for further processing.

### DIFF
--- a/src/ext/sse.js
+++ b/src/ext/sse.js
@@ -50,7 +50,7 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 
 				// Try to create EventSources when elements are processed
 				case "htmx:afterProcessNode":
-					createEventSourceOnElement(evt.target);
+					ensureEventSourceOnElement(evt.target);
 					registerSSE(evt.target);
 			}
 		}
@@ -187,14 +187,14 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 	}
 
 	/**
-	 * createEventSourceOnElement creates a new EventSource connection on the provided element.
+	 * ensureEventSourceOnElement creates a new EventSource connection on the provided element.
 	 * If a usable EventSource already exists, then it is returned.  If not, then a new EventSource
 	 * is created and stored in the element's internalData.
 	 * @param {HTMLElement} elt
 	 * @param {number} retryCount
 	 * @returns {EventSource | null}
 	 */
-	function createEventSourceOnElement(elt, retryCount) {
+	function ensureEventSourceOnElement(elt, retryCount) {
 
 		if (elt == null) {
 			return null;
@@ -207,7 +207,7 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 				return;
 			}
 
-			createEventSource(child, sseURL, retryCount);
+			ensureEventSource(child, sseURL, retryCount);
 		});
 
 		// handle legacy sse, remove for HTMX2
@@ -217,12 +217,12 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 				return;
 			}
 
-			createEventSource(child, sseURL, retryCount);
+			ensureEventSource(child, sseURL, retryCount);
 		});
 
 	}
 
-	function createEventSource(elt, url, retryCount) {
+	function ensureEventSource(elt, url, retryCount) {
 		var source = htmx.createEventSource(url);
 
 		source.onerror = function(err) {
@@ -240,7 +240,7 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 				retryCount = retryCount || 0;
 				var timeout = Math.random() * (2 ^ retryCount) * 500;
 				window.setTimeout(function() {
-					createEventSourceOnElement(elt, Math.min(7, retryCount + 1));
+					ensureEventSourceOnElement(elt, Math.min(7, retryCount + 1));
 				}, timeout);
 			}
 		};


Renerick:
Thank you for this PR! I'll have a look at this within a day

src/ext/ws.js
```diff
@@ -122,12 +122,7 @@ This extension adds support for WebSockets to htmx.  See /www/extensions/ws.md f
...
 		});
 
-		socketWrapper.addEventListener('message', function (event) {
-			if (maybeCloseWebSocketSource(socketElt)) {
-				return;
-			}
-
-			var response = event.data;
+		var responseHandler = function (response) {
 			if (!api.triggerEvent(socketElt, "htmx:wsBeforeMessage", {
```
Renerick, line 126: The initial idea behind supporting binary formats was to give users simple way to manage websocket connections, while leaving actual message processing to their custom JS via events. By forcibly converting everything to strings, we disable this path for those who walk it. It's also a breaking change

I think, there are two ways to compromise:

1. Put this behavior behind configuration value, `false` by default
2. Offload this responsibility to event handler by allowing to modify `message` field below

src/ext/ws.js
```diff
@@ -151,6 +146,21 @@ This extension adds support for WebSockets to htmx.  See /www/extensions/ws.md f
...
+			}
+
+			var response = event.data;
+			if (socketWrapper.socket.binaryType === "blob") {
+				response.text()
+					.then(responseHandler);
+			} else { // arraybuffer
+				var responseStr = String.fromCharCode.apply(null, new Uint16Array(response));
+				responseHandler(responseStr);
+			}
```
Renerick, line 163: This change breaks support for `text` message type, which still can be `string`, regardless of socket's binary type (https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/data#value). I checked the test server in `test/realitime/server.go`, it doesn't work.

test/ext/ws.js
```diff
@@ -88,11 +88,36 @@ describe("web-sockets extension", function () {
         this.messages.length.should.equal(1);
     })
 
-    it('handles message from the server', function () {
```
Renerick, line 91: Related to `text` message types, this test was covering that. I tried to add it, but just like the test server, it was failing due to string values having no `text()` method. Please add it back and keep it green, it's primary use case of the extension.

test/ext/ws.js
```diff
@@ -88,11 +88,36 @@ describe("web-sockets extension", function () {
...
+        var buf = new ArrayBuffer(str.length*2); // 2 bytes for each char
+        var bufView = new Uint16Array(buf);
+        for (var i=0, strLen=str.length; i < strLen; i++) {
+            bufView[i] = str.charCodeAt(i);
+        }
+        return buf;
+    }
+
+    it('handles message from the server (arraybuffer)', function () {
+        htmx.config.wsBinaryType = "arraybuffer";
```
Renerick, line 115: I think this value should be reset to default between tests. Best way to do that is setting it to default at the top of the file in `afterEach` call

Renerick (changes requested):
Thank you again for your contribution. I reviewed your proposal and found that it introduces quite significant contract changes and breaks handling of `text` message type. Please see code comments for more details

src/ext/ws.js
alittlebrighter: Ah, my net was too wide. The `makeFragment` function assumes it is receiving a string and that is the code path I was having an issue with. I've changed my solution so the response only forcibly converts the response to a string for that function.

src/ext/ws.js
alittlebrighter: It looks like everything is working with the latest changes.

test/ext/ws.js
alittlebrighter: I added another test covering `text` message types.

alittlebrighter:
Okay, I believe I have constrained my changes to only change what was needed and allow the events that get emitted to retain the original data types. I had no idea that `text` was an option, the prompt for the PR was a toy app I was building with Go and I was trying to swap in rendered Go templates using a gorilla websocket handler which apparently comes down as a binary blob. This scenario was causing htmx to throw an error in the console due message data being binary and not a string.

Renerick:
Great! I'll look into this once I  free up some time from other things

test/ext/ws.js
```diff
@@ -88,12 +89,49 @@ describe("web-sockets extension", function () {
...
+        htmx.config.wsBinaryType = "blob";
+        var div = make('<div hx-ext="ws" ws-connect="ws://localhost:8080"><div id="d1">div1</div><div id="d2">div2</div></div>');
+        this.tickMock();
+
+        this.socketServer.emit('message', new Blob(["<div id=\"d1\">replaced</div>"]));
+
+        setTimeout(function() {
+            byId("d1").innerHTML.should.equal("replaced");
+            byId("d2").innerHTML.should.equal("div2");
+        }, 10);
```
Renerick, line 114: `setTimeout` should not be used for testing due to sinon.js mocking the timers in synchronous manner. Since Blob.text() callback relies on promises, I think it's only fitting to test it with promises as well

```suggestion
        this.tickMock();

        return new Promise((resolve) => div.addEventListener("htmx:wsAfterMessage", resolve))
            .then(() => {
                byId("d1").innerHTML.should.equal("replaced");
                byId("d2").innerHTML.should.equal("div2");
            });
```

test/ext/ws.js
```diff
@@ -88,12 +89,49 @@ describe("web-sockets extension", function () {
...
         this.tickMock();
 
         this.socketServer.emit('message', "<div id=\"d1\">replaced</div>");
 
+        setTimeout(function() {
+            byId("d1").innerHTML.should.equal("replaced");
+            byId("d2").innerHTML.should.equal("div2");
+        }, 10);
+    })
+
```
Renerick, line 103: This `setTimeout` call also won't work as expected


```suggestion
        this.tickMock()

        byId("d1").innerHTML.should.equal("replaced");
        byId("d2").innerHTML.should.equal("div2");
    })

```

Renerick (commented):
So sorry for such big delay, but finally I was able to review this. There were some points regarding testing of this feature (I've taken the liberty to provide code suggestions for them), but overall this is looking pretty good and I have no objections to including this in the extension.

@1cg @benpate , I ask for a second opinion on that one. Basically, if the socket receives binary data instead of a string and this data is not handled by an event handler, the extension will try to convert it to text and process normally, where previously it would fail with runtime error. To me this looks ok, but maybe some other behavior is deemed preferable for such cases

Unfortunately, this change is not compatible with IE11 at all due to [Blob.text()](https://caniuse.com/mdn-api_blob_text) and Promises. The extension will run fine as long as a Blob is not received by the socket, but the tests will break (CC: @Telroshan , I know you are currently working on a related PR)

alittlebrighter:
Thank you for the fixes. Sorry for the errors in tests, all of the tests were passing in Firefox and I didn't try it out on my windows machine.

alexpetros:
Okay I'd like to get some movement on this. Are you interested in rebasing it so we can run the test suite and put it up as a candidate for being merged?

PR #1677, opened 2023-08-05
Move load tests to the bottom of the core file

alexpetros:
## Description 
This "resolves" a timing bug that was occurring in the CI, where these tests would run before the htmx was ready (only in the GitHub actions servers). Not proud of this as a fix but I would like the CI to work again ASAP.

The bug is related to the change from these PRs: 
* https://github.com/bigskysoftware/htmx/pull/1672
* https://github.com/bigskysoftware/htmx/pull/1659

I still think the behavior in these PRs is correct. As best I can tell from [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState), the `loading` state happens before all the deferred scripts are loaded but `DOMContentLoaded` fires after deferred scripts are loaded, even though the MDN page says "[loading] state indicates that the DOMContentLoaded event is about to fire". Which is what we want—we want it to be ready when the scripts are loaded, but we don't care if all the images are loaded. 

What needs to probably happen in the future is that we hold the tests until the ready event has fired.

### DIFF
--- a/test/core/api.js
+++ b/test/core/api.js
@@ -8,64 +8,6 @@ describe("Core htmx API test", function(){
         clearWorkArea();
     });
 
-    it('onLoad is called... onLoad', function(done){
-        // also tests on/off
-        this.server.respondWith("GET", "/test", "<div id='d1' hx-get='/test'></div>")
-        var helper = htmx.onLoad(function (elt) {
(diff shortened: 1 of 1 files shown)


1cg:
oof

Issue #1679, opened 2023-08-08
Add more test cases for hx attributes when allowEval is false


jollytoad:
There appears to currently be a single test case for allowEval in test/core/api.js

I propose adding test cases for every attribute that can potentially evaluate JS.

* `hx-on`
* `hx-on:*`
* `hx-vals` (with `javascript:`/`js:` prefixed values only)
* `hx-vars`

The expected behaviour if `allowEval` is `false` is for the JS to **NOT** be evaluated AND a `htmx:evalDisallowedError` event to be triggered.

1cg:
no longer relevant

PR #1680, opened 2023-08-08
Added more tests for allowEval=false

jollytoad:
See #1679

This introduces tests for `hx-on`, `hx-on:*`, `hx-vals`, `hx-vars`, when `allowEval` is false.

`hx-on` & `hx-on:*` tests are currently failing as they do not conform to the expected behaviour, as defined in #1679

I'll follow up with a separate PR to fix these.


alexpetros:
Closing in favor of #1682

PR #1681, opened 2023-08-08
[fix] events between DOMContentLoaded and readyState.complete might be lost

xhaggi:
Relying on [readyState](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) `complete` in the `ready` function might cause some of the events registered via `htmx.on()` to be lost between `DOMContentLoaded` and `readyState === complete`.

Here is an example that simulates such a situation.
https://jsfiddle.net/k6pxg5wf/

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3603,16 +3603,28 @@ return (function () {
         //====================================================================
         // Initialization
         //====================================================================
-        var domContentLoaded = false;
+
+        var isReady = false;
         function ready(fn) {
-            if (domContentLoaded) {
+            if (isReady) {
                 fn();
             } else {
-                getDocument().addEventListener('DOMContentLoaded', function() {
-                    domContentLoaded = true;
-                    fn();
-                });
+                document.addEventListener('htmx:ready', fn);
             }
+        }
+
+        function loadComplete() {
+            document.removeEventListener('DOMContentLoaded', loadComplete);
+            window.removeEventListener('load', loadComplete);
+            isReady = true;
+            triggerEvent(document, 'htmx:ready');
+        }
+
+        if (document.readyState === 'complete') {
+            loadComplete();
+        } else {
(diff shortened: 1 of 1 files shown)


alexpetros:
If you're going to do this, what's the point of the conditional at all? Why not just

```js
function ready(fn) {
    getDocument().addEventListener('DOMContentLoaded', fn);
}
```

xhaggi:
The `DOMContentLoaded` event is only triggered once and after that the call to `ready` should immediately call the provided function. This is necessary because the implementation of `htmx.on/off` is wrapped in `ready`.

alexpetros:
Tagging for visibility @1cg 

The way you've done it, if the `DOMContentLoaded` event has already fired before `ready()` is called for the first time, it will always add an event listener for an event that will never fire, no?

xhaggi:
You are right, but how can this happen at all? Ready is always called during htmx initialisation.

xhaggi:
@alexpetros I added a second commit to address the concerns. The determination if ready is now separate from the `ready` function and also listens for `load` to not miss anything.

alexpetros:
To be honest, I think what you have here is too complex, and that we need to analyze the existing code a little better, not write more code around it. We just need to be more confident in our understanding of when the DOM event fires.

xhaggi:
What do you think I've been doing for the last few hours? So let me explain the relevant part for you.

* `document.readyState === 'complete'` is necessary if you inject `htmx.js` dynamically after the DOM and all resources are loaded.
* `document.addEventListener('DOMContentLoaded', loadComplete);` is used if `htmx.js` is used as normal script, deferred script or as JavaScript module. This is by far the most common case.
* `window.addEventListener('load', loadComplete);` is a fallback if `htmx.js` is executed between `DOMContentLoaded` and `load`.

And I decided to use the event-driven approach (`htmx:ready`) to queue all functions called via `ready`.

Issue #1390, opened 2023-04-18
hx-on ignores allowEval config


jollytoad:
In the 1.9.0 release `hx-on` scripts are executed even if `allowEval` is set to `false`.

jollytoad:
How about something like this, that lazily constructs the function the first time the event is triggered, and allows eval to be prevented if allowEval is false...

```js
var func = null;
var listener = elt.addEventListener(eventName, function (e) {
    return maybeEval(elt, function() {
        if (!func) {
            func = new Function("event", code + "; return;");
        }
        return func.call(elt, e)
    });
});
```

jollytoad:
it appears this is now fixed for hx-on, but not for hx-on:* ... i've prepared some additional tests for this, I'll raise a PR for later with a patch.

jollytoad:
See https://github.com/bigskysoftware/htmx/issues/1679 for new test cases.

jollytoad:
Neither hx-on or hx-on:* triggers `htmx:evalDisallowedError` event either.

PR #1682, opened 2023-08-08
Fix hx-on/hx-on:* to suppress evaluation if allowEval is false

jollytoad:
This builds upon PR #1680

It fixes the cases in #1390, and causes the new tests as defined in #1679 to pass.

Before this, `hx-on` handlers would not be added or evaluated if `allowEval` was false, but the restriction didn't apply to `hx-on:*` attributes which would still be called. Also neither attribute triggered `htmx:evalDisabledError`.

This fix registers handlers for `hx-on` & `hx-on:*` still but applies `maybeEval` at the point of execution, so that `htmx:evalDisabledError` is triggered when `allowEval` is `false` ... this is consistent with the behaviour elsewhere in htmx, and allows `allowEval` to be toggled live and event handlers will disabled/enabled accordingly.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1929,9 +1929,13 @@ return (function () {
         function addHxOnEventHandler(elt, eventName, code) {
             var nodeData = getInternalData(elt);
             nodeData.onHandlers = [];
+            var func;
             var listener = function (e) {
                 return maybeEval(elt, function() {
-                    new Function("event", code).call(elt, e);
+                    if (!func) {
+                        func = new Function("event", code);
+                    }
+                    func.call(elt, e);
                 });
             };
             elt.addEventListener(eventName, listener);
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1929,18 +1929,18 @@ return (function () {
         function addHxOnEventHandler(elt, eventName, code) {
             var nodeData = getInternalData(elt);
             nodeData.onHandlers = [];
-            var func = new Function("event", code + "; return;");
             var listener = function (e) {
-                return func.call(elt, e);
+                return maybeEval(elt, function() {
+                    new Function("event", code).call(elt, e);
+                });
             };
             elt.addEventListener(eventName, listener);
             nodeData.onHandlers.push({event:eventName, listener:listener});
-            return {nodeData:nodeData, code:code, func:func, listener:listener};
         }
 
         function processHxOn(elt) {
             var hxOnValue = getAttributeValue(elt, 'hx-on');
-            if (hxOnValue && htmx.config.allowEval) {
+            if (hxOnValue) {
                 var handlers = {}
                 var lines = hxOnValue.split("\n");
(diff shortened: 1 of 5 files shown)


alexpetros:
Yup, I know exactly what you're talking about. Really appreciate the extra tests too!

src/htmx.js
```diff
@@ -1929,18 +1929,18 @@ return (function () {
         function addHxOnEventHandler(elt, eventName, code) {
             var nodeData = getInternalData(elt);
             nodeData.onHandlers = [];
-            var func = new Function("event", code + "; return;");
             var listener = function (e) {
-                return func.call(elt, e);
+                return maybeEval(elt, function() {
+                    new Function("event", code).call(elt, e);
```
1cg, line 1934: seems like this would recompile every time the event is called, should maybe be left up above?

htmx doesn't dynamically reevaluate stuff for the most part, you have to explicitly call `htmx.process()` on a node again to update behavior

1cg (changes requested):
wdyt?

src/htmx.js
1cg: thank you for the tests, by the way, very well done!

jollytoad: Re: function eval, so I wasn't too sure about this, it seemed to me that the caching of the function constructor was a premature optimization that may become a source of surprise if you were to dynamically change the content of the attribute. ie. the behaviour would be inconsistent with regular html `on` event attrs. It also seemed to me to be a more consistent use of `maybeEval` to interpret the code at the point of call, which is how `maybeEval` is used elsewhere. 
Does `new Function` actually pre-compile anyway?
Does it actually have a measurable speed-up by caching the `new Function` construction?
Personally I'd avoid the caching of the function construction, unless there is a noticable performance issue, and caching noticably improves upon that.

jollytoad: Hmmm, just thinking about this, I guess the dynamic ability is a moot point for the hx-on attr anyway as it encodes the event name, and that can't be changed without the `htmx.process()` call to update.

Although, I'd expect `new Function` to never be called if `allowEval` was false, so maybe if there is an advantage to caching the `new Function` call, maybe do so on first trigger of the event.

I'm happy to adjust the PR if a consensus on approach is reached.

alexpetros: > if you were to dynamically change the content of the attribute

If you dynamically change the content of the attribute without calling `htmx.process` you're going to break a lot of stuff anyway. You're right that it's not consistent with html `on` events but it *is* consistent with the current behavior `hx-on`.

jollytoad: I've updated the PR to create and cache the function on first trigger.

alexpetros: I looked at your update, thought "there's no way that's the simplest way to accomplish that," tried to improve it, failed, and concluded that this is the simplest way to accomplish that.

alexpetros:
Tests pass 👍

PR #1687, opened 2023-08-09
Make htmx IE11 compatible again + tests IE11 compatible

Telroshan:
I know IE11 compatibility is almost a meme at this point, but as htmx 2 isn't that far ahead and with it the official end of IE11 compatibility, let's stick to the gimmick til the end!

# Core changes:
- Htmx is currently using mocha 10, that [dropped support for IE11](https://github.com/mochajs/mocha/discussions/4727) in this release
  -  Downgraded to mocha 9.2.2
- IE [doesn't support](https://caniuse.com/?search=string%20startswith) startswith / endswith on strings
  - Added startsWith/endsWith utility wrappers, using string.indexOf instead
- IE [doesn't support](https://caniuse.com/?search=case%20insensitive%20selector) case insensitive query selectors
  - Added a try catch to fallback to standard (case sensitive) selectors instead of raising on error
- IE [doesn't support](https://caniuse.com/?search=url) `URL`
  - Added a if statement to make a basic string comparison in that case, instead of raising an error
- The loading-states extension was using a lot of IE11 incompatible syntax such as arrow functions or template literals
  - Replaced by IE11 compatible syntax (functions and strings concatenation)
- The morphdom-swap extension was using fragment.firstElementChild, which IE [doesn't support](https://caniuse.com/?search=fragment%20firstelementchild)
  - Added a fallback to fragment.firstChild
- IE11 [doesn't support](https://caniuse.com/?search=xpath) XPath
  - Added a for-loop approach as a fallback when [document.evaluate](https://developer.mozilla.org/docs/Web/API/Document/evaluate) is undefined

I also updated a lot of tests to be IE11 compatible:
- IE may automatically remove empty class attribute + swap order between id and class, so we can't rely on a innerHTML equality check
  - Replaced with id / class / innerText equality checks
- IE doesn't support some features (svg titles, form attribute, submitter property in event, templates)
  - Added skips for IE11 in those tests as we simply can't expect them to ever work in IE11
- hyperscript isn't IE11 compatible
  - Added skips for IE11 in tests involving hyperscripts
- [mock-socket](https://github.com/thoov/mock-socket), used in the `ws-ext` tests, isn't IE11 compatible
  - Replaced it with a handmade socket mock, using the same syntax so the initial tests code didn't require changes (and could easily be replaced in the future with mock-socket again when we drop IE11 for htmx2)

# How to test on IE11
Let me know if it'd be relevant to add a section on the repo's README for contributors, about how to test in IE11, as it seems pretty common that new features / commits break IE11 compatibility
- In Edge's settings, make sure [`Allow sites to be reloaded in Internet Explorer`](https://support.microsoft.com/en-us/microsoft-edge/internet-explorer-mode-in-microsoft-edge-6604162f-e38a-48b2-acd2-682dbac6f0de) is enabled
- In Edge, open the settings tab and click [`Reload in Internet Explorer Mode`](https://support.microsoft.com/en-us/microsoft-edge/internet-explorer-mode-in-microsoft-edge-6604162f-e38a-48b2-acd2-682dbac6f0de)
- To open the dev console in IE mode, and see the logs, hit `Windows + R`, run [`%systemroot%\system32\f12\IEChooser.exe`](https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/ie-mode/#open-devtools-on-a-tab-in-ie-mode) and select the tab to inspect


Telroshan:
Thanks for your input! Just realized the old attribute is indeed mocked with a handwritten system, so I might as well just do that... I'll investigate and update this PR when I get the chance

Telroshan:
Should be good to go!
image

alexpetros:
@Telroshan does `npm run test` work for you locally? Because the tests are failing here.

Telroshan:
Oh right @alexpetros, didn't even know that command existed to be honest, I was running `npx serve` and accessing the test page with both IE11 and my regular Chrome
I don't get errors in browsers, but indeed, `npm run test` raises errors locally...
I have `4 failing` though, I wonder why the runner has 27 of them 😮 

Investigating!

Telroshan:
Well, turns out running `npm run test` locally, on the current `dev` branch (i.e without any of my changes here), give me the same result : 4 failing tests, on hx-boost, all tests involing a `a` tag (forms seem to be ok)

@alexpetros did this happen before?

alexpetros:
What I can say is that when I pull the current dev branch, use node 15, and run `npm t` I get:
```
  616 passing (12s)
  1 pending
```

I know the test suite in headless mode is mildly finicky—it has a couple timing issues, we're working on it—but it's reliable *enough* that  we can't really consider PRs that don't pass the CI.

alexpetros:
What's happening is definitely specific to your machine and very likely a timing problem with our tests; sucks that you have to be the one to figure that out, but that's just the limitation of the bandwidth that we have 😞

Telroshan:
No worries, I'll figure it out eventually!

Telroshan:
So I uninstalled node (noticed I had node 16 and not 15...)
- Installed node 15 (made sure no node nor npm was remaining)
- Cloned htmx into a fresh new folder
- Ran `npm i` then `npm run test`
- And somehow it... got even worse ?
```
572 passing (5s)
1 pending
44 failing
```
Kill me now

alexpetros:
This person had the same issue with the local test setup: https://github.com/bigskysoftware/htmx/pull/1650

Can you open a separate issue for the test stuff and we'll consolidate your findings there?

Telroshan:
Yay tests finally pass on the pipeline with your fix in #1742 @alexpetros

Issue #1690, opened 2023-08-10
Missing UMD wrapper and minification of extensions


xhaggi:
At the moment you need some workarounds to use an extension with Vite, Rollup or other build tools because of a missing UMD wrapper. Furthermore, the extensions are not minified like `htmx.js`.

**Workaround via window**

Create a local `htmx.js` with the following content:
```js
import htmx from 'htmx.org';
window.htmx = htmx;
```
Import this file before importing an extension.
```js
import './htmx.js';
import 'htmx.org/dist/ext/<extension>.js'
```

**Workaround via rollup-plugin-inject (Vite)**
```js
// vite.config.js
import inject from '@rollup/plugin-inject';
import { defineConfig } from 'vite';

export default defineConfig({
  plugins: [
    inject({
       htmx: 'htmx.org'
    }),
  ],
});
```

alexpetros:
I'm going to close this issue. If you're already using a bundler then it is not a huge hassle to configure your bundler to import htmx properly. I don't think this is worth polluting the `window` object in the based htmx for.

I'm open to re-opening it if there's a vocal constituency that feels differently.

alexpetros:
Re-opening per discussion in #1692 

> It doesn't feel right to have to use a workaround for importing the htmx extensions properly

Importing htmx extensions *into the Vite build system* which is not how we recommend installing htmx. I understand that some people do it that way. But in that case they are already using a build system, so the advantage of us having already UMD-ifyed the extension is basically nil, because the purpose of wrapping htmx core in a UMD wrapper is to allow people to import it natively in the browser, no matter which module system they're using.

So from my perspective, if you want to bundle htmx, it's not crazy to ask you to use a plugin that will bundle all that js for you.

matteocontrini:
> which is not how we recommend installing htmx

> it's not crazy to ask you to use a plugin that will bundle all that js for you.

I opened the original issue about this problem, and while I don't have a strong opinion on the solutions you're discussing, I don't think it's currently clear enough from the documentation that using npm and a build system is not recommended or has drawbacks.

I think it's a reasonable requirement to have users install that Vite plugin, but I also think that if that's the "solution" it should be made more clear in the installation instructions, maybe alongside the existing npm/webpack instructions.

alexpetros:
Yeah, I use htmx with npm just entirely to serve the `dist` file, I don't bundle it into anything. I'm open to a docs PR for what you described @matteocontrini.

[...]

PR #1692, opened 2023-08-11
Make htmx available in window scope to import extensions without UMD in modules

### LINKED ISSUE #1469: Using extensions with Vite

xhaggi:
This will fix #1469 and partially #1690 without introducing a UMD wrapper for each extension by simply making htmx available in window scope.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3678,6 +3678,9 @@ return (function () {
             }, 0);
         })
 
+        // make htmx available in window scope to import extensions without UMD in modules
+        window.htmx = htmx;
+
         return htmx;
     }
 )()


alexpetros:
Closing for reasons discussed in #1690. I don't think this is the responsibility of htmx core, which already has a lot of kludges to make it importable natively in the browser using a variety of mechanisms. If you're going to use a build step, you should take the responsibility for configuring that build step.

xhaggi:
It doesn't feel right to have to use a workaround for importing the htmx extensions properly. So why don't we fix the problem in such a simple way? Of course, we could introduce a UMD wrapper for all extensions, but this approach has already been rejected by you. What if we move this line of code to the UMD wrapper so that we don't pollute the core, i.e. to where it is necessary. What do you think?

alexpetros:
I'll re-open the issue so we can discuss it, and respond there.

PR #1694, opened 2023-08-13
Document hx-select-oob swap-strategy capabilities

BlakeWilliams:
👋 I've been playing with HTMX in a side project and so far it's been pretty great! I had run into a use-case for using `hx-select-oob` with `afterbegin` swapping behavior but didn't see it publicly documented and discovered the behavior after diving into the source since I had a feeling it would support it somehow.

I found that behavior [here](https://github.com/bigskysoftware/htmx/blob/47af93bdb0c9f5d356e36fd6073e0b74d12b9242/www/static/test/1.9.0/src/htmx.js#L803-L808) and wanted try documenting it so it's easy for folks to find in the future.


alexpetros (approved):
That rocks, thanks Blake :)

Issue #1660, opened 2023-08-02
Support use of hx-boost and hx-history-elt on the same page


HeyHugo:
A page with both a `hx-history-elt` element and `hx-boost` links will not work nicely in regard to browser back button.

Desired behaviour:
Clicking a hx-boost link and then back button will replace body and then restore previous body

Actual behaviour:
Clicking hx-boost replaces body like expected but what is stored in history cache is the contents of `hx-history-elt` instead of the body. When browser back is clicked:
A. If `hx-history-elt` is present on the page this will have it's content replaced by the cached content. (This might not be intended behaviour if the linked page's `hx-history-elt` is for something entirely different)
B. If `hx-history-elt` is not present on the page the body content will be replaced with the cached content

I've made a working POC patch that adds a flag `boosted` the each cache record to know wether to restore to body or `hx-history-elt` if present. If you like I can open a PR with this patch.

### Sidenote
I got a wild idea while thinking about this problem and looking at the `hx-history-elt` code. It would be nice if you could have "namespaced" `hx-history-elt` to potentially support multiple history partials. I think an API where instead there was a `hx-history-target` attribute would be nice. 
Example:
```html
<button hx-get="/test" hx-target="#some-container" hx-history-target="#some-container">
<div id="some-container">
  Content swapped 
</div>
```

The selector of `hx-history-target` would then be stored with each history cache record as `target` enabling htmx to know what element to target for the restore. For the hx-boost case `target` would be `"body"`.

Super breaking unfortunately, but something to consider for v2?

alexpetros:
Would love to see the PR for patch. Probably can't do the namespace history because it's a breaking change, unfortunately, at lease not for now.

PR #1695, opened 2023-08-14
Bugfix: hx-boost + hx-history-elt working on same page

HeyHugo:
Proposed fix for #1660

This fixes issues with browser back button when boosted links and hx-history-elt are on the same page. Whenever hx-boost adds a record to the history cache we store the property `boosted: true` along with the record. That way we can know if we should target the body when restoring from cache.


test/index.html
```diff
@@ -23,7 +23,7 @@ <h2>Scratch Page</h2>
 </ul>
 
 <h2>Manual Tests</h2>
-<a href="manual">Here</a>
+<a href="manual/">Here</a>
```
alexpetros, line 26: Why is this necessary?

alexpetros:
Don't use these features myself so sending this one up to @1cg

test/index.html
HeyHugo: The link to manual tests was broken on the `/test` page without trailing slash

I now noticed this is a problem within more sub-directories, e.g all the history subdirectories under `manual` shuld probably have trailing slash as well.

HeyHugo:
Added a commit to fix `undefined` being set as argument in `saveCurrentPageToHistory()` so we only ever set `boosted` to `true`/`false`

croxton:
History restores should inject cached content into the `hx-history-elt` element whether from a boosted link or not. That's the intended behaviour, not a bug. 

By default htmx targets `<body>` if `hx-history-elt` isn't defined,  but I usually use `<main>`. This PR would seem to force all history restores from boosted links to snapshot and restore into `<body>` which would break that implementation.

HeyHugo:
@croxton Hmm I see now what you mean, yes that would break things. I probably didn't use `hx-history-elt` like it was intended really. I used `hx-history-elt` on a modal type of element where back button had to close the modal. There were hx-boosted links to pages unrelated with the modal which had their back navigation broken.

`hx-history-elt` is a all pages or none kind of feature since it leaks into other pages by requiring them to also have `hx-history-elt` to not break hx-boost + back navigation.

This PR can be closed as I see it'd break backwards compatibility. I still think my idea in the sidenote of #1660 regarding `hx-history-target` would be really useful in cases like mine. Perhaps something for a future extension or htmx v2 :)

croxton:
@HeyHugo a pattern I often use it to put a container for a modal _outside_ of the `hx-history-elt` element and swap content into it and pushing the url so it creates a history entry. Then I ensure that the url loads as a standalone page showing the modal content. This works with the browser's forward/backwards buttons just like visiting a normal page. Here's a real example: https://level.law/team

In terms of managing areas of the content saved in the history cache, that is also possible using the `htmx:historyItemCreated` event if you have a specific requirement. This extension should give you some ideas:
https://gist.github.com/croxton/e2c33bd22591f9a5bd8c9d23a56c9edc

PR #1714, opened 2023-08-23
chore: simplify js functions

Zerotask:


src/htmx.js
```diff
@@ -273,12 +269,15 @@ return (function () {
             return responseNode;
         }
 
+        /**
+         * @param {string} resp
+         * @returns {boolean}
+         */
         function aFullPageResponse(resp) {
-            return resp.match(/<body/);
+            return resp.match(/<body>/) !== null;
```
prabodhmeshram, line 277: Are we intentionally changing the function implementation here from returning an Array to returning a Boolean?

src/htmx.js
```diff
@@ -289,27 +288,26 @@ return (function () {
...
-                }
+            }
+            var startTag = getStartTag(resp);
+            switch (startTag) {
+                case "thead":
+                case "tbody":
+                case "tfoot":
+                case "colgroup":
+                case "caption":
+                    return parseHTML("<table>" + resp + "</table>", 1);
```
prabodhmeshram, line 299: I see `resp` is a string, so can we just use template literals here and all the similar places afterwards?

```suggestion
                    return parseHTML(`<table> ${resp}</table>`, 1);
```

andryyy: I think they are keeping the IE11 compatibility for nostalgic reasons alive, and this would break it. :)

src/htmx.js
Zerotask: Yes. The result of the function is used as a boolean in a condition. This change makes its use more precise. Probably the name of the function could be adjusted too: `isFullPageResponse`

andryyy: For reference:

var partialResponse = !aFullPageResponse(resp);

src/htmx.js
Zerotask: I also prefer template literals. In that case I didn't change the original code, as this is subjective.
If it's wished, I can change it. There are probably many more candidates for template literals in the code.

src/htmx.js
prabodhmeshram: I agree, Renaming the function will clear the intent.

src/htmx.js
prabodhmeshram: Thanks for bringing it to my attention @andryyy , that seems like a valid point. I guess we can never get rid of the backward compatibility issue till IE(n) is alive 😀

PR #1716, opened 2023-08-24
log missing target element

boreddad:
Currently, htmx just logs "htmx:targetError" when the target element doesn't exist. It can sometimes be difficult to determine which element was missing when using `hx-target` if it fails

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2880,7 +2880,12 @@ return (function () {
             }
             var target = etc.targetOverride || getTarget(elt);
             if (target == null || target == DUMMY_ELT) {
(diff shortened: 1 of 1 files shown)


xhaggi:
It would make more sense to adapt the following line https://github.com/bigskysoftware/htmx/blob/93bd81b6d003bb7bc445f10192bdb8089fa3495d/src/htmx.js#L2116 so that `detail` is logged instead of `detail.error`, which is just the name of the event.

alexpetros:
Closing for the same reasons as this PR: https://github.com/bigskysoftware/htmx/pull/1518

PR #1718, opened 2023-08-24
Introducing the `classify` Extension

YidiSprei:
The `classify` extension aims to provide enhanced control over the CSS class properties of htmx-driven elements. Here's a brief overview of the changes and features introduced:

1. **Class Manipulation**: With the new `hx-classify` attribute, users can now specify class modifications such as `add`, `remove`, `replace`, and `toggle`.
2. **Reversion Mechanism**: The extension allows for the temporary application of class changes which can be reverted after a specified duration.
3. **Fallback Behavior**: Elements bearing the `hx-classify` attribute without a designated HTTP verb (hx-get, hx-post, etc.) will be assigned a `/noop-classify` endpoint. This ensures that the extension behaves predictably even without explicit HTTP actions. (such actions will be cancelled before a request goes out but the class change behavior will still take place)

##Testing:

I've tested the extension with various scenarios, ensuring consistent behavior across different class manipulation directives. Further testing by the community would be appreciated!

##Documentation:

The codebase includes comprehensive comments explaining the purpose and usage of the `classify` extension. Additional documentation updates might be needed to inform users of this new feature.

Conclusion:

The `classify` extension extends htmx's capabilities, giving developers fine-tuned control over the class attributes of their elements. I believe this addition will be beneficial for many users and use-cases. Feedback and further testing are welcome!


src/ext/classify.js
```diff
@@ -0,0 +1,103 @@
...
+ * with the `hx-classify` attribute but without a specified HTTP verb (like hx-get, hx-post, etc.)
+ * ensuring a fallback behavior.
+ * 
+ * Note: Ensure to use this extension judiciously to prevent unintended behaviors.
+ */
+
+htmx.defineExtension('classify', {
+    onEvent: function(name, evt) {
+        if (name === 'htmx:configRequest') {
+            const el = evt.detail.elt;
```
prabodhmeshram, line 32: Just a general note that the ES6 features are not supported in the library, as it is made compatible with IE11. Check other extensions defined in this library, everything is `var` based. Consider refactoring in other places where `let`, `const` is used

YidiSprei:
Fixed that

alexpetros:
This looks really great! For the time being, it might be best that you just publish this on your own as we don't really have the bandwidth to officially support it. Very excited that you're building with htmx though, and we might be able to showcase your work with a link on the website.

PR #1720, opened 2023-08-25
Syntax error when using CSS descendant combinators on hx-trigger modifiers

xhaggi:
If a CSS selector with [descendant combinators] (https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator) is used in `hx-trigger` for `from`, `target` and `root` modifiers, htmx logs an `htmx:syntax:error` and the selector is not parsed properly.

**Example**
```html
<div hx-get="/test" hx-trigger="input from:form input"></div>
```

This is an attempt to fix the problem by changing the token processing in these cases.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1126,6 +1126,7 @@ return (function () {
         }
 
         var WHITESPACE = /\s/;
+        var COMMA = /,/;
         var WHITESPACE_OR_COMMA = /[\s,]/;
         var SYMBOL_START = /[_$a-zA-Z]/;
         var SYMBOL_CONT = /[_$a-zA-Z0-9]/;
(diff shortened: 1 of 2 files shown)


alexpetros:
I think the reason you can't do descendant combinators in the `hx-trigger` syntax is because the whitespace in that syntax is already significant—it allows you to specify another option. In the example you posted, it's not at all clear what the second `input` is doing inside that attribute. 

I'm going to close this PR because I don't think we're going to pursue this option, but I'm sure that descendant combinator support is something that people would be interested in, so we would definitely welcome an issue for it and we can hash out a syntax there.

xhaggi:
Please create the issue yourselves.

PR #1721, opened 2023-08-25
Make htmx.ajax always return a Promise

beni69:
According to the [docs](https://htmx.org/api/#ajax), the `htmx.ajax` function returns a `Promise<void>`, but the typedefs say it's just `void`.

This was once fixed in #1265 but it seems to have been overriden in 15d4cac.

### DIFF
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -20,7 +20,7 @@ export function addClass(elt: Element, clazz: string, delay?: number): void;
  * @param path the URL path to make the AJAX
  * @param element the element to target (defaults to the **body**)
  */
-export function ajax(verb: string, path: string, element: Element): void;
+export function ajax(verb: string, path: string, element: Element): Promise<void>;
 
 /**
  * Issues an htmx-style AJAX request
@@ -31,7 +31,7 @@ export function ajax(verb: string, path: string, element: Element): void;
  * @param path the URL path to make the AJAX
  * @param selector a selector for the target
  */
-export function ajax(verb: string, path: string, selector: string): void;
+export function ajax(verb: string, path: string, selector: string): Promise<void>;
 
 /**
  * Issues an htmx-style AJAX request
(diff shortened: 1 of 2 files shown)


src/htmx.d.ts
```diff
@@ -20,7 +20,7 @@ export function addClass(elt: Element, clazz: string, delay?: number): void;
  * @param path the URL path to make the AJAX
  * @param element the element to target (defaults to the **body**)
  */
-export function ajax(verb: string, path: string, element: Element): void;
+export function ajax(verb: string, path: string, element: Element): Promise<void>;
```
prabodhmeshram, line 23: Should we also update the documentation to mention the expected return type?

beni69: The docs on the website linked above already state this return type. Do you mean adding a jsdoc `@returns` comment?

alexpetros:
Is it possible that the return type should be `Promise<> | undefined`? 

https://github.com/bigskysoftware/htmx/blob/101f255f5c054a0c204c6f59c98132a90ac4ceb7/src/htmx.js#L2922

beni69:
I'm not really familiar with the htmx codebase, but at face value, I feel like it would be more consistent to return a rejected promise instead of these early empty returns

src/htmx.d.ts
prabodhmeshram: Yeah, that's correct.

alexpetros:
If you would like to turn this into a fix for the `issueAjaxRequest` function that changes all the undefined returns into `Promise.resolve()` returns, I think that would be the best path forward.

beni69:
I agree, should I continue it in here or create a new PR?

alexpetros:
Go ahead and re-use this PR and we can just change the title

alexpetros (dismissed):
Great work, especially the choice of when to resolve or reject the promises.

1cg:
thank you!

PR #1726, opened 2023-08-26
docs: add unit to htmx file size info

Zerotask:


alexpetros:
The `~` means "approximate" so there's no need to be this specific, especially as that number will probably change.

If you're looking to contribute, check out some of the issues that we have open.

Zerotask:
The change was mainly because of the missing unit...

PR #1734, opened 2023-08-28
feat: add GoTH stack

tomanagle:
# Description

Adds the GoTH stack to the Go section of the server-side examples page


alexpetros:
I recognize that there's no consistent guideline on this, but I think to be added to this page it should either be

a) an example implementation
b) a framework with some demonstrated use

Really do appreciate you building with htmx!

PR #1737, opened 2023-08-28
Expand contribution guidelines to separate file

alexpetros:
## Description
Expanded our contribution guidelines to a `CONTRIBUTING.md` file, which is pretty standard. We're getting enough of the same kind of content that I want to give people more stable answers for why we do or don't merge PRs.

At some point I'll add issue and PR templates as well.

Feedback welcome, especially if you think anything is unfair, or poorly explained.


bencroker:
Looking great, thanks for this!

PR #1747, opened 2023-08-30
Include default config values in TypeScript definitions

metonym:
This PR proposes including default config values in the types. The desired DX improvement is to make it easier to reference a property's default value, as the default behavior for IDEs is to jump to the type definitions and not the source code.

---

VS Code hover


alexpetros (approved):
Looks good to me 👍

alexpetros:
Ah, that maybe should have been in `dev`... well we'll see if there's an issue

PR #1748, opened 2023-08-30
partial fix for sticky form submissions

jwhitaker-swiftnav:
(see issue #1749 )

I think this test reproduces the issue, but I'm having trouble with the test case: it seems to send the test runner into an endless loop and I can't figure out why.

I think the fix also is incomplete because it breaks another test. :(


alexpetros:
It looks like we got this fixed with #1749

PR #1762, opened 2023-09-04
[Doc] Fix missing beforeCleanupElement event in doc

Telroshan:
This adds a mention to the [htmx:beforeCleanupElement](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L955) event, which was missing from the docs


alexpetros (approved):
Thank you!

PR #1763, opened 2023-09-05
fix(sse): re-use existing event sources and get closest connect url

hesxenon:
I could still use some help for the tests - no idea what's available in scope or how to test the amount of connections without adapting the source code


hesxenon:
closing, since I apparently can't change the source anymore and it should come from dev as well

PR #1764, opened 2023-09-05
fix(sse): actually re-use event source and register swapped children

### LINKED ISSUE #916: SSE extension doesn't take into account children that are added after initialisation
The SSE extension add listeners to children with `sse-swap` tags when the EventSource is created. 

If, for any reason, a node is added later with its own `sse-swap`, the event will be ignored. Same if an existing node is replaced with another with same `sse-swap`.

As an example, let's say that I have a list of items and I want to change the one marked as "current" (via a class), so I use `hx-swap='outerHTML' for this:

```html
<ul hx-ext="sse" sse-connect="/items" hx-swap="outerHTML">
    <li sse-swap="Item1" class="current">1</li>
    <li sse-swap="Item2">2</li>
    <li sse-swap="Item3">3</li>
</ul>
```

If I send these events:
```
event: Item1
data: <li sse-swap="Item1">1</li>
event: Item2
data: <li sse-swap="Item2" class="current">2</li>
```

It will work the first time, ie the `<li>` will be replaced, but not the next times as the list of children is not updated, so the only event that will work, until sent once, will be `Item3`.

There is a lot of things that can impact the children, like a form adding a new element. Reloading the whole list is not a valid solution (the list can be long). Same for adding, in the example, a "container" node in the `<li>` (and removing the `hx-swap="outerHTML"`) because then I won't be able to change the classes or other attributes of the `<li>`.

hesxenon:
This PR introduces a dictionary that caches `EventSource`s by url so that newly processed nodes re-use existing connections instead of instantiating a new one.

closes #916


alexpetros:
Hi @hesxenon,

Unfortunately we don't have the resources to help with draft PRs. If you need some help, pop into the discord and there might be someone available to work with you. When you're ready to submit a PR, make sure it has a clear description of the problem, your solution to the problem, and (if relevant) tests that demonstrate your fixes.

hesxenon:
Hi @alexpetros 

I get that you don't have the resources to help with PRs, that's why I asked in discord when I opened this PR - the outcome was that someone tried to add tests but that didn't make it into main as well.

So with that in mind I would ask that you merge this BUGFIX without tests, since the underlying issue is still open and relevant and I don't see why I should expend my resources to propose, discuss and add a completely new test setup when the core contributors initially didn't.

Again, I get that you don't have the time to help with PRs that propose new features, but this is a BUGFIX for an ongoing issue that you also don't have the time to fix.

alexpetros:
You're right, I should have kept this open; I incorrectly read your description + the draft as saying that your PR was incomplete. In the future, describing what your PR does (beyond just a link to an issue) is very helpful for us, because it forefronts the work you did and are proposing we merge. I'll add that to the Contribution guidelines as well. 

Anyway, could please update this so that it doesn't incorrectly redindent that whole chunk of code, and we can see the actual change you made?

hesxenon:
@alexpetros I'm sorry for the confusion, yes I should have written a better description.

As for the dedenting - I don't think that makes sense, since the change wraps that whole chunk in an `if` statement which probably should be indentend?

The actual change is pretty simple conceptually:
1. create a dictionary of `sseUrl->EventSource`
2. check if an eventsource has already been created before creating a new one
3. if no eventsource has been created, do it now and store it in the dictionary.

So you can see, the second step is why a whole block appears to have been changed, because it's wrapped in the if statement. What actually has been changed is just the last line in this block and a bit of setup above.

alexpetros:
@Renerick do you have the bandwidth to take a look at this?

Renerick:
Yup. At a quick glance, I noticed that some lines in the diff are indented with spaces instead of tabs, so @hesxenon ,I would like to ask to fix that. I'll be able to do a deeper review this weekend.

hesxenon:
Honestly it's been a bit too long for me to remember all the details, it'll probably take some time for me to reconstruct the problem and the solution again.

I'll try to start from scratch this weekend, incorporating your feedback @Renerick.

hesxenon:
After looking at #2025 and taking into account that my [comment](https://github.com/bigskysoftware/htmx/issues/916#issuecomment-1705712626) already pointed out that the _easiest_ (not the best) solution would be what's implemented here I'm going to close this PR in favor of #2025.

I don't see how it would make sense for me to get into the details again if the "quick-and-dirty" fix isn't what you want (which is very understandable of course) and the proper solution is already being implemented by somebody who's probably much deeper/more recently involved in these issues than me.

Issue #1754, opened 2023-08-31
[New feature] Support inherit as list-keyword in hx-include


Telroshan:
Hey, I've run into a few cases where I defined an hx-include on some parent elements, then wanted to expand those includes on a child element.
Consider the following example:
```html
<input type="text" id="someInput" name="foo" value="bar">
<input type="text" id="anotherInput" name="name" value="whatever">

<div hx-include="#someInput">
  <form hx-post="/test">
    <input type="text" name="something" value="idk">
    <button type="submit">Submit</button>
  </form>

  <form hx-post="/test2" hx-include="inherit, #anotherInput">
    <input type="text" name="somethingElse" value="idk either">
    <button type="submit">Submit</button>
  </form>
</div>
```

Currently, `hx-include` supports the `inherit` keyword as other attributes, but only if it's the only keyword in the value.
Until now, in those cases I've been simply duplicating the parent `hx-include`, so in this case I would write
```html
<form hx-post="/test2" hx-include="#someInput, #anotherInput">
```
Though you can imagine it gets quickly annoying if you had declared multiple values on the parent `hx-include`. And even more if you want multiple layers of `hx-include` to inherit

Hence this proposal: I feel it'd make sense to support it in a comma-separated list selector, i.e. allow inherit along other specifiers, and allow syntax like
```html
<div hx-include="#something, #somethingElse">
  <div hx-include="inherit, input[name='whatever']">
    <form hx-include="inherit, .selector">
      ...
    </form>
  </div>
</div>
```

PR #1766, opened 2023-09-05
Inherit list-keyword in hx-include / hx-indicator

Telroshan:
This addresses #1754 
This change allows inheriting parent `hx-include` / `hx-indicator` values and expanding on them, such as:
```html
<div hx-include="#someInput">
  <div hx-include="inherit, .some-class">
      <button hx-include="inherit, #anotherThing" hx-post="/url"></button>
  </div>
</div>
```
Which would result in the button's request including `#someInput`, `.some-class` as well as `#anotherThing`


1cg:
>  Sep 5, 2023

thank u for your patience sir

PR #1768, opened 2023-09-06
fix problem with styles at the start of the server content

dmgd:
I was playing around with htmx with a site that has styles embedded in weird places, and noticed that when I navigated through history, occasionally styles were disappearing.

After a bit of investigation I found `makeFragment(..)` and it all made (a bit more) sense.  I've attached a screen recording showing the new tests before + after the fix. I recorded it with firefox but I checked it in chrome + safari too. Don't have IE to check that . . .

https://github.com/bigskysoftware/htmx/assets/183554/2e1367a4-f4ea-44a9-87a0-3a64212e313a

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -306,6 +306,7 @@ return (function () {
                     case "th":
                         return parseHTML("<table><tbody><tr>" + resp + "</tr></tbody></table>", 3);
                     case "script":
+                    case "style":
                         return parseHTML("<div>" + resp + "</div>", 1);
                     default:
                         return parseHTML(resp, 0);
--- a/test/manual/history_style/1.html
+++ b/test/manual/history_style/1.html
@@ -1,0 +1,16 @@
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <script type="application/javascript" src="../../../src/htmx.js"></script>
+    <title>History - 1</title>
+</head>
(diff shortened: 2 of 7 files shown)


Telroshan:
Ran a quick test in the browser's console, confirming the issue:
image

The problem is that passing a `<style>` element to the DOMParser will have it in the `head` of the generated document fragment, while the [`parseHTML`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L263) function retrieves its body, thus loses the style element in this case.
Wrapping it in a div makes it go in the body instead.

That's the same behaviour as for a `script` tag, which is already handled in the lib by this `div` wrapping, so seems like a good fix to me as it's reusing the same switch case

As for testing in IE @dmgd , you can run Edge in IE mode, see [this PR](https://github.com/bigskysoftware/htmx/pull/1687) for reference, in section `How to test on IE11`
For peace of mind, could you make sure it also works in IE mode ?

dmgd:
Thanks for the pointers on how to test against IE11. I've tested it with Edge in IE compatibility mode and it's working fine!

https://github.com/bigskysoftware/htmx/assets/183554/5ce5b37b-68be-41ee-8ae9-f8196ad2295c

Telroshan:
Perfect thanks!

PR #1776, opened 2023-09-08
Add support for array templates to the client-side templates extension

### LINKED ISSUE #1775: add array support in client-side-extensions (non-breaking)
I was recently bitten by #1115 (closed), and looking at the proposed fix #1155 (also closed) I realized that the issue could be addressed in non-breaking fashion by adding alternate template extensions with slightly different selectors. This will allow existing code to continue working, but allow folks (like myself) that need to make calls against APIs returning arrays to also function.

I have a PR ready to go for this (largely inspired by #1155 ), will submit that for consideration momentarily.

vadave:
Resolves #1775 by introducing `<template-engine>-array-template` renderers that can be used to work with APIs returning arrays of data.

### DIFF
--- a/src/ext/client-side-templates.js
+++ b/src/ext/client-side-templates.js
@@ -13,11 +13,30 @@ htmx.defineExtension('client-side-templates', {
             }
         }
 
+        var mustacheArrayTemplate = htmx.closest(elt, "[mustache-array-template]");
+        if (mustacheArrayTemplate) {
+            var data = JSON.parse(text);
+            var templateId = mustacheArrayTemplate.getAttribute('mustache-array-template');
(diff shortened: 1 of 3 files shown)


1cg:
beautiful, great solution!

biocarl:
Are you aware that at least moustache templates allow to reference the current context with `.`? So when passing in an array (e.g. [{"a":2, "b":4}, {"a":3, "b":4}] you can simply do:
```
{{#.}}  
    <h3>{{a}}</h3>
    <p>{{b}}</p>
 {{/.}}
```
Resulting in 
```
<h3>2</h3>
<p>4</p>
<h3>3</h3>
<p>4</p>
```

PR #1789, opened 2023-09-12
adds another htmx/go example

jritsema:


alexpetros (approved):
Looks great, thank you!

PR #1791, opened 2023-09-12
Update www.js - Adding a closing tag </li>

dcourtin:
Just saw the missing while reviewing the code.


alexpetros:
Fun fact about list elements, in [most circumstances](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li#technical_summary) you don't need the closing tag!

Issue #1198, opened 2023-01-12
htmx-oob-swap not replacing <tr> (table row) fragments


radcool:
Given the following starting web page:

````html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title>HTMX Test</title>
    <script
      src="https://unpkg.com/htmx.org@1.8.4"
      integrity="sha384-wg5Y/JwF7VxGk4zLsJEcAojRtlVp1FKKdGy1qN+OMtdq72WRvX/EdRdqg/LOhYeV"
      crossorigin="anonymous"
    ></script>
    <script>
      htmx.config.useTemplateFragments = true;
    </script>
  </head>
  <body>
    <div
      id="test"
      hx-get="/htmx-test2"
      hx-trigger="load delay:1s"
      hx-swap="outerHTML"
    >
      <p>Test BEFORE swap</p>
    </div>
    <hr />
    <h2>Contacts</h2>
    <table class="table">
      <thead>
        <tr>
          <th>Name</th>
          <th>Email</th>
          <th></th>
        </tr>
      </thead>
      <tbody id="contacts-table">
        <tr id="tr-test">
          <td>Cell 1a</td>
          <td>Cell 1b</td>
        </tr>
      </tbody>
    </table>
  </body>
</html>
````

with the following response returned from the `/htmx-test2` endpoint after 1s:

````html
<div id="test">
  <p>Test AFTER swap</p>
</div>

<tr hx-swap-oob="beforeend:#contacts-table">
  <td>Joe Smith</td>
  <td>joe@smith.com</td>
</tr>
````

The `<div>` with `id="test"` gets swapped, as expected. However, the OOB `<tr>` fragment does *not* get replaced. The console in the Chrome Dev Tools doesn't show the OOB swap firing. Using `afterbegin` instead of `beforeend` doesn't work either. Even this doesn't work:

````html
<div id="test">
  <p>Test AFTER swap</p>
</div>

<tr id="tr-test" hx-swap-oob="true">
  <td>Joe Smith</td>
  <td>joe@smith.com</td>
</tr>
````

In all three cases, the *whole* content from the response replaces the `<div>` with `id="test"`.

Note that `htmx.config.useTemplateFragments` is indeed set to `true`, as prescribed.

 This test is quasi-identical to the example given at https://htmx.org/examples/update-other-content/#oob.

radcool:
[...]

Issue #1538, opened 2023-07-04
hx-swap-oob adding a new table row only adds the <td>'s and strips the wrapping <tr>


roguefalcon:
When you attempt an oob-swap on appending a `<tr>` to a `<tbody>` the `<tr>` is stripped and only the `<th>` and `<td>` tags are actually added.  That makes the documentation example incorrect.  Specifically the [Updating Other Content example for Solution 2: Out of Band Responses](https://htmx.org/examples/update-other-content/) doesn't work.

I posted this issue on Stack Overflow [here](https://stackoverflow.com/questions/76612680/htmx-hx-swap-oob-adding-a-new-table-row-only-adds-the-tds-and-not-the-wrappin).  So you can see my full code and the corresponding answer there.

The solution is to put the hx-oob-swap on a `<tbody>` tag and wrap the `<tr>` tag.  I was told in the discord chat that this was a bug and the example in the docs should work as is.  Maybe update the docs until the bug is squashed?

kimmoeklund:
I run into same issue, and to isolate issue from rest of my code, I made repository with the "update other content" example implemented (https://github.com/kimmoeklund/htmx-examples-with-zio) using zio and scala3. After completing the example, I found this solution, but I guess someone might be interested in zio + scala3 implementation of the example. Maybe I add more of the examples to the repository as my htmx learning proceeds.

1cg:
fixed

PR #1794, opened 2023-09-14
Improve parsing of HTML response to support combining OOB and table elements

xhaggi:
After a talk on Discord with @yazalulloa about the issue #1538, I have been working on a fix to be able to use table elements like `<tbody>`, `<tr>` etc. together with OOB elements in a response.

```html
<div id="counter" hx-swap-oob="true">1 Item</div>
<tr>
  <td>Row 1</td>
</tr>
<tr>
  <td>Row 2</td>
</tr>
```

While there is a workaround via `htmx.config.useTemplateFragments` introduces in 1d52b23 (not IE11 compatible), this also leads to problems. If you do not place the OOB elements after the table elements, it will not work (see https://jsfiddle.net/2o957cpu/). So with this PR, the order of the elements should no longer matter.

Mind taking a look in the hope that this PR will not be closed again immediately 😄

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -236,20 +236,6 @@ return (function () {
         }
 
         /**
-         * @param {string} str
-         * @returns {string}
-         */
-        function getStartTag(str) {
-            var tagMatcher = /<([a-z][^\/\0>\x20\t\r\n\f]*)/i
-            var match = tagMatcher.exec( str );
-            if (match) {
-                return match[1].toLowerCase();
-            } else {
-                return "";
-            }
-        }
-
-        /**
          *
          * @param {string} resp
          * @param {number} depth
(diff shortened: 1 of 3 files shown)


src/htmx.js
```diff
@@ -273,43 +259,123 @@ return (function () {
...
          * @returns {Element}
          */
         function makeFragment(resp) {
-            var partialResponse = !aFullPageResponse(resp);
+            var partialResponse = !resp.match(/<body/);
             if (htmx.config.useTemplateFragments && partialResponse) {
                 var documentFragment = parseHTML("<body><template>" + resp + "</template></body>", 0);
                 // @ts-ignore type mismatch between DocumentFragment and Element.
                 // TODO: Are these close enough for htmx to use interchangeably?
                 return documentFragment.querySelector('template').content;
```
xhaggi, line 278: @1cg I'm not sure we really need this anymore.

Telroshan:
Nice work! I wonder if there could be some edge cases to add tests for? 
Maybe a crazy-nested one, or one using void elements as I see you have a special if for those? Just to make sure every part of the algorithm is tested by the pipeline to avoid potential future regressions
That regex is surely going to give me nightmares 😄

xhaggi:
Thx. Of course we can add more tests if you want.

>or one using void elements

Already there https://github.com/bigskysoftware/htmx/pull/1794/files#diff-9f65e208de1ea83b3075facd667931ccd7be292f6572a02df2bdff3eb1cf196eR25

xhaggi:
@Telroshan I have added more tests as requested by you. I also optimized the code a bit and was able to save a few lines. 

BTW the test failure in the CI pipeline are not related to my changes.

Telroshan:
Thanks! There are already a lot of PRs planned for the next update, I'll mark this as ready for review once the next update is released! 😄

blag:
I can confirm that this fixes issues #1198. It would be nice to see this merged before the next release.

ostcar:
Thank you for this PR. If fixes the issue.

But I found another bug, that only happens, when I use the version of this PR or when I merge this PR with the current master.

When I use htmx from this PR and the sse extension, and the sse-message contains `<a>` elements, then the sse-message is only swaped until the `<a>` element.

For example this sse-message:
```
data: hello world. <a href="https://github.com/bigskysoftware/htmx/pull/1794">Here</a> is a fantastic pr.
```

only gets rendert into the document as 
```
hello world. 
```

I creates a small example repo to demonstrate the bug: https://github.com/ostcar/htmx-sse-example

xhaggi:
@ostcar thanks for the detailed tests. I added a silly bug in the regex, but it is now fixed. I have also added your example response as a test case to be sure everything now works as expected. Could you please give it another try?

xhaggi:
@Telroshan I have made further optimizations to the current implementation. I removed the complex regex for custom elements, since we delegate the parsing to the browser anyway. I also fixed some issues with self-closing tags and text nodes, for which I also added additional tests.

ostcar:
> @ostcar thanks for the detailed tests. I added a silly bug in the regex, but it is now fixed. I have also added your example response as a test case to be sure everything now works as expected. Could you please give it another try?

It works now. Thank you very much.

1cg:
@xhaggi i really like this change, but it is pretty dramatic.  I think we are going to instead look into using `<template>` tags for OOB swaps, which should involve a lot less code than this and will handle sibling table elements gracefully.

PR #1811, opened 2023-09-19
Fix URL Params Appending on Boost

### LINKED ISSUE #1788: [bug] hx-boost: form with get method keeps appending param in the URL
# Summary
Given a form with method GET to implement a search with a param named `query`.

After submit the form, the param is appended to the URL. This is fine, but after each form submit the `query` param keeps getting appended in the URL resulting in multiple identical params appended.

## Version
Version 1.9.5 (not tested in previous versions)

## Reproducer

Given a simple form (search form) with method GET:

```html
<!DOCTYPE html>

<html lang="en">

<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
    <script src="https://unpkg.com/htmx.org@1.9.5"
        integrity="sha384-xcuj3WpfgjlKF+FXhSQFQ0ZNr39ln+hwjN3npfM9VBnUskLolQAcN80McRIVOPuO"
        crossorigin="anonymous"></script>
</head>

<body hx-boost="true">
    <form method="get">
        search <input type="text" name="query">
    </form>
</body>

</html>
```

_Without_ `hx-boost` URL is becomes `?query=<query1>`

_With_ `hx-boost`, the query keeps getting appended after every form submit resulting in: `?query=test&query=test ...`

## Expected

I would have expected the query not to be appended, it should honor the form action resulting in always one `query=<query>`

Rounak-stha:
This PR fixes #1788.
The change simply takes the URL part without any query params when AJAX request is issued on boosted element.
In case of anchor tags, the **href** value simply provides the requested URL but in case of forms, the location object is used to get the current href value.

Should this change include a test?
If yes then can somebody help me with that?

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3102,7 +3102,7 @@ return (function () {
 
             // behavior of anchors w/ empty href is to use the current URL
             if (path == null || path === "") {
-                path = getDocument().location.href;
+                path = getDocument().location.href.split('?')[0];
             }
 
 


src/htmx.js
```diff
@@ -3102,7 +3102,7 @@ return (function () {
 
             // behavior of anchors w/ empty href is to use the current URL
             if (path == null || path === "") {
-                path = getDocument().location.href;
+                path = getDocument().location.href.split('?')[0];
```
xhaggi, line 3105: As the above comment indicates, this is also applied to anchors with an empty `href` attribute and that changes would be a change in behavior. Not sure if we should fix it here or better move the handling to https://github.com/bigskysoftware/htmx/blob/7274454360aa5ff5f9a5ed1657450b93e9529b89/src/htmx.js#L1372

knpwrs:
I've just started looking into possibly using HTMX and noticed this behavior in my trials. It would be really great to get this or another fix merged!

alexpetros:
We can't merge this into the core right now because we are going to refactor this section of it.

This is also a big change—would love to see this as an extension first, and we'd be happy to link to it. It's not impossible this could be in the core eventually.

PR #1818, opened 2023-09-21
[Tests] Fix test pipeline timing issues

Telroshan:
- Most important change is calling mocha run on DOMContentLoaded instead of directly from the inline script (it could run mocha before htmx initialized itself, that caused the 84 errors scenario, probably the 24 errors one too)
- Changed a few hardcoded setTimeouts to rely on event listeners instead
- Increased the timeout to 4s (formerly 2) on the 4 tests making a request to hypermedia.systems ; those were causing random local errors where the (at least my) network connection would apparently not be enough to fetch the page in less than 2s

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -161,8 +161,10 @@
 <div id="mocha"></div>
 
 <script class="mocha-exec">
-    mocha.setup({globals: ['$0', '$1', '$2', '$3', '$4', 'performance', 'requestAnimationFrame', 'cancelAnimationFrame']}); <!-- IE11 -->
-    mocha.run();
+    document.addEventListener("DOMContentLoaded", function () {
+        mocha.setup({globals: ['$0', '$1', '$2', '$3', '$4', 'performance', 'requestAnimationFrame', 'cancelAnimationFrame']}); <!-- IE11 -->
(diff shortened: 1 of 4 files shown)


Telroshan:
Ran the pipeline a few times, passed tests everytime

test/index.html
```diff
@@ -161,8 +161,10 @@ <h2>Mocha Test Suite</h2>
 <div id="mocha"></div>
 
 <script class="mocha-exec">
-    mocha.setup({globals: ['$0', '$1', '$2', '$3', '$4', 'performance', 'requestAnimationFrame', 'cancelAnimationFrame']}); <!-- IE11 -->
-    mocha.run();
+    document.addEventListener("DOMContentLoaded", function () {
+        mocha.setup({globals: ['$0', '$1', '$2', '$3', '$4', 'performance', 'requestAnimationFrame', 'cancelAnimationFrame']}); <!-- IE11 -->
+        mocha.run();
+    })
```
alexpetros, line 167: This rocks

alexpetros:
I do think we should eventually remove the network requests entirely, I am somewhat anti having live network requests in your test suite.

Issue #1819, opened 2023-09-22
Button names not submitted with form after form is swapped


ekwoka:
Example:

```html
<body>
    <form
      id="hello"
      hx-target="#hello"
      hx-select="#hello"
      hx-swap="outerHTML"
      hx-get="/">
      <input type="text" name="name" />
      <button name="value" type="submit">Submit</button>
    </form>
    <button name="outside" form="hello" type="submit">Outside</button>
  </body>
  ```
  
  When submitting a form with a button external to the form, only the first use of the external button sends the buttons name with it. 
  
  Similarly, if anything else triggers htmx, it will similarly break (like a submit inside the form or some other event being listened on).

Telroshan:
Hey, I'm not sure I understand the issue
On [this JSFiddle](https://jsfiddle.net/gusr4wtq/), every time I click the outside button, its name correctly appears in the request payload. Am I misunderstanding something?

ekwoka:
Yes, that's missing the fact that the form isn't being swapped out.

It's just 404ing.

The issue is that the FORM is replaced (becomes a new element) but the outside BUTTON is the same as before. So it's reference to its FORM is scoped to the OLD form before the new one replaced it.

That JSfiddle never actually swaps out the form.

Telroshan:
Oh right, I had missed that hx-target value... thanks for the details, it's indeed a bug

BubbaJoe:
looks like this was fixes i guess, we can close this?

ekwoka:
Yes fixed with #1820

PR #1820, opened 2023-09-22
Fix stale form references

ekwoka:
Solves #1819 

Basically, the `form` reference was stale because it was stored outside the listener.

This also means that the form is more likely to be cleaned up since the closure no longer references the form element.

Tests seem broken/flaky. Change nothing and anywhere from 50 to 150 tests failed. The tests that seem like they'd be relevent to this change all passed, and the failures seem pretty unrelated.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1915,30 +1915,27 @@ return (function () {
         // Handle submit buttons/inputs that have the form attribute set
         // see https://developer.mozilla.org/docs/Web/HTML/Element/button
          function maybeSetLastButtonClicked(evt) {
-            var elt = closest(evt.target, "button, input[type='submit']");
-            if (!elt) {
-              return;
-            }
-            var form = resolveTarget('#' + getRawAttribute(elt, 'form')) || closest(elt, 'form');
-            if (!form) {
-                return;
-            }
-            if (elt !== null) {
-                var internalData = getInternalData(form);
-                internalData.lastButtonClicked = elt;
+            var internalData = getRelatedFormData(evt)
+            if (internalData) {
+              internalData.lastButtonClicked = elt;
             }
         };
         function maybeUnsetLastButtonClicked(evt){
-            var elt = closest(evt.target, "button, input[type='submit']");
-            if (!elt) {
-              return;
+            var internalData = getRelatedFormData(evt)
+            if (internalData) {
+              internalData.lastButtonClicked = null;
             }
-            var form = resolveTarget('#' + getRawAttribute(elt, 'form')) || closest(elt, 'form');
-            if (!form) {
-              return;
-            }
-            var internalData = getInternalData(form);
-            internalData.lastButtonClicked = null;
+        }
+        function getRelatedFormData(evt) {
+           var elt = closest(evt.target, "button, input[type='submit']");
+           if (!elt) {
+             return;
+           }
+           var form = resolveTarget('#' + getRawAttribute(elt, 'form')) || closest(elt, 'form');
+           if (!form) {
+             return;
+           }
+           return getInternalData(form);
         }
         function initButtonTracking(elt) {
             var form = resolveTarget('#' + getRawAttribute(elt, 'form')) || closest(elt, 'form');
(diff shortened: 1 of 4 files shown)


Telroshan:
Hey, could you please retarget your PR against the `dev` branch as per the [contribution guidelines](https://github.com/bigskysoftware/htmx/blob/master/CONTRIBUTING.md) ?

Telroshan:
Also, would you mind adding a test case for that specific scenario?
I'm not afraid of regressions as we already have a lot of tests about those buttons outside forms (and the test pipeline will tell), but we don't have any that tests this situation where the form gets replaced

ekwoka:
test added. It's one of the few that passes.

Telroshan:
See my review comments about the null checks, I would expect those to cause the tests issues
Tests are running fine on other PRs especially since the recent [timing issues fix](https://github.com/bigskysoftware/htmx/commit/bb5413dfd2fc5842eda7a29a1bdc53aa92e177c4) for the pipeline, so it seems unlikely to be unrelated to your changes

Telroshan:
Also please don't push code changes to the `www` directory, as it's the directory for the htmx website itself (to which tests & htmx are copied on new releases)

ekwoka:
Got it!

The feedback on the tests was just so unhelpful it was difficult to tell what was going on (and lots of difficulty actually running tests locally since the readme is out of date)

Telroshan:
> The feedback on the tests was just so unhelpful

Yeah unfortunately, every async test will throw that unhelpful `Uncaught Error: Script error. (:0)` error, I still haven't found a way to get the actual errors in the pipeline 😕 

That's where it helps to run tests locally as mentioned in [the README](https://github.com/bigskysoftware/htmx#hacking-guide), as your browser will have all the errors in the console

> since the readme is out of date

Could you elaborate on what you felt was missing from the README? Any feedback could help future contributors to find their way around!

ekwoka:
Actually running the tests in the terminal required 

```
npm run www
npm run test
```

Telroshan:
Weird, you normally shouldn't need to do that, as the [`test` npm command](https://github.com/bigskysoftware/htmx/blob/master/package.json#L28) runs tests in the `test` folder, while the [`www` npm command](https://github.com/bigskysoftware/htmx/blob/master/package.json#L31) populates the `www` directory (which you don't need to interact with unless you're adding some documentation)
What error were you getting?

Telroshan:
Btw, only the files in `www` left to remove and I'll be able to mark this PR as ready for review

1cg:
hey @ekwoka thanks for this fix.  Do you mind walking me through it so I make sure I understand the issue?

ekwoka:
@1cg Sure thing.

Basically, buttons are only evaluated at the start (either htmx initializing or the button being added by HTMX to the page).

This means that buttons outside forms are connected to the form they can submit at a specific point in time. But if that form then is removed/readded/replaced like by htmx swapping out the form element during a submission, the buttons handler will still only reference the old form data, even though it's actually submitting the new form.

The tests should demonstrate it quite clearly.

Since the buttons outside the form do a process to label themselves as "last clicked", this was being added to the wrong forms data. So the form would submit and not capture that button's information.

Hence this PR fixes stale form references, by not evaluating the linked form on the initial load, but instead only doing that once the button is actually clicked, making the form reference absolutely up to date.

ekwoka:
Any update on this?

1cg:
Reswapping this back into my brain now @ekwoka

src/htmx.js
```diff
@@ -1906,32 +1906,45 @@ return (function () {
             }
         }
 
+        // Handle submit buttons/inputs that have the form attribute set
+        // see https://developer.mozilla.org/docs/Web/HTML/Element/button
+         function maybeSetLastButtonClicked(evt) {
+            var elt = closest(evt.target, "button, input[type='submit']");
```
1cg, line 1912: feel like the first bit of these two functions could be refactored to use a common method, maybe `findRelatedForm(elt)` that returns nothing if there is none, then the logic can be boiled down to a check and set in both.

does that make sense?

1cg (changes requested):

1cg:
beauty!  thank you!

PR #1835, opened 2023-09-27
Update parseInterval to handle "0" correctly

tncowart:
When a parameter like "0ms" is passed in to parseInterval it gets parsed to 0. Previously this would result in a return value of "undefined" because 0 is falsy and thus the `return 0 || undefined` statements return undefined.

The purpose of the form `parseFloat(str) || undefined` was to return "undefined" if parseFloat failed (parseFloat returns NaN, a falsy value, if it can't parse its argument). Unfortunately, as mentioned, parseFloat can also succeed and return a falsy value -- when the argument is "0" (or "0.0", etc.). So the new code, rather than depending on the falsiness of the result of parseFloat, explicitly checks for a NaN.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1827,7 +1827,7 @@ return (function () {
                 if (!maybeFilterEvent(triggerSpec, elt, makeEvent("load", {elt: elt}))) {
                                 loadImmediately(elt, handler, nodeData, triggerSpec.delay);
                             }
-            } else if (triggerSpec.pollInterval !== undefined) {
+            } else if (triggerSpec.pollInterval > 0) {
                 nodeData.polling = true;
                 processPolling(elt, handler, triggerSpec);
             } else {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1484,14 +1484,14 @@ return (function () {
                             return;
                         }
 
-                        if (triggerSpec.throttle) {
+                        if (triggerSpec.throttle > 0) {
                             if (!elementData.throttle) {
                                 handler(elt, evt);
                                 elementData.throttle = setTimeout(function () {
                                     elementData.throttle = null;
                                 }, triggerSpec.throttle);
                             }
-                        } else if (triggerSpec.delay) {
+                        } else if (triggerSpec.delay > 0) {
                             elementData.delayed = setTimeout(function() { handler(elt, evt) }, triggerSpec.delay);
                         } else {
                             triggerEvent(elt, 'htmx:trigger')
@@ -1768,7 +1768,7 @@ return (function () {
                     handler(elt);
                 }
             }
-            if (delay) {
+            if (delay > 0) {
                 setTimeout(load, delay);
             } else {
                 load();
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -133,16 +133,18 @@ return (function () {
             if (str == undefined)  {
                 return undefined
             }
+
+            let interval = NaN;
             if (str.slice(-2) == "ms") {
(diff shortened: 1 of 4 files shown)


Telroshan:
Good catch! Falsy checks indeed become an issue when 0 is an acceptable value
To expand on this, while I don't know when you would need a 0 interval _(which would be so spammy)_, it is an accepable value as per the [setInterval doc](https://developer.mozilla.org/en-US/docs/Web/API/setInterval#syntax), thus should be handled properly as this PR suggest
> delay Optional
Defaults to 0 if not specified.

Telroshan:
Now that I think about it @tncowart 
I see at least [one other falsy check usage](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1828C24-L1828C45) in the code on the retrieved value from setInterval, that was not modified in this PR.

Could you add tests with 0 values to make sure this actually fixes the end behavior?
Also note that this applies to `every`, `throttle` and `delay` keywords, as well as `swapDelay` and `settleDelay`, so there are a few features impacted.

Internally, htmx uses `setTimeout`, which also [accepts the 0 value](https://developer.mozilla.org/docs/Web/API/setTimeout), and would benefit from this PR. Just gotta make sure it works as expected and fixing the `parseInterval` function alone resolves the issue (right now I think other falsy checks in the code will counter this)

tncowart:
@Telroshan Do I need to update tests in both `test` and `www/static/test` ? What is the difference?

Telroshan:
@tncowart The `www` directory is the htmx website's code, and we automatically copy the tests to it on new release so that the tests are runnable on the public website.
You may ignore them and should only update the files in the `test` directory
Hope this helps!

tncowart:
@Telroshan Are these sufficient? I also updated one of the falsy conditionals in htmx.js.

Telroshan:
Thanks a lot for the tests @tncowart!
Looking at the core, I think there are a few more places with falsy checks that should probably also be addressed _(linking to the corresponding line of code for each of the following)_
- [`if (triggerSpec.delay)`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1492)
- [`if (delay)`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1769) that is [called with `triggerSpec.delay`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1826)
- [`if (triggerSpec.throttle)`](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L1485)

Telroshan (commented):
In https://github.com/bigskysoftware/htmx/pull/1835/commits/a7c9bcdfb43c58945792b29156ad1f8fab312ad9

Why not use the `!== undefined` check here as you did for the `pollInterval`?
As per the [setTimeout doc](https://developer.mozilla.org/docs/Web/API/setTimeout), 
> delay
If this parameter is omitted, a value of 0 is used, meaning execute "immediately", or more accurately, the next event cycle.

So having a delay of 0 is not the same as calling the function right away, as it'll in fact execute in the next cycle
Let me know if I'm missing something there!

tncowart:
I think checking for x > 0 keeps the same behavior as before (0 and undefined are both false), whereas checking for x !== undefined subtly changes the behavior (only undefined is false). So maybe I should do > 0 everywhere, instead of !== undefined. Does that sound right -- change the !== undefined test to > 0?

Telroshan:
Oh you're right, it could be a breaking change as what was formerly called immediately would suddenly be delayed to the next cycle

I see that the default swap delay is set to 0 in the default config, so it would impact all existing apps using htmx, forget that!

PR #1838, opened 2023-09-27
Fix reference in notes from `hx-disable-elts` to `hx-disable-elt`

elliottinvent:
https://htmx.org/attributes/hx-disabled-elt/

I think this was originally `hx-disabled-elts` and all the references to it were then changed to `hx-disabled-elts` except the reference in the notes, in the footer.


Telroshan (approved):
Good catch, thanks!

PR #1842, opened 2023-09-28
fix unexpected behavior of hx-target on ancestors

### LINKED ISSUE #1841: Extended CSS selectors find targets relative to the issues, not the attribute
## The issue
Using [hx-target with extended CSS selectors](https://htmx.org/docs/#extended-css-selectors) are not selecting targets relative to the `hx-target` attribute, but instead relative to the "issuer" element (one with an `hx-get` for example). I find this being unintuitive and makes the extended selectors less useful.

### Example

```html
<div hx-target="next .target">
  <p class="target">Target</p>
  <button hx-get="/request">Make request</button>
</div>
```

#### Expectation
Clicking the button would update `Target` with the result of the request. As it would find the ancestor's `hx-target`, where the `next .target` is `Target`.

#### Reality
Clicking the button to make the request results in a `htmx:targetError`, because it uses the ancestor's hx-target selector `next .target` but does the search from the button—hence doesn't find one.

### Fix
Make the search relative to the where the used `hx-target` element. I can't think of a case where the current implementation is preferable.

### Code to easily reproduce
Here's a Python script to spin up a modified SimpleHTTPServer to serve the an example of the expected behavior more visually. Using bulma stylesheet to make it a bit more easier on the eyes–looking like this:
image

Save and run with `python3`.

```python
from http.server import SimpleHTTPRequestHandler, HTTPServer
import os
import secrets
from tempfile import NamedTemporaryFile
from urllib.request import urlopen
import weakref
import webbrowser

virtual_files = {}
def add_response(file_extension, contents, path=None):
    """convenience to fake making a file, served on a path"""
    path = path or f"/{secrets.token_urlsafe(8)}"
    with NamedTemporaryFile("wb", suffix=f".{file_extension}", delete=False) as f:
        f.write(contents if isinstance(contents, bytes) else contents.encode())
        virtual_files[path] = f
        weakref.finalize(f, os.unlink, f.name)
    return path


HTMX_URL = "https://raw.githubusercontent.com/bigskysoftware/htmx/ec126d2e3d973325f53fc69ed591ee908901e839/dist/htmx.js"

add_response("html", path="/", contents=f"""\
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8" />
<title>htmx bug reproduction</title>
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bulma@0.9.4/css/bulma.min.css">
<script src="{add_response("js", urlopen(HTMX_URL).read())}"></script>
</head>
<body class="section">
<div class="container block">

<!-- REPRODUCTION -->

<div class="box content">
    <p>Example from issue; see <code>htmx:targetError</code> in DevTools</p>
    <div hx-target="next .target">
      <p class="target">Target</p>
      <button hx-get="{add_response("html", 'Success')}">Make request</button>
    </div>
</div>

<hr/>

<div class="box content" hx-target="next .message-body" hx-swap="beforeend">
    <p>This box has <code>hx-target="next .message-body"</code></p>
    <hr/>
    <article class="message is-primary is-light">
        <div class="message-header">.message-body #1</div>
        <div class="message-body"></div>
    </article>
    <div class="columns block content">
        <div class="column">
            <p>This button doesn't have <code>hx-target</code>; will use box's <code>hx-target</code></p>
            <button class="button is-primary is-light"
                    hx-get="{add_response("html", '<p class="has-background-primary-light">Without hx-target result</p>')}"
            >Make request</button>
        </div>
        <div class="column">
            <p>This button has: <code>hx-target="next .message-body"</code></p>
            <button class="button is-info is-light"
                    hx-get="{add_response("html", '<p class="has-background-info-light">With hx-target result</p>')}"
                    hx-target="next .message-body"
            >Make request</button>
        </div>
    </div>
    <article class="message is-info is-light">
        <div class="message-header">.message-body #2</div>
        <div class="message-body"></div>
    </article>
</div>

<!-- END REPRODUCTION -->

</div>
</body>
</html>
""")


class RequestHandler(SimpleHTTPRequestHandler):
    def translate_path(self, path):
        file = virtual_files.get(path)
        if file is not None:
            return virtual_files[path].name
        return super().translate_path(path)


if __name__ == "__main__":
    server = HTTPServer(server_address=('localhost', 8000), RequestHandlerClass=RequestHandler)
    host, port = server.socket.getsockname()[:2]
    url = "http://{}:{}/".format(f'[{host}]' if ':' in host else host, port)
    print(f"Serving on: {url}")
    webbrowser.open(url)
    server.serve_forever()
```

tim-timman:
Fixes #1841

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -615,7 +615,7 @@ return (function () {
             var results = getDocument().querySelectorAll(match);
             for (var i = 0; i < results.length; i++) {
                 var elt = results[i];
-                if (elt.compareDocumentPosition(start) === Node.DOCUMENT_POSITION_PRECEDING) {
+                if (elt.compareDocumentPosition(start) & Node.DOCUMENT_POSITION_PRECEDING) {
                     return elt;
                 }
             }
@@ -625,7 +625,7 @@ return (function () {
             var results = getDocument().querySelectorAll(match);
             for (var i = results.length - 1; i >= 0; i--) {
                 var elt = results[i];
-                if (elt.compareDocumentPosition(start) === Node.DOCUMENT_POSITION_FOLLOWING) {
+                if (elt.compareDocumentPosition(start) & Node.DOCUMENT_POSITION_FOLLOWING) {
                     return elt;
                 }
             }
@@ -712,10 +712,11 @@ return (function () {
         function getTarget(elt) {
             var targetStr = getClosestAttributeValue(elt, "hx-target");
             if (targetStr) {
+                var thisHxTarget = findThisElement(elt,'hx-target');
                 if (targetStr === "this") {
-                    return findThisElement(elt,'hx-target');
+                    return thisHxTarget;
                 } else {
-                    return querySelectorExt(elt, targetStr)
+                    return querySelectorExt(thisHxTarget, targetStr)
                 }
             } else {
                 var data = getInternalData(elt);
--- a/test/attributes/hx-target.js
+++ b/test/attributes/hx-target.js
@@ -58,6 +58,26 @@ describe("hx-target attribute", function(){
         div1.innerHTML.should.equal("Clicked!");
     });
 
+    it('targets an ancestor defined `closest` element properly', function()
+    {
+        this.server.respondWith("GET", "/test", "Clicked!");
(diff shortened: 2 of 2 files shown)


alexpetros:
Closing this PR because it's a breaking change. I'll add some more comments on the original issue.

Issue #1852, opened 2023-10-02
feature: loading-states should support filtering based on HTTP method


jacob-ebey:
I would like to be able to support filtering based on HTTP method to avoid buttons flickering into loading states when another request to the same URL is in-flight with a different verb.

PR: https://github.com/bigskysoftware/htmx/pull/1853

Example:
```html
<div data-loading data-loading-method="post">Submitting</div>
<div data-loading data-loading-method="get">Loading</div>
```

jacob-ebey:
@alexpetros got another small quality of life improvement.

1cg:
extension issue, no longer relevant in 4.x

PR #1853, opened 2023-10-02
feat: support filtering state based on HTTP method

jacob-ebey:
Implements: https://github.com/bigskysoftware/htmx/issues/1852

Example:
```html
<div data-loading data-loading-method="post">Submitting</div>
<div data-loading data-loading-method="get">Loading</div>
```

### DIFF
--- a/src/ext/loading-states.js
+++ b/src/ext/loading-states.js
@@ -20,6 +20,15 @@
 		return pathElt.getAttribute('data-loading-path') === requestPath
 	}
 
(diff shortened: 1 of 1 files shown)


jacob-ebey:
Oops, didn't test this lolol. Will get to it eventually if no one else does.

alexpetros:
Yup, needs tests, documentation, and I'm not entirely sure about the name because the PR doesn't have any description so I don't really know what you're trying to do. I'll leave it open for a bit if you want to add those things.

PR #1858, opened 2023-10-04
misc :: docs:: Add form-group class for consistency.

nadim1992:
> Trust is built with consistency - **Lincoln Chafee**


alexpetros:
seems fine to me

PR #1860, opened 2023-10-05
Add dates to essays and sort chronologically

### LINKED ISSUE #1859: Add /essays/all page with chronologically sorted list of essays
It would be helpful to add dates to the list of essays on the website, and sort chronologically. This would make it easier to identify new essays that I haven't yet read.

nickchomey:
It seems like it would be easier to navigate (and keep up to date with) the essays if they were sorted chronologically and labelled with dates. I've sorted the first section and can do the rest if there's interest in merging this. 

Closes #1859

### DIFF
--- a/www/content/essays/_index.md
+++ b/www/content/essays/_index.md
@@ -5,17 +5,17 @@ page_template = "essay.html"
 +++
 
 ### Hypermedia and REST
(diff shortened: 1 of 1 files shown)


alexpetros:
Not interested in this particular solution, but I do think we should have a chrono-sorted list somewhere. I'll comment on the issue.

PR #1862, opened 2023-10-05
fix: prevent nodeData.onHandlers overwrite

raphael-p:
Related to this discussion I created: https://github.com/bigskysoftware/htmx/discussions/1861

I started digging into the code a little and realised that `nodeData.onHandlers` can only hold one on-handler at a time since it is wiped every time `addHxOnEventHandler` is called. The fix is to just not do that if `nodeData.onHandlers` already exists.

This is fixes the cause of the problem in the linked discussion because when there are multiple on-handlers on an element, only the last one is stored. So then only the last on-handler gets removed when the DOM updates. Every other on-handler just gets re-added without getting removed beforehand.

PS: I made this PR against the master branch, but please shout if it should be dev.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1950,7 +1950,9 @@ return (function () {
 
         function addHxOnEventHandler(elt, eventName, code) {
             var nodeData = getInternalData(elt);
-            nodeData.onHandlers ||= [];
+            if (!Array.isArray(nodeData.onHandlers)) {
+                nodeData.onHandlers = [];
+            }
             var func;
             var listener = function (e) {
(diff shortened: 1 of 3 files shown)


Telroshan:
Good catch!
You should indeed target the dev branch as per the [contributing guidelines](https://github.com/bigskysoftware/htmx/blob/master/CONTRIBUTING.md), could you please retarget your PR?

raphael-p:
closed in favour of https://github.com/bigskysoftware/htmx/pull/1894, which  targets the dev branch

PR #1863, opened 2023-10-05
Adds htmx/go example with stdlib approach.

rjxby:
Adds another htmx/go example but this time with golang stdlib approach in template handling.


alexpetros:
Generally speaking, we're looking for more full-features examples here, rather than skeletons. I know that criteria is sort of arbitrary though—at some point we'll make it a little clearer.

PR #1864, opened 2023-10-06
Escape OOB query selector

imankulov:
Escape OOB query selector to support special characters in element ids.

Refs #1537


imankulov:
Huh, I overlooked that it's been already addressed and rejected in https://github.com/bigskysoftware/htmx/pull/1602. Closing this down

PR #1867, opened 2023-10-06
Ignore formmethod when value is "dialog"

### LINKED ISSUE #1866: Ignore formmethod when value is "dialog"
Buttons support `dialog` as a valid value for the `formmethod` attribute.

[MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method)

Given the following example:

```html
<dialog>
    <form hx-post="/test">
        <button formmethod="dialog" name="foo" value="bar">Submit</button>
    </form>
</dialog>
```

htmx will issue a request with a method of `DIALOG`

There are a couple of issues with this:

- `DIALOG` is not a valid request method
- backend frameworks don't understand the request and the sent form data is ignored

Here is the line of code which causes this:

https://github.com/bigskysoftware/htmx/blob/04250d57f4fe15e7fab90e8114569853eb03fa5b/src/htmx.js#L2960-L2961

As `dialog` is a valid `formmethod` attribute, but not a valid request method, I think htmx should ignore the `formmethod` in this case when creating the AJAX request.

I'm going to create a PR with these changes and add some tests, but I'm more than happy to discuss further.

SamDudley:
In the following example htmx will issue a `DIALOG` request instead of a `POST` request.

```html
<dialog>
    <form hx-post="/test">
        <button formmethod="dialog" name="foo" value="bar">Submit</button>
    </form>
</dialog>
```

This PR changes the behaviour so that htmx will issue a `POST` request.

Closes https://github.com/bigskysoftware/htmx/issues/1859

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2959,7 +2959,10 @@ return (function () {
 
                 var buttonVerb = getRawAttribute(submitter, "formmethod")
                 if (buttonVerb != null) {
-                    verb = buttonVerb;
+                    // ignore buttons with formmethod="dialog"
+                    if (buttonVerb.toLowerCase() !== "dialog") {
+                        verb = buttonVerb;
+                    }
                 }
             }
 
--- a/test/core/ajax.js
+++ b/test/core/ajax.js
@@ -1230,4 +1230,32 @@ describe("Core htmx AJAX Tests", function(){
         this.server.respond();
         values.should.deep.equal({t1: 'textValue', b1: ['inputValue', 'buttonValue'], s1: "selectValue"});
     })
+
+    it('handles form post with button formmethod dialog properly', function () {
+        var values;
+        this.server.respondWith("POST", "/test", function (xhr) {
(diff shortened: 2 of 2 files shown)


alexpetros:
Heads up that I don't think you tagged the right issue in the description.

Otherwise, good catch, thanks for the PR.

pokonski:
Hey, @SamDudley . According to MDN and what I observed, such forms should never be submitted to the backend at all:

> When a <form> within a <dialog> is submitted via the dialog method, the dialog box closes, the states of the form controls are saved but not submitted,

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

SamDudley:
Hey @pokonski, I've left a [comment](https://github.com/bigskysoftware/htmx/pull/2753#issuecomment-2243683562) on #2753 discussing this issue. In short, I agree that my change is inconsistent with the MDN docs and support it being changed.

PR #1868, opened 2023-10-06
Sortable example fixes

mohamedhaddi:
### Main changes:

- Fix bug where, when an item is dragged fast enough before a previous request is completed, dropping it after the request's completion resulted in the addition of an extra item to the list. This is fixed by disabling the sortable on its `end` event, and re-enabling it on its `htmx:afterSwap` event.

    Before:
    
    https://github.com/bigskysoftware/htmx/assets/6943864/290ca6e8-8d17-42db-9fa7-1f58b75abffa
    
    After:
    
    https://github.com/bigskysoftware/htmx/assets/6943864/cafa81ec-8204-42ef-9a7d-af8e41835172
    
- Disable draggability of the `.htmx-indicator` ("Updating...") element.

    Before:
    
    https://github.com/bigskysoftware/htmx/assets/6943864/5f292592-5966-42ca-b7a4-312bac161e1a
    
    After:
    
    https://github.com/bigskysoftware/htmx/assets/6943864/e0eb73c4-4962-4bf8-b533-0d67587bdb46

### Other changes:

- Indicate draggable items by showing a grabbing cursor, and change the cursor on `.htmx-indicator` to `default` rather than `text`, since it's not supposed to be selectable especially while hidden.
- Add missing quote to the form's id.

### DIFF
--- a/www/content/examples/sortable.md
+++ b/www/content/examples/sortable.md
@@ -15,7 +15,11 @@ htmx.onLoad(function(content) {
       var sortable = sortables[i];
       new Sortable(sortable, {
           animation: 150,
-          ghostClass: 'blue-background-class'
+          ghostClass: 'blue-background-class',
+          filter: ".htmx-indicator",
+          onMove: function (evt) {
+            return evt.related.className.indexOf('htmx-indicator') === -1;
+          }
       });
     }
 })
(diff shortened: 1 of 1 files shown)


alexpetros:
Is this one ready for review?

mohamedhaddi:
> Is this one ready for review?

yes it is @alexpetros

alexpetros:
Can you add some inline comments in the code explaining what you explain here in the description. It's a good (and I think necessary) fix, but it does take away somewhat from the simplicity of the example so it would help to offset that with a little explanation.

mohamedhaddi:
@alexpetros Sure! How does it look?

alexpetros (approved):
Love it, thanks for the update!

PR #1876, opened 2023-10-09
Fix WS/SSE tests and replace go backend with NodeJS

alexpetros:
## Background
The live WebSocket/Server-Side Events tests were causing us a couple problems:

* Every release that we cut we'd get flooded with dependabot spam for the golang dependencies
* The golang dependencies also spam our GitHub security tab, where all but 4 of them are from go
* A couple of the tests were allegedly not working (I didn't actually verify this myself)
* You needed to install the golang toolchain to run them, which isn't a requirement anywhere else in the project

So, to the fix the dependabot thing and partially as an excuse to learn what WebSockets are, I reimplemented the tests' backend in NodeJS.

## Description
The main changes are as follows:
* I removed all the golang dependencies, at the cost of adding a single JS one (the excellent [ws](https://github.com/websockets/ws) library). This resolves the dependabot/security issue, and removes a full language toolchain from htmx.
* I re-implemented the routing mechanism to use hard links instead of `hx-boost` and `hx-get`, so now refreshing the page takes you back the tab you were on, instead of the homepage, and links are persisted in the URL bar/history. This makes developing on them much nicer.
* The tests use relative URLs so no CORS headers are required, and the syntax is a little easier to follow
* You can run the test suite from `package.json` now
* The code is actually a bit smaller line-wise

And as far as I can tell, all the tests work now! There's a couple small changes to the timing of the events that I'll play around with, but what I have now gets the job done.

## Testing
Run `npm run ws-tests` from the package root! There's also a readme in `/tests/ws-sse` directory with some granular instructions.

Note that the old go code is still mirrored on the website, which I left in for this PR so that the comparisons are more obvious. When we cut the next release it will go away and officially remove the the last of the go code.

## Credits
Huge props to @benpate for writing the initial version of these tests. It's much easier to adapt exist code than write a whole new paradigm, and his test suite and server code was really easy to follow.


Renerick (approved):
LGTM

benpate:
Hey, this looks really cool.  And Javascript will probably be a better / more portable environment to do this than Go has been. Awesome work :)

alexpetros:
That means a lot, thank you Ben :)

alexpetros:
@nickchomey That's great to know, these are good reads. I do think for the test purposes that I'll stick with `ws`, because it's popular and has no dependencies, so I feel pretty good about that it will run the tests more or less indefinitely.

PR #1903, opened 2023-10-18
Fix websocket polling trigger

### LINKED ISSUE #1896: Websocket Extension Errors when used with a polling trigger
The following example run with HTMX 1.9.6 and the webscoket extension fails

```
<div hx-ext="ws" ws-connect="/chatroom"  hx-trigger="every 60s">
    <form id="form" ws-send>
        <input name="chat_message">
    </form>
</div>
```

With the attached error
Screenshot from 2023-10-16 17-55-25

Tracking this down, it appears to be the [call to api.shouldCancel](https://github.com/bigskysoftware/htmx/blob/master/src/ext/ws.js#L382) in the websocket extension. The event is undefined when using a polling trigger so this call fails with the attached error.

I think this could be fixed by just checking the event is not undefined before calling shouldCancel

---

Correction: the version of HTMX this was seen in is 1.9.6, the file saved with the wrong version in the name which is why it appears as 1.6.9 in the stack trace.

Renerick:
Closes #1896

### DIFF
--- a/src/ext/ws.js
+++ b/src/ext/ws.js
@@ -379,7 +379,7 @@ This extension adds support for WebSockets to htmx.  See /www/extensions/ws.md f
 
 				socketWrapper.send(body, elt);
 
-				if (api.shouldCancel(evt, elt)) {
+				if (evt && api.shouldCancel(evt, elt)) {
 					evt.preventDefault();
 				}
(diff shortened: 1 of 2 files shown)


1cg:
my dude

PR #1906, opened 2023-10-19
Allow user to override Content-Type header

gbourant:
I'm using a 3rd party js library (openstreetmap) which returns me a `veryComplexObject` which i need to make an ajax request. At the moment htmx.ajax overwrites the `Content-Type`. With this commit we can do use any `Content-Type` we want.

```javascript
htmx.ajax('POST',"/map", {
        target:'#myTarget',
        swap:'morph:innerHTML',
        headers: {
            'Content-Type': 'application/json'
        },
        values: JSON.stringify(veryComplexObject)
})
````

This works great. But it would be cool if we had more control over of some (if not all) htmx properties like `hx-push-url`.

### CODE REVIEW

File: src/htmx.js [hunk 1]
```diff
@@ -3093,9 +3100,7 @@ return (function () {
             var allParameters = mergeObjects(rawParameters, expressionVars);
             var filteredParameters = filterValues(allParameters, elt);
 
-            if (verb !== 'get' && !usesFormData(elt)) {
-                headers['Content-Type'] = 'application/x-www-form-urlencoded';
-            }
```
alexpetros, line 3098: Correct me if I'm wrong, but I think we can just solve the issue by moving these lines to line 3080, with no other changes.

alexpetros (changes requested):
We also need a couple tests to validate the new and old behavior.

src/htmx.js
gbourant: Unfortunately it cannot be done, the headers variable is defined after the line 3080.

gbourant:
> We also need a couple tests to validate the new and old behavior.

I will add them within the next week

src/htmx.js
alexpetros: What about after 3082 then?

gbourant: I tried a few ways different variations of this but without any success. At my tries either the Content-Type was not set or when it was set other tests would fail.

Telroshan: Hey @gbourant, giving a look to this
I think @alexpetros is right, moving the lines he mentioned should be enough
- if `etc` defines a header property, it will override the initial one thanks to [the `mergeObjects` call](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3084), so you wouldn't need that userSetContentType check anymore as it would override anyway if it's ever defined
- Similarly, I don't think you would need [that userSetContentType check](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3278) when defining the params variable, as the value of `useUrlParams` is [based on htmx.config.methodsThatUseUrlParams](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3116)
If you wanted to override this, if you're going to define some custom headers in the `etc` through htmx.ajax or an event listener, then you may override the value of `useUrlParams` too since it also [retrieves the potentially modified value](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3145)

Let me know if I'm missing something there!

alexpetros:
Oh, also, needs to be retargeted against `dev` please

src/htmx.js
gbourant: > * if `etc` defines a header property, it will override the initial one thanks to [the `mergeObjects` call](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3084), so you wouldn't need that userSetContentType check anymore as it would override anyway if it's ever defined
> * Similarly, I don't think you would need [that userSetContentType check](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3278) when defining the params variable, as the value of `useUrlParams` is [based on htmx.config.methodsThatUseUrlParams](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3116)

Alright, for that part of the code i updated it so it won't use the `userSetContentType` variable and the tests are passing.

>   If you wanted to override this, if you're going to define some custom headers in the `etc` through htmx.ajax or an event listener, then you may override the value of `useUrlParams` too since it also [retrieves the potentially modified value](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3145)

Overriding `useUrlParams` alone doesn't solve the issue, does it? I mean we need some kind of way to determine if `param` variable is going to be initialized with etc.values.

What do you think?

Telroshan: I'm not sure to understand the last part, sorry if I'm off the mark
- `etc.values` is already included in `filteredParameters` as it's first [merged in `rawParameters`](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3095), which is then [used to construct `filteredParameters`](https://github.com/bigskysoftware/htmx/pull/1906/files#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR3099)
- On the other hand, I think using solely `etc.values` as you're doing here would ditch away any `hx-include`-d value for example

gbourant: Sorry i'm not familiar with htmx internals either. The point of this pull request is to be able to send JSON not just changing the `Content-Type` header. So the `userSetContentType` determines if we are going to send the raw etc.values property and not do any transformation (`encodeParamsForBody`). As a reference check the request below (or the added tests) 

```javascript
htmx.ajax('POST',"/map", {
        target:'#myTarget',
        swap:'morph:innerHTML',
        headers: {
            'Content-Type': 'application/json'
        },
        values: JSON.stringify(veryComplexObject)
})
```

So, how we can achieve that without the `userSetContentType` variable and without affecting `hx-include` value? Should we update the `encodeParamsForBody` method?

gbourant:
> Oh, also, needs to be retargeted against `dev` please

Yeap, i will change that after we find a solution.

src/htmx.js
Telroshan: Hey @gbourant, oh sorry I misread that ternary operator (was imagining the other way around) + I didn't realize the full context of your issue there.
### Current way of doing it
The current way to send JSON that I know of is using the `json-enc` extension
- use the [`json-enc` extension](https://htmx.org/extensions/json-enc/) that overrides the Content-Type but also the mime type of the XHR request (as htmx initially sets the latter to `text/html`)
- define the `hx-ext='json-enc'` attribute on your htmx ajax' [`source`](https://htmx.org/api/#ajax) element (which you don't define here, I see you're only using the `target`, whereas `source` is the element that initiates the request), and the extension will take care of formatting your JSON request
- for reference, here's [the extension code](https://github.com/bigskysoftware/htmx/blob/master/src/ext/json-enc.js) btw, it's a small one

As for
> But it would be cool if we had more control over of some (if not all) htmx properties like hx-push-url

That's something you can already accomplish also with the `source` element. Define the `hx-push-url` on it and it'll be applied!

### As for the PR
As the presented issue was the `Content-Type` override, only moving the lines mentioned above would fix it.

Now about the body encoding, the issue I see with the suggested system here is that overriding the content type to anything, even to `application/x-www-form-urlencoded` _(you could want to keep the same MIME type but [change the charset](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Type) or something)_, would here result in the params not being encoded, and I don't think you would want that in _every_ situation when overriding the Content-Type as your need seems to be JSON specific here right?

So the solve your JSON issue, I'd encourage you to look into `json-enc` first, though this PR to fix the order of `Content-Type` overriding is still valid as the code is currently incorrect in the sense that it ignores overrides supplied to htmx.jax (but the body encoding would be another topic imo)

gbourant: Alright, i was aware about `json-enc` but `htmx.ajax` was not working because i was not setting the source attribute. If this PR is good let me know so i will remove the `userSetContentType` part and then retarget against dev so you can merge it.

Although `json-enc` sets the `Content-Type` to `application/json` it tries to encode the values but does it wrong.

For example.

```java
    htmx.ajax('POST',"/mean", {
        source: '#right',
        target:'#left',
        swap:'morph:innerHTML',
        values: [
    {
        "lat": 37.02886944696474,
        "lng": 21.159667968750004
    },
    {
        "lat": 38.212288054388175,
        "lng": 21.159667968750004
    },
    {
        "lat": 38.212288054388175,
        "lng": 23.532714843750004
    },
    {
        "lat": 37.02886944696474,
        "lng": 23.532714843750004
    }
]
    })
```

is encoded to 

```json
{
   "0":{
      "lat":37.02886944696474,
      "lng":21.159667968750004
   },
   "1":{
      "lat":38.212288054388175,
      "lng":21.159667968750004
   },
   "2":{
      "lat":38.212288054388175,
      "lng":23.532714843750004
   },
   "3":{
      "lat":37.02886944696474,
      "lng":23.532714843750004
   }
}
```

It turns a json array to a json object. As i can see this not solvable with an extension unless we pass `etc` values to `encodeParamsForBody` method. If that's the case i will open a new issue.

Telroshan: Hmm yeah I guess that's due to how htmx merges `etc.values` with the other params.
It seems the doc doesn't make it explicit than an object is expected and not an array, a clarification would be nice there

The issue I imagine here is: how would htmx handle merging values relative to the source element (whether it's via hx-include or if the source element is a form), with an array like this?
If your array was a property of a root object then it would work, for ex
```json
{
    "someInput": "someValue",
    "yourArray": [
        {
            "lat": 37.02886944696474,
            "lng": 21.159667968750004
        },
        ...
    ]
}
```
But if the array is the root, how should htmx merge the values with it?
In your situation, do you have no other choice than sending an array as the root JSON? Or is it worth considering sending a root object with the array as a property of it?

Telroshan: As for the PR, to get your changes merged, I can only encourage you to split things up yep! I'd say keep this PR solely about the issue with Content-Type overriding, then feel free to submit another PR for the other things that were discussed here!

gbourant: Well i think i will open a new issue about the encoding so we can discuss there the questions you arise. In my situation i could do some workarounds but i think htmx supporting this by default would make the library more "complete".

As for the PR, check the changes and let me know.

Telroshan: Sounds & seems good to me!

www/content/api.md
```diff
@@ -58,7 +58,7 @@ or
     * `handler` - a callback that will handle the response HTML
     * `target` - the target to swap the response into
     * `swap` - how the response will be swapped in relative to the target
-    * `values` - values to submit with the request
+    * `values` - values to submit with the request (json object, not an array)
```
alexpetros, line 61: And this one as well

src/htmx.js
```diff
@@ -3301,7 +3302,7 @@ return (function () {
                 });
             });
             triggerEvent(elt, 'htmx:beforeSend', responseInfo);
-            var params = useUrlParams ? null : encodeParamsForBody(xhr, elt, filteredParameters)
+            var params = useUrlParams ? null : encodeParamsForBody(xhr, elt, filteredParameters);
```
alexpetros, line 3305: Please remove this change

alexpetros (commented):
Need to remove the two unrelated changes (@Telroshan is correct about splitting up unrelated PRs) and then we're good to go.

alexpetros:
Terrific, thank you!

1cg:
beauty

Issue #569, opened 2021-08-09
hx-trigger: "from:find" not working as documented


dzerrenner:
The following example code:
```html
<div hx-get="..." hx-trigger="click from:find .item">
  <div class=".item"></div>
</div>
```
throws an `htmx:syntax:error` at the browser console and does not trigger as expected on click on the inner item. I can't use the inner div to trigger, because it listens to its own event and has its own hx- attributes.

The docs or functionality seem off here: <https://htmx.org/attributes/hx-trigger/>, section "Standard event modifiers; from; find"

1cg:
@denysvega that's a great idea, are you interested in contributing a pull request for this?

denysvega:
@1cg sure!

sde-cdsp:
Seems like this whitespace issue is still an unresolved topic?
I was trying to do `hx-trigger="click from:#id input"` and was forced to do `hx-trigger="click from:#id>x>y>z>input" `

Encephala:
@sde-cdsp I believe that's actually a different issue. The original post wants to use an extended CSS selector, but the whitespace broke parsing of it. I believe that was fixed in 2021 by [this commit](https://github.com/bigskysoftware/htmx/commit/d6cd2e6fbe9305d0143e9f0bbb196a114e862b7c) (so @1cg I do believe this issue can be closed).

You are however correct that your use case isn't possible; any whitespace in a CSS selector besides the space after `find`, `closest` etc. will still cause the same parsing error.
I ran into this problem as well and implemented the fix suggested by Denys Vega above.

xhaggi:
I already made a fix in PR https://github.com/bigskysoftware/htmx/pull/1720, but that was rejected.

1cg:
fixed

PR #1913, opened 2023-10-21
Allow CSS selectors with whitespace in `hx-trigger`

Encephala:
A CSS selector that uses the descendant combinator, e.g. `form input` to target an `input` that is in a `form`, causes an HTMX syntax error, as the parsing of the `hx-trigger` consumes tokens until it finds whitespace, at which point it will try to parse `input` as a token and flag it as invalid syntax.

As suggested in #569, we could allow for such selectors when wrapping them in for instance curly brackets or parentheses. 

I implemented this logic, allowing for either parentheses or curly braces to wrap a selector. This works for both regular CSS selectors as well as extended CSS selectors using e.g. `from`.

I'd like to have allowed for square brackets or single/double quotes as well, but since those are both used already when parsing `hx-trigger`, this wasn't possible.

---
Examples:

```html
<div id="outer">
    <div class="first">Click me!</div>
    <div>Or click me</div>
</div>

<!-- Matches the second child div above -->
<div hx-get="/second-clicked" hx-trigger="click from:previous (#outer div)">Don't click me.</div>
<!-- Or, equivalently: -->
<div hx-get="/second-clicked" hx-trigger="click from:{#outer div:last-child}">Don't click me.</div>

<!-- Matches the first child div above -->
<div hx-get="/first-clicked" hx-trigger="click from:(#outer div.first)">Don't click me.</div>
```

---
Of course, feel free to say "no thanks" and close this PR 😊

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1144,6 +1144,8 @@ return (function () {
         var SYMBOL_CONT = /[_$a-zA-Z0-9]/;
         var STRINGISH_START = ['"', "'", "/"];
         var NOT_WHITESPACE = /[^\s]/;
+        var COMBINED_SELECTOR_START = /[{(]/;
+        var COMBINED_SELECTOR_END = /[})]/;
         function tokenizeString(str) {
             var tokens = [];
             var position = 0;
@@ -1232,6 +1234,18 @@ return (function () {
             return result;
         }
 
+        function consumeCSSSelector(tokens) {
+            var result = "";
+            if (tokens.length > 0 && tokens[0].match(COMBINED_SELECTOR_START)) {
+                tokens.shift();
+                result += consumeUntil(tokens, COMBINED_SELECTOR_END).trim();
+                tokens.shift();
+            } else {
+                result += consumeUntil(tokens, WHITESPACE_OR_COMMA);
+            }
+            return result;
+        }
+
         var INPUT_SELECTOR = 'input, textarea, select';
 
         /**
@@ -1280,16 +1294,17 @@ return (function () {
                                     triggerSpec.delay = parseInterval(consumeUntil(tokens, WHITESPACE_OR_COMMA));
                                 } else if (token === "from" && tokens[0] === ":") {
                                     tokens.shift();
-                                    var from_arg = consumeUntil(tokens, WHITESPACE_OR_COMMA);
-                                    if (from_arg === "closest" || from_arg === "find" || from_arg === "next" || from_arg === "previous") {
-                                        tokens.shift();
-                                        var selector = consumeUntil(
-                                            tokens,
-                                            WHITESPACE_OR_COMMA
-                                        )
-                                        // `next` and `previous` allow a selector-less syntax
-                                        if (selector.length > 0) {
-                                            from_arg += " " + selector;
+                                    if (tokens[0].match(COMBINED_SELECTOR_START)) {
+                                        var from_arg = consumeCSSSelector(tokens);
+                                    } else {
+                                        var from_arg = consumeUntil(tokens, WHITESPACE_OR_COMMA);
+                                        if (from_arg === "closest" || from_arg === "find" || from_arg === "next" || from_arg === "previous") {
+                                            tokens.shift();
+                                            var selector = consumeCSSSelector(tokens);
+                                            // `next` and `previous` allow a selector-less syntax
+                                            if (selector.length > 0) {
+                                                from_arg += " " + selector;
+                                            }
                                         }
                                     }
                                     triggerSpec.from = from_arg;
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1235,13 +1235,13 @@ return (function () {
         }
 
         function consumeCSSSelector(tokens) {
-            var result = "";
-            if (tokens.length > 0 && tokens[0].match(COMBINED_SELECTOR_START)) {
+            var result;
+            if (tokens.length > 0 && COMBINED_SELECTOR_START.test(tokens[0])) {
                 tokens.shift();
-                result += consumeUntil(tokens, COMBINED_SELECTOR_END).trim();
+                result = consumeUntil(tokens, COMBINED_SELECTOR_END).trim();
                 tokens.shift();
             } else {
-                result += consumeUntil(tokens, WHITESPACE_OR_COMMA);
+                result = consumeUntil(tokens, WHITESPACE_OR_COMMA);
             }
             return result;
         }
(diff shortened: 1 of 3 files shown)


Encephala:
(force pushed to fix a typo)

xhaggi:
FYI #1720 fixes the issue for all modifiers, not just `from`, without introducing a new syntax.

Encephala:
I now see Alex' comment there. I suppose we should create an issue to discuss it and leave this PR open/close it until further notice and reopen it if desired

alexpetros:
I think this one is reasonable, we'll look at it soon.

alexpetros:
Do you think you could add a couple examples to the description?

Encephala:
Sure! Do you mean just as a comment/edit to this PR, or in the docs?

And should the examples just illustrate the new syntax or also larger possible applications where this PR is useful?

alexpetros:
Just two examples as part of the PR description, so it's easily scannable. The docs are great :)

Encephala:
Done!

src/htmx.js
```diff
@@ -1232,6 +1234,18 @@ return (function () {
...
         }
 
+        function consumeCSSSelector(tokens) {
+            var result = "";
+            if (tokens.length > 0 && tokens[0].match(COMBINED_SELECTOR_START)) {
+                tokens.shift();
+                result += consumeUntil(tokens, COMBINED_SELECTOR_END).trim();
+                tokens.shift();
+            } else {
+                result += consumeUntil(tokens, WHITESPACE_OR_COMMA);
```
xhaggi, line 1244: Same here. And please do not set `result` to empty at the beginning of the function.

src/htmx.js
```diff
@@ -1280,16 +1294,17 @@ return (function () {
...
-                                    if (from_arg === "closest" || from_arg === "find" || from_arg === "next" || from_arg === "previous") {
-                                        tokens.shift();
-                                        var selector = consumeUntil(
-                                            tokens,
-                                            WHITESPACE_OR_COMMA
-                                        )
-                                        // `next` and `previous` allow a selector-less syntax
-                                        if (selector.length > 0) {
-                                            from_arg += " " + selector;
+                                    if (tokens[0].match(COMBINED_SELECTOR_START)) {
```
xhaggi, line 1297: Same here, please use `COMBINED_SELECTOR_START.test(token[0])`.

src/htmx.js
```diff
@@ -1232,6 +1234,18 @@ return (function () {
             return result;
         }
 
+        function consumeCSSSelector(tokens) {
+            var result = "";
+            if (tokens.length > 0 && tokens[0].match(COMBINED_SELECTOR_START)) {
+                tokens.shift();
+                result += consumeUntil(tokens, COMBINED_SELECTOR_END).trim();
```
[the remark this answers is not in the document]
Encephala: Did this because it felt more logical that the variable `result` always exists, as we're going to return it, figured you'd rather return an empty string if the code is wrong than error out.
Either way, I adjusted it now to be as you suggest.

Encephala:
Thanks for pointing out those modifiers @xhaggi. 
I expanded the functionality to those modifiers and added the other changes suggested.

All tests pass, but I don't quite get the purpose of the `root` and `target` modifiers so I'm not sure how to expand the tests for them (if it's even possible in Mocha).

Encephala:
Whoops I didn't mean to do that merge, let me know if you want the commit removed

xhaggi:
Yes please. Rebase onto dev.

Encephala:
Done.

src/htmx.js
```diff
@@ -1235,13 +1235,12 @@ return (function () {
...
-            if (tokens.length > 0 && tokens[0].match(COMBINED_SELECTOR_START)) {
+            if (tokens.length > 0 && COMBINED_SELECTOR_START.test(tokens[0])) {
                 tokens.shift();
-                result += consumeUntil(tokens, COMBINED_SELECTOR_END).trim();
+                var result = consumeUntil(tokens, COMBINED_SELECTOR_END).trim();
                 tokens.shift();
             } else {
-                result += consumeUntil(tokens, WHITESPACE_OR_COMMA);
+                var result = consumeUntil(tokens, WHITESPACE_OR_COMMA);
             }
```
xhaggi, line 1244: Mhh, that looks wrong to me. i meant:
```js
var result;
if (...) {
  result = consumeUntil(...)
} else {
  result = consumeUntil(...)
}
return result;
```

Encephala: Right, that makes more sense to me as well. Fixed it in the force push below.

src/htmx.js
```diff
@@ -1280,29 +1294,33 @@ return (function () {
...
-                                            tokens,
-                                            WHITESPACE_OR_COMMA
-                                        )
-                                        // `next` and `previous` allow a selector-less syntax
-                                        if (selector.length > 0) {
-                                            from_arg += " " + selector;
+                                    if (COMBINED_SELECTOR_START.test(tokens[0])) {
+                                        var from_arg = consumeCSSSelector(tokens);
+                                    } else {
+                                        var from_arg = consumeUntil(tokens, WHITESPACE_OR_COMMA);
```
alexpetros, line 1300: Am I missing something, or is this `if` construct redundant, since `consumeCSSSelector` already has a `if` construct where if checks for the start of the combined selector?

Encephala: I believe it is necessary, to differentiate between `from:previous (#a #b)` and `from:(#a #b)`; in the prior case, if we don't have this check and immediately jump to the block in the else statement, we'll first extract "previous" as a CSS selector, followed by a htmx syntax error being thrown due to the remaining `(#a #b)`

It'd be more readable if the condition of that `if` was "if one of the keywords (closest, find...) is there", but we only know that once we've grabbed then next token, but to grab the next token we first have to check if it is a descendant combinator selector, as those parse differently than other keywords.

Encephala: Scratch my previous comment, you're entirely correct.

Encephala: That test has to be there to make f.i. `hx-trigger="click from:(#a #b)"` work. If you remove the `if`-part and jump straight to the `else`-block, it'll only work if you actually use a keyword like `previous` etc..

Basically the code reads "if it's a descendant combinator, parse it, otherwise, check if it's either a simple selector or a keyword with a simple/descendant combinator and parse accordingly".

Your comment dit make me realize that the syntax of `from:<descendant combinator>` wasn't covered by the tests, so I improved them a little.

1cg:
lgtm

Issue #1920, opened 2023-10-23
Declare global and window types


romanzy313:
I am having a hard time declaring htmx types globally and on the window object. I load htmx via a script tag in the browser, but I would still like to have type-safety inside my typescript web-components.

Here are my attempts in `globals.d.ts`:

```ts
import * as Htmx from "htmx.org/dist/htmx.d.ts";
import Htmx from "htmx.org/dist/htmx.d.ts";
import Htmx from "htmx.org";

declare global {
  type htmx = Htmx;

  interface Window {
    htmx: Htmx;
  }
}
```

I think this has to do with the way .d.ts files are generated. Something to do with namespaces... But I have no idea how to make it work

i7N3:
Try this:
```
import * as Htmx from 'htmx.org'

declare global {
	interface Window {
		htmx: typeof Htmx
	}
}
```

MikeMoolenaar:
Thanks to @i7N3, I also got the global htmx working:
```ts
import * as Htmx from 'htmx.org';

declare global {
    var htmx: typeof Htmx;

    interface Window {
        htmx: typeof Htmx;
    }
}
```

1cg:
Four declares global types

PR #1926, opened 2023-10-28
Wrap type declarations in top-level interface

romanzy313:
Solves #1920

All I have done is wrapped all functions and constants defined on top level in an interface, changed formatting to make ts happy.

Htmx types declared as an interface will allow library consumer to do the following in `types.d.ts` file:
```ts
import type { HTMX } from 'htmx.org';

declare global {
  // globalThis definition
  var htmx: HTMX; 

  // window defintion
  interface Window {
    htmx: HTMX; 
  }
}
```
which is then included in tsconfig.json:
```json
"compilerOptions": {
    "types": [
        "./types.d.ts"
    ]
}
```

There is 2 ways the package can declare its exports, as default or as named export. Both implementations and import syntax can be seen below:
```ts
export default HTMX;    // import htmx from 'htmx.org'
export var htmx: HTMX;  // import { htmx } from 'htmx.org'
```

`npm run test-types` had 20 issues before and after the changes.

Also, I have noticed that many types are not correct, as they differ from the api reference. And `this` is not typed at all. I'd love to commit more PRs to remove this discrepancy.

### DIFF
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -1,295 +1,302 @@
 // https://htmx.org/reference/#api
 
-/**
- * This method adds a class to the given element.
- *
- * https://htmx.org/api/#addClass
- *
- * @param elt the element to add the class to
- * @param clazz the class to add
- * @param delay the delay (in milliseconds before class is added)
- */
-export function addClass(elt: Element, clazz: string, delay?: number): void;
+// 2 possibilities to declare exports from 'htmx.org' npm package, and how imports look like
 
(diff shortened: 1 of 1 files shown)


alexpetros:
related: https://github.com/bigskysoftware/htmx/pull/1936

alexpetros:
At the moment, we're not going to change the structure of this types declaration. We welcome additions to the existing structure, but we have a couple big changes to make before we re-evaluate that.

PR #1929, opened 2023-10-29
Add hx-target-error attribute to response-targets extension

phpfan72:
The hx-target-error attribute handles both 4xx and 5xx requests, allowing for simpler error handling.


alexpetros:
This is terrific, and I think is going to help us out a lot with a different blocker we're having.

PR #1934, opened 2023-10-30
Add missing config settings to type declarations in dist/htmx.d.ts

sjc5:
Add missing config settings to type declarations in dist/htmx.d.ts

PR #1936, opened 2023-10-31
Add missing config settings to type declarations in dist/htmx.d.ts

sjc5:
Add missing config settings to type declarations in dist/htmx.d.ts

### DIFF
--- a/dist/htmx.d.ts
+++ b/dist/htmx.d.ts
@@ -385,6 +385,16 @@ export interface HtmxConfig {
     disableSelector?: "[hx-disable], [data-hx-disable]" | string;
     /** @default "smooth" */
     scrollBehavior?: "smooth" | "auto";
+    /**
+     * If set to false, disables the interpretation of script tags.
+     * @default true
+     */
+    allowScriptTags?: boolean;
+    /**
+     * If set to true, disables htmx-based requests to non-origin hosts.
+     * @default false
+     */
+    selfRequestsOnly?: boolean;
 }
 
 /**


sjc5:
Follow-up to earlier PR, closed due to my git incompetence: https://github.com/bigskysoftware/htmx/pull/1934

1cg:
thank you!

sjc5:
Shoot, I didn't realize your actual source of truth for type declarations was in the `src` folder (I thought you all didn't use a build step?).

This was pulled against dist, not src. Sorry about that @alexpetros, @1cg 

Am I correct it should have been against this file instead?  

https://github.com/bigskysoftware/htmx/blob/dev/src/htmx.d.ts

Also, slightly off topic, are you open to a PR adding the various hx attributes as a new export in the `.d.ts` as well?

alexpetros:
Yes it should have been, that's my bad for not catching that. 

> Also, slightly off topic, are you open to a PR adding the various hx attributes as a new export in the .d.ts as well?

What's the point of typing HTML attributes, what does that have to do with typescript? (I'm sure there's an answer, I just don't know it)

sjc5:
For auto complete. So when you are say here, mid-typing:

```
<div hx-
```

You can hit ctrl-space or whatever your editor's key combo for autocomplete is, and see all the possible htmx attributes. It's great if your brain is old and forgetful like me.

And some take booleans or strings -- so you'd get a type error if you pass a boolean to an attribute that is expecting a string.

sjc5:
Someone in the discord was discussing a third-party package that exports these types, but the source of truth should really live here

Screenshot 2023-11-02 at 7 10 38 PM

alexpetros:
Yup, we're more than open to PRs that update these.

PR #1945, opened 2023-11-02
Update animations.md

bomberstudios:
Fix indent in example code


alexpetros:
It's fine the way it is.

PR #1949, opened 2023-11-03
Fix WS sendImmediately

### LINKED ISSUE #1930: sendImmediately not sending data to the server
I have the following implementation

```html
<div class="attempt d-flex" hx-ext="ws" ws-connect="/ws/attempt/{{ id }}/" hx-trigger="load">
  <div id="response-content"></div>

  <form ws-send>
    <input type="hidden" name="action" value="next_item">
    <button type="submit">Next</button>
  </form>
</div>
```

This is working fine and the `ws-send` works fine to send data to the server.

However, for a specific case, I have to send data to the backend directly from the script. For this, I have the following script
```javascript
let ws = null;

startTimer() {
    const interval = setInterval(() => {
        if (this.timeLeft === 0) {
            clearInterval(interval);
            if (ws) {
                console.log('websocket: ', ws);
                ws.sendImmediately(JSON.stringify({
                    action: 'timeout'
                }));
            }
        }
        if (this.timeLeft > 0) {
            this.timeLeft--;
        }
    }, 1000);
}

document.addEventListener('htmx:wsOpen', (e) => {
    ws = e.detail.socketWrapper;
});
```

the `ws` logs the methods available `send`, `sendImmediately`, etc., but the request is not going to the backend.

matiboy:
## Description

Fixes #1930. 

When no sending element is provided to sendImmediately, the message is not sent to the socket, which contradicts the documentation which states that sendingElt is optional.

This PR:
- Fixes the issue
- Adds tests for correct current behavior of calling `beforeSend` and `afterSend` on the correct element if provided
- Adds a test that would previously have failed (doesn't send if no sendElt provided)

To the reviewers, I've noticed that no second level "describe" is used to group the tests. In order to better focus on my own tests, it was easier for me to have a sub section of tests under ws extension. I've left it as is but please let me know if that's an issue.

### DIFF
--- a/test/ext/ws.js
+++ b/test/ext/ws.js
@@ -614,4 +614,168 @@ describe("web-sockets extension", function () {
         this.messages[1].should.contains('"foo":"bar"')
         this.messages[1].should.contains('"action":"B"')
     })
+
+    describe("Send immediately", function() {
+        function checkCallForWsBeforeSend(spy, wrapper, message, target) {
+            // Utility function to always check the same for htmx:wsBeforeSend caught by a spy
+            spy.calledOnce.should.be.true;
+            var call = spy.getCall(0);
+            call.args.length.should.equal(1);
+            var arg = call.args[0];
+            arg.target.should.equal(target);
+            arg.detail.socketWrapper.should.equal(wrapper);
+            arg.detail.message.should.equal(message);
+        }
+        it('triggers wsBeforeSend on body if provided to sendImmediately', function (done) {
+            var myEventCalled = sinon.spy();
+            var message = '{"foo":"bar"}';
+            var handler = function(e){
+                var socketWrapper = e.detail.socketWrapper;
+                window.document.body.addEventListener("htmx:wsBeforeSend", myEventCalled)
+                try {
+                    socketWrapper.sendImmediately(message, window.document.body)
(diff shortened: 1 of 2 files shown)


alexpetros:
Looks good, can you explain in more detail on the description what's bugged about the current behavior?

matiboy:
> Looks good, can you explain in more detail on the description what's bugged about the current behavior?

Issue #1930 : when no sending element is provided to `sendImmediately`, the message is not sent to the socket, which contradicts the documentation which states that sendingElt is optional.

test/ext/ws.js
```diff
@@ -749,6 +749,24 @@ describe("web-sockets extension", function () {
                 window.document.removeEventListener("htmx:wsOpen", handler)
             }
         })
+        it('sends message if no sending element is provided', function (done) {
```
matiboy, line 752: @alexpetros This test is the one which would currently fail; no sending element would result in the message being gobbled up.

alexpetros:
Added your comment to the description, hope that's alright. Just making it easier for 1cg to review.

1cg:
@Renerick can you review, looks like a great change but I want you to see it as well before I merge.

Renerick (approved):
That was an unfortunate oversight on my part, haha. The fix looks good and tests are excellent. LGTM

PR #1952, opened 2023-11-03
Replace keyup with input event

alexpetros:
Rebased version of #1181. Still has the original author in the commit.  

`keyup` only works with a keyboard, and ignores pasting with a right click or programmatically updating the value with JS. `input` is a better event for detecting any "inputs to the input".


alexpetros:
@Telroshan @Renerick can you one of you look at this and merge if it looks good?

Telroshan (approved):

PR #1956, opened 2023-11-04
Minor edits to article

bencroker:
Just a few minor edits and improvements to an otherwise great article.

### DIFF
--- a/www/content/essays/another-real-world-react-to-htmx-port.md
+++ b/www/content/essays/another-real-world-react-to-htmx-port.md
@@ -1,5 +1,5 @@
 +++
-title = "Another Real World React -> htmx Port"
+title = "Another Real World React → htmx Port"
 date = 2023-09-20
 updated = 2023-09-20
 [taxonomies]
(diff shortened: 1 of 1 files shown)


alexpetros:
I know this is well-intentioned but we can't open the door on non-typographical edits to essays. There's a million ways to write something, and if the authors of essays (by which I mean, at this time, the one guy who writes all the essays) want structural feedback they'll ask for it. Otherwise we'd have people PRing "hey this sentence would work better this way" all the time.

bencroker:
If you say so.

bencroker:
I used to frequently give @1cg feedback on essays “in the beginning”, in the hopes of improving quality and readability, but I guess things have changed. You’re also right that, in this case, the (unsolicited) changes I proposed are based on stylistic preference rather than grammatical fixes.

alexpetros:
For sure, and I know you're not just some rando, I'm just doing my best to triage the decisions we have the bandwidth to make. I just know I personally wouldn't want to be thinking about word choice on a 6 week old essay unless there was some egregious error.

PR #1967, opened 2023-11-06
Fix www.sh, copy scripts to theme's static/js folder too

Telroshan:
Kind of an emergency fix, see [this reported bug](https://github.com/bigskysoftware/htmx/pull/1610#issuecomment-1794835076)
- The website's htmx.js included in examples is currently outdated, it's using 1.9.5
- PR #1733 had replaced the old www js script by a new bash one, but it seems some copies [got lost in the way ](https://github.com/bigskysoftware/htmx/pull/1733/files#diff-5c1e18f5cb674fccbaffee74a972fbb1bfd76aec8a0a676367e4854445e3988fL31-L33)
- This PR adds those file copies again to the script

Note: I'm not familiar with Zola, but I wonder if we could get rid of those JS files at all and only use the ones in static instead? Wanted to avoid it for now as I'm not sure of the side effects


alexpetros:
Great stuff, thank you for fixing.

1cg:
Thanks @Telroshan really appreciate you having time to jump on this!

PR #1969, opened 2023-11-06
Experimental sqlite extension, request for comments/ideas

jyrimatti:
This is an experimental extension to use Htmx to query a SQLite database served over HTTP, instead of using XHRs. I'm not expecting this to be merged, but I decided to submit this as a pull-request for discussion and ideas and just visibility, since I think this is an extremely cool possibility :)

The idea is to use SQLite (WASM) running in the browser using the database file served via HTTP as range requests so that even large databases are usable. The functionality is implemented in https://github.com/mmomtchev/sqlite-wasm-http.

This extension tries to use it through Htmx preserving all relevant Htmx functionality. While Htmx itself sometimes removes the need for a complex frontend, a SQLite database could in some cases remove the need for a complex backend. Please see https://spot.lahteenmaki.net/test.html for some examples demonstrating basic usage and how Htmx still works as expected. Only backend for that "application" is a SQLite database served over HTTP (https://spot.lahteenmaki.net/spot.db).

I'm already using this extension in some hobby projects and I'm planning to keep using it. With this PR I'm hoping to get some additional insight and suggestions for improvement. If this functionality in any form will not be part of official Htmx, I'm going to publish it as an external extension, in which case I would still gladly welcome any ideas for improvement.

Known issues:
- elements need hx-boost to get "registered", I don't know if there's another way to do this? Would be cool if extensions could add additional VERBS so that hx-sql could be used identically to hx-get.
- since sqlite-wasm-http uses Web Workers and they need a _file_ to execute, the code cannot be bundled to a single js file. I think.
- since Web Workers don't work with cross-domain requests, the files cannot be used directly from a CDN, which is unfortunate for blog-post-examples but probably not an issue for a real project.
- Htmx doesn't currently export some functionality like swapping in its internal API, which forces extensions like this to copy-paste a lot of code. Maybe this should be improved?
- Htmx (including extensions API) assumes presence of XHR in several places, and since we don't have any XHR here, there may be some interoperability issues.

What do you think?

### DIFF
--- a/src/ext/sqlite.js
+++ b/src/ext/sqlite.js
@@ -1,6 +1,11 @@
 /*
 
-Extension to use SQLite database over HTTP as a backend for Htmx.
+Extension to use SQLite database backend for Htmx over:
+- HTTP as a range requests (used when hx-db starts with http:)
+- OPFS
+
+Specify databases with hx-db attributes.
+Specify queries with hx-sql attributes (will use closest hx-db).
 
 Requires sqlite-wasm-http bundled and served from same host where your application is,
 since Web Workers don't work with cross-domain requests.
@@ -42,6 +47,7 @@ And include in your HTML:
 
 (function(){
     var api;
+    var httpBackendConfig;
 
     // lot's of copying from htmx source. It should probably export some of this stuff as functions in internal API?
     var getParameters = function (elt) {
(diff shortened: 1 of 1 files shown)


jyrimatti:
> I wonder if this would be a better solution to your needs.

No, not really, since I wasn't interested in storing anything locally.

But you are correct that OPFS is highly related, so I modified the extension (as well as the test.html page) to support a local OPFS database.

alexpetros:
Hey, this looks really cool! We can't merge it because we can't officially support it, but if you publish it yourself and keep working on it, we can definitely promote it :)

jyrimatti:
Yeah, publishing this was a bit delayed since it has a couple of annoyances regarding copy-paste from Htmx source and impossibility to provide WASM stuff as a CDN bundle. But I just published it anyway, with some minor modifications: https://www.npmjs.com/package/htmx-sqlite

I will later have a look into how it's going to work with Htmx 2.

PR #1973, opened 2023-11-07
Homogenize configuration docs

Telroshan:
The configuration docs between the api reference and the docs reference are inconsistent, between each other but also with htmx.js:
- missing properties
- incorrect defaults (`disable-htmx` instead of `hx-disable`)
- 1 property was missing from htmx.config (`ignoreTitle`)
- different descriptions
- different order

This PR is an attempt to homogenize those 2 docs, consider it as a first step.
- ✅ missing properties
- ✅ incorrect defaults (`disable-htmx` instead of `hx-disable`)
- ~~✅ 1 property was missing from htmx.config (`ignoreTitle`)~~ => moved to a separate PR: #1981 
- 🚫 different descriptions
- 🚫 different order

Please note I didn't reorder the lines nor did I make all descriptions match.
I did so that it would be easier to see the diff and review the changes, let me know @1cg @alexpetros if I should have went all the way and reordered everything around

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -75,7 +75,6 @@ return (function () {
                 globalViewTransitions: false,
                 methodsThatUseUrlParams: ["get"],
                 selfRequestsOnly: false,
-                ignoreTitle: false,
                 scrollIntoViewOnBoost: true
             },
             parseInterval:parseInterval,
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -75,6 +75,7 @@ return (function () {
                 globalViewTransitions: false,
                 methodsThatUseUrlParams: ["get"],
                 selfRequestsOnly: false,
+                ignoreTitle: false,
                 scrollIntoViewOnBoost: true
             },
             parseInterval:parseInterval,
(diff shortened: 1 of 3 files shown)


src/htmx.js
```diff
@@ -75,6 +75,7 @@ return (function () {
                 globalViewTransitions: false,
                 methodsThatUseUrlParams: ["get"],
                 selfRequestsOnly: false,
+                ignoreTitle: false,
```
alexpetros, line 78: This is a source change, was that a mistake?

Telroshan: @alexpetros It's a configuration option that was mentioned in the doc, and used in the code, but not defined in the config object
> 1 property was missing from htmx.config (ignoreTitle)

This option was introduced by c230931 which also [added documentation](https://github.com/bigskysoftware/htmx/commit/c230931d42489a629d685bedb440fadd7eaf1416#diff-e772080787482bb78eeddd4d3c75a4ecd414c5e06674fde21d4383c1c3927735R1635) for it, but it seems it was forgotten from the htmx.config object iself
That documentation itself was saying:
> defaults to `false`, if set to `true` htmx will not update the title of the document when a `title` tag is found in new content

So I added that default value of false in the config

alexpetros: It's still gotta be in a separate PR. I know that seems picky but it's for a good reason. If you pull it out I can merge this one.

Telroshan: Oh ok let me split that then

alexpetros (approved):

PR #1980, opened 2023-11-08
Add a PR template

alexpetros:
Starting off pretty basic, just to address the most common complaints and make sure people write explanations.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,7 @@
 *Please describe what changes you made, and why you feel they are necessary. Make sure to include
 code examples, where applicable.*
 
-Corresponding issue (leave blank if none):
+Corresponding issue:
 
 ## Testing
 *Please explain how you tested this change manually, and, if applicable, what new tests you added. If
@@ -12,6 +12,7 @@ you're making a change to just the website, you can omit this section.*
 
 * [ ] I have read the contribution guidelines
 * [ ] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
-* [ ] If this is a source code change, I ran the test suite locally (`npm run test`) and verified
-  that it succeeded
-* [ ] If this is a new feature, I have implemented it as an extension
+  source changes)
+* [ ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
+  approved via an issue
+* [ ] I ran the test suite locally (`npm run test`) and verified that it succeeded
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,5 +12,6 @@ you're making a change to just the website, you can omit this section.*
 
 * [ ] I have read the contribution guidelines
 * [ ] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
-  source changes)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+## Description
+*Please describe what changes you made, and why you feel they are necessary. Make sure to include
+code examples, where applicable.*
+
+Corresponding issue:
+
+## Testing
+*Please explain how you tested this change manually, and, if applicable, what new tests you added. If
+you're making a change to just the website, you can omit this section.*
```
xhaggi, line 9: Is this section really necessary? What do you think people have to write here?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Checklist
+
+* [ ] I have read the contribution guidelines
+* [ ] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
+  source changes)
+* [ ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
+  approved via an issue
+* [ ] I ran the test suite locally (`npm run test`) and verified that it succeeded
```
xhaggi, line 18: I would change this point to:

>I added test cases that back my changes and run it locally (`npm run test`)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+
+Corresponding issue:
+
+## Testing
+*Please explain how you tested this change manually, and, if applicable, what new tests you added. If
+you're making a change to just the website, you can omit this section.*
+
+## Checklist
+
+* [ ] I have read the contribution guidelines
```
xhaggi, line 13: It would be nice if you link to the contribution guidelines.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+* [ ] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
+  source changes)
+* [ ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
```
xhaggi, line 16: When do I have to select this? Linking an issue should be enough, shouldn't it?

.github/PULL_REQUEST_TEMPLATE.md
alexpetros: Mostly people will interact with this in an edit view, where it will not display at as link.

.github/PULL_REQUEST_TEMPLATE.md
alexpetros: We want people to know what kind of PRs we're looking for, which is what's described here.

.github/PULL_REQUEST_TEMPLATE.md
alexpetros: There are situations in which new tests aren't required, so all I want to know is that they've run the test suite.

.github/PULL_REQUEST_TEMPLATE.md
alexpetros: I get your point, but I want people to describe their testing strategy, partially as a way to verify that they have one (and encourage them to develop one, if they don't already)

PR #1988, opened 2023-11-10
Get rid of htmx._ internal eval

xhaggi:
This removes the underscore function (`htmx._`) and thus the internal eval stuff from the public API and the codebase. This function is only used to access internal functions in tests, but this can also be achieved by exposing the internal API via a pseudo-extension (see `test/util/util.js`).

**Motivation**
[MDN eval()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval?retiredLocale=de#never_use_eval!):

>Warning: Executing JavaScript from a string is an enormous security risk. It is far too easy for a bad actor to run arbitrary code when you use eval(). See [Never use eval()!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval?retiredLocale=de#never_use_eval!), below.

and

[Vite](https://vitejs.dev/) build warning:

>Use of eval in "node_modules/htmx.org/dist/htmx.min.js" is strongly discouraged as it poses security risks and may cause issues with minification.

@1cg would you mind taking a look.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -78,7 +78,6 @@ return (function () {
                 scrollIntoViewOnBoost: true
             },
             parseInterval:parseInterval,
-            _:internalEval,
             createEventSource: function(url){
                 return new EventSource(url, {withCredentials:true})
             },
@@ -95,10 +94,13 @@ return (function () {
             addTriggerHandler: addTriggerHandler,
             bodyContains: bodyContains,
             canAccessLocalStorage: canAccessLocalStorage,
+            cleanInnerHtmlForHistory: cleanInnerHtmlForHistory,
+            encodeParamsForBody: encodeParamsForBody,
             findThisElement: findThisElement,
             filterValues: filterValues,
             hasAttribute: hasAttribute,
             getAttributeValue: getAttributeValue,
+            getCachedHistory: getCachedHistory,
             getClosestAttributeValue: getClosestAttributeValue,
             getClosestMatch: getClosestMatch,
             getExpressionVars: getExpressionVars,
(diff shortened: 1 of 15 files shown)


Telroshan:
That extension trick is neat!

In anticipation of potential relunctance to review huge PRs, I'm wondering if it would be easier to get your PR reviewed and merged if you were to keep existing calls in all test files, but make your `expose-internal-api` extension define `htmx._` itself? So that the current test code wouldn't need to change, and the htmx core would still be cleaned from that eval

Just an idea there, let me know what you think about it, please note that either way I'll mark this as ready for review as this looks good to me!

xhaggi:
I agree with you on all the points and even if we don't make htmx "safer" by doing so, we don't expose the `htmx._` eval method in the public API to use it for the tests. This would already be an improvement on what we have now.

BTW different reviewers different point of views. @Telroshan does not immediately reject these changes. But maybe he has less to say 😄  What about @1cg, do you have the same opinion.

alexpetros:
Genuinely appreciate that, and thanks for talking it through (that's why I wrote so much)

> we don't expose the htmx._ eval method in the public API to use it for the tests. This would already be an improvement on what we have now.

I don't object to users having access to the internals. If you want to mod your car and use a bunch of undocumented stuff, that's fine with me.

The more salient point is the security issue, and it sounds like we both agree that this doesn't affect that one way or another. I think it's a little obnoxious of Vite to run their code scanner on the libraries that people bundle with it—it's scary for arguably not good reasons—but I understand why they do it.

Issue #1131, opened 2022-11-11
User submitted content is not HTML escaped in the examples (XSS vulnerable)


cherouvim:
Steps:
1) Visit https://htmx.org/examples/click-to-edit/
2) Click `CLICK TO EDIT`.
3) Set `First Name` to `Joe<script>alert(1)</script>`.
4) Press enter.

The browser alerts "1" and this is XSS. Obviously this is not (easily) exploitable since there is no backend, nor a way to prepopulate the form via URL parameters, but I think that the examples should render user submitted content using HTML escaping. Otherwise you run the risk of having a percentage of the developers using HTMX replicating this sort of security issue on their websites.

cherouvim:
This is the case in https://htmx.org/examples/edit-row/ as well.

1cg:
would you mind putting together a pull request for this?

cherouvim:
@1cg I am not familiar with HTMX, nor do I wish to learn using it at this point. I just used the public example(s) and noticed the vulnerability so I thought that it would be good to report it.

PR #1995, opened 2023-11-13
Escape HTML on demo server

PabloNicolasDiaz:
I've fixed the www demo server to prevent returning unsanitized values wherever a request is made. When a 

It fixes https://github.com/bigskysoftware/htmx/issues/1131


alexpetros:
Sweet, thanks.

alexpetros (approved):

Issue #2023, opened 2023-11-21
`sse-swap`, `hx-target="sse:*"` not functioning when placed by htmx


fhp-mec:
Due to the way the sse extension identifies what is processed and what is not, if there is no `sse-connect` on the element being placed, it is not processed. 

When trying to insert this element
```html
<button sse-swap="remove-button" hx-swap="delete">click me</button>
```
this is what happens
```javascript
var sseURL = api.getAttributeValue(elt, "sse-connect"); // Looks for sse-connect

if (sseURL == undefined) { // if sse-connect isn't found
	var legacyURL = getLegacySSEURL(elt) // try again the old way
	if (legacyURL) {
		sseURL = legacyURL;
	} else { // still isnt found, as there isn't
		return null; // exit early
	}
}
```

Other mentions:
[(discord) Not getting sse-swap event updates on element placed with SSE - @eclipsenova](https://discord.com/channels/725789699527933952/1173478143173066752)
[(stackoverflow) HTMX does not re-process sse-swap attribute for Server-Sent-Events](https://stackoverflow.com/questions/76827939/htmx-does-not-re-process-sse-swap-attribute-for-server-sent-events)

fhp-mec:
I wrote a small test application that shows off the issue if used with an unpatched `sse.js`

https://gist.github.com/fhp-mec/5dc1980b8b77b062de3435c64b9734e9

fhp-mec:
This is a duplicate of #916

PR #2025, opened 2023-11-22
Fix: Make the SSE extension work properly and adhere to tests

fhp-mec:
## Description

- Add: tests for the sse extension
- Partially Fix: `sse:*` inside of hx-trigger <sup>[(1)](#note-1)</sup>
- Fix: `sse-swap` events inside of previously swapped content

Previously `sse-swap` were only processed when `sse-connect` was on the root of what was being swapped in.
This is now fixed so that the closest sseEventSource is found and used as the event source.

While fixing this issue I ported the existing sse tests (`attributes/hx-sse.js`) to the sse extension, and while running those realized that the sse extension was not compliant with those tests, the change was one line so I decided to add it and make this a general sse extension fix pull request.

In some ways this is similar to the #1764 which seems to be stalled. However, it fixes a few more bugs, adds the missing testing. 

Corresponding issues: #2023 #916 

<a name="note-1"></a>(1) I'm calling this a partial fix because actually supporting the full `hx-trigger` functionality for sse might not be possible without a substantial refactor of the trigger functionality in htmx itself. Right now the most basic usage of `hx-trigger` with sse events is supported (i.e. `hx-trigger="sse:<event name>"`) however, comma separation of events, etc are not supported and can't really be supported unless there htmx allows new keywords and behavior to be added to `hx-trigger` by extensions. Or I copy the entirety of the behavior of `hx-trigger` into the sse extension which seems deeply silly and a real pain for the future maintainability of the sse extension.

## Testing

I initially used the code in this [gist](https://gist.github.com/fhp-mec/5dc1980b8b77b062de3435c64b9734e9) to test this manually as there are no tests for the sse backend that I could find

I then realized that there were test cases in `attributes/hx-sse.js` so I copied those, edited them to work with the sse extension.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/sse.js
+++ b/src/ext/sse.js
@@ -5,311 +5,322 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 
 */
 
-(function(){
+(function() {
 
-	/** @type {import("../htmx").HtmxInternalApi} */
-	var api;
+    /** @type {import("../htmx").HtmxInternalApi} */
+    var api;
 
-	htmx.defineExtension("sse", {
+    htmx.defineExtension("sse", {
 
-		/**
-		 * Init saves the provided reference to the internal HTMX API.
-		 * 
-		 * @param {import("../htmx").HtmxInternalApi} api 
-		 * @returns void
-		 */
-		init: function(apiRef) {
-			// store a reference to the internal API.
-			api = apiRef;
+        /**
+         * Init saves the provided reference to the internal HTMX API.
+         * 
+         * @param {import("../htmx").HtmxInternalApi} api 
+         * @returns void
+         */
+        init: function(apiRef) {
+            // store a reference to the internal API.
+            api = apiRef;
 
-			// set a function in the public API for creating new EventSource objects
-			if (htmx.createEventSource == undefined) {
-				htmx.createEventSource = createEventSource;
-			}
-		},
+            // set a function in the public API for creating new EventSource objects
+            if (htmx.createEventSource == undefined) {
+                htmx.createEventSource = createEventSource;
+            }
+        },
 
-		/**
-		 * onEvent handles all events passed to this extension.
-		 * 
-		 * @param {string} name 
-		 * @param {Event} evt 
-		 * @returns void
-		 */
-		onEvent: function(name, evt) {
+        /**
+         * onEvent handles all events passed to this extension.
+         * 
+         * @param {string} name 
+         * @param {Event} evt 
+         * @returns void
+         */
+        onEvent: function(name, evt) {
 
-			switch (name) {
+            switch (name) {
 
-			// Try to remove remove an EventSource when elements are removed
-			case "htmx:beforeCleanupElement":
-				var internalData = api.getInternalData(evt.target)
-				if (internalData.sseEventSource) {
-					internalData.sseEventSource.close();
-				}
-				return;
+                // Try to remove remove an EventSource when elements are removed
+                case "htmx:beforeCleanupElement":
+                    var internalData = api.getInternalData(evt.target)
+                    if (internalData.sseEventSource) {
+                        internalData.sseEventSource.close();
+                    }
+                    var swapValue = api.getAttributeValue(evt.target, "sse-swap") || getLegacySSESwaps(evt.target)
+                    if (swapValue) {
 
-			// Try to create EventSources when elements are processed
-			case "htmx:afterProcessNode":
-				createEventSourceOnElement(evt.target);
-			}
-		}
-	});
+                    }
 
-	///////////////////////////////////////////////
-	// HELPER FUNCTIONS
-	///////////////////////////////////////////////
+                    if (api.hasAttribute(evt.target, "sse-swap")) {
+                        var sourceElement = api.getClosestMatch(elt, function(node) {
+                            return api.getInternalData(node).sseEventSource != null;
+                        })
+                        var eventName = api.getAttributeValue(evt.target, "sse-swap");
+                    }
+                    return;
+
+                // Try to create EventSources when elements are processed
+                case "htmx:afterProcessNode":
+                    createEventSourceOnElement(evt.target);
+            }
+        }
+    });
+
+    ///////////////////////////////////////////////
+    // HELPER FUNCTIONS
+    ///////////////////////////////////////////////
 
 
-	/**
-	 * createEventSource is the default method for creating new EventSource objects.
-	 * it is hoisted into htmx.config.createEventSource to be overridden by the user, if needed.
-	 * 
-	 * @param {string} url 
-	 * @returns EventSource
-	 */
-	 function createEventSource(url) {
-		return new EventSource(url, {withCredentials:true});
-	}
+    /**
+     * createEventSource is the default method for creating new EventSource objects.
+     * it is hoisted into htmx.config.createEventSource to be overridden by the user, if needed.
+     * 
+     * @param {string} url 
+     * @returns EventSource
+     */
+    function createEventSource(url) {
+        return new EventSource(url, { withCredentials: true });
+    }
 
-	function splitOnWhitespace(trigger) {
-		return trigger.trim().split(/\s+/);
-	}
+    function splitOnWhitespace(trigger) {
+        return trigger.trim().split(/\s+/);
+    }
 
-	function getLegacySSEURL(elt) {
-		var legacySSEValue = api.getAttributeValue(elt, "hx-sse");
-		if (legacySSEValue) {
-			var values = splitOnWhitespace(legacySSEValue);
-			for (var i = 0; i < values.length; i++) {
-				var value = values[i].split(/:(.+)/);
-				if (value[0] === "connect") {
-					return value[1];
-				}
-			}
-		}
-	}
+    function getLegacySSEURL(elt) {
+        var legacySSEValue = api.getAttributeValue(elt, "hx-sse");
+        if (legacySSEValue) {
+            var values = splitOnWhitespace(legacySSEValue);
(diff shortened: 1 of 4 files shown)


fhp-mec:
This should be ready to go now. In a rush I commited the test but not the fix for the test.

Test coverage is decent at this point

Renerick:
Hi! I was looking into related SSE PR and it appears that you got potentially proper fix for the important issue of `sse-swap` initialization, so I will be doing the review, as I think the issue is basically critical and should be fixed ASAP. At a quick glance, I noticed that your changes are indented with spaces instead of tabs, please fix that. 

Since it's a medium-to-large PR, I'll be properly focusing on the review during the following week. In the meantime, I provided some comments about current state of the extension and ideas for fixes in this comment https://github.com/bigskysoftware/htmx/pull/1764#issuecomment-1837099176. Please see if there is something useful here and consider incorporating those proposals in your PR.

Thank you for your contribution!

src/ext/sse.js
```diff
@@ -170,23 +178,28 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
...
-			for (var i = 0 ; i < sseEventNames.length ; i++) {
+			for (var i = 0; i < sseEventNames.length; i++) {
 				var sseEventName = sseEventNames[i].trim();
 				var listener = function(event) {
 
-					// If the parent is missing then close SSE and remove listener
-					if (maybeCloseSSESource(elt)) {
-						source.removeEventListener(sseEventName, listener);
+					// If the source is missing then close SSE
+					if (maybeCloseSSESource(sourceElement)) {
```
Renerick, line 186: The approach with `sse-swap` here is the correct one (using `queryAttributeOnThisOrChildren`), and it should be applied to `sse-connect` as well. Just make sure to perform a separate `getClosestMatch` lookup for SSE event source instead of reusing `sourceElement` in this line and the whole relevant block of code

Renerick (commented):
First of all, I'm sorry I took this long to leave my response and thank you for taking the time to work on improving SSE extension. Adding tests for the extension is HUGE improvement and will definitely help us with maintaining the extension. 

As for the fixes, I many your changes definitely positive, but unfortunately, I also found some cases when the extension still behaves not the way users would expect, which I detailed in the code comments. About, separating code into separate functions vs. keeping changes minimal, I very much respect your approach and will not insist on doing it exactly this way, but I also won't object to it, if you find such changes necessary. But i'm still convinced that having `sse-connect` search performed on the entire tree of swapped content is crucial for the extension and this is the first thing that should be fixed.

Best regards, and thank you once again for contributing to htmx!

test/ext/sse.js
```diff
@@ -0,0 +1,186 @@
...
+    function mockEventSource() {
+        var listeners = {};
+        var wasClosed = false;
+        var url;
+        var mockEventSource = {
+            removeEventListener: function(name) {
+                delete listeners[name];
+            },
+            addEventListener: function(message, l) {
+                listeners[message] = l;
```
Renerick, line 12: `addEventListener` should allow to register multiple event listeners for a single event, so `listeners` should store arrays, not just listener functions. I don't believe it has lead to any bugs yet, but better be safe

src/ext/sse.js
```diff
@@ -116,50 +117,57 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
...
-					createEventSourceOnElement(elt, Math.min(7, retryCount+1));
-				}, timeout);
-			}			
-		};
-
-		source.onopen = function (evt) {
-			api.triggerEvent(elt, "htmx:sseOpen", {source: source});
+			// Set internalData and source
+			internalData = api.getInternalData(sourceElement);
+			source = internalData.sseEventSource;
```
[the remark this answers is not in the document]
fhp-mec: I'm not sure I see where I'm storing a reference to the event source in an element not marked with `sse-connect`. I had thought I purposely avoided doing that. 

`elt` is the element being swapped, `sourceElement` is the element with `sseEventSource` on it retrieved and as far as I can tell the only place I set sseEventSource is in the block if postive case of this if statement.

I wrote a quick little test and it seems to agree with me. I'll push it up as a regression test with the rest of the fixes for this review
```javascript
    it('only adds sseEventSource to elements with sse-connect', function() {
        var div = make('<div hx-ext="sse" sse-connect="/event_stream" >\n' +
            '<div id="d1" sse-swap="e1"></div>\n' +
            '</div>');

        (byId('d1')["htmx-internal-data"].sseEventSource == undefined).should.be.true

        // Even when content is swapped in	
        this.eventSource.sendEvent("e1", '<div id="d2" sse-swap="e2"></div>');

        (byId('d2')["htmx-internal-data"].sseEventSource == undefined).should.be.true
    })
```

src/ext/sse.js
```diff
@@ -116,50 +117,57 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 			return null;
 		}
 
-		var internalData = api.getInternalData(elt);
-
 		// get URL from element's attribute
-		var sseURL = api.getAttributeValue(elt, "sse-connect");
+		var sseURL = api.getAttributeValue(elt, "sse-connect") || getLegacySSEURL(elt);
```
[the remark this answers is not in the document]
fhp-mec: This appears to be an issue with the legacy implementation as well... I'll fix it for both in the extension but it will create an inconsistency in the behavior of legacy swapping handled by the extension and handled by htmx itself. 

I'm willing to fix it in htmx itself, but that seems like it significantly increases the scope of this pr beyond the sse extension. Let me know if you feel I should, or if I should make legacy handling in the extension mirror the existing legacy handling.

test/ext/sse.js
fhp-mec: Should I change this in the legacy sse tests I filched this from?

fhp-mec:
> As for the fixes, I many your changes definitely positive, but unfortunately, I also found some cases when the extension still behaves not the way users would expect, which I detailed in the code comments. 

I think I've resolved almost all of your comments that I didn't have questions about in my latest commit.

> About, separating code into separate functions vs. keeping changes minimal, I very much respect your approach and will not insist on doing it exactly this way, but I also won't object to it, if you find such changes necessary. 

I've now separated the `eventSource` creation logic from the event registering logic. On reflection now that the scope has spiraled wildly, its better to do the right thing then to try to limit scope creep :)

> Best regards, and thank you once again for contributing to htmx!

Thanks for your diligence and your work on this fantastic library!

I look forward to your further comments.

src/ext/sse.js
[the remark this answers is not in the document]
Renerick: Whoops, that was my bad. I may have played arround with your code and introduced a bug myself and then erroneously attributed it to you. I apologize, please disregard this comment

test/ext/sse.js
Renerick: If not tests break, it would be great. If tests break, please report it and we shall fix them or report them separately

src/ext/sse.js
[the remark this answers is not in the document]
Renerick: We already have such inconsistency with legacy WS anyway, so I say, you may leave legacy SSE as it is

src/ext/sse.js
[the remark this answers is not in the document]
fhp-mec: No problem. When I split the event source and event listener creation logic as you suggested it became a lot easier to follow... Before it was not...

test/ext/sse.js
fhp-mec: Should be resolved by my latest commit

Renerick (approved):
Ok, I've reviewed the latest changes, and all seems to be in order. My only nitpick would be to rename `registerSSE` into something more explicit, like `registerSSEEventHandlers` for example. There is also a lot of some whitespace changes which aren't strictly necessary, but I'm fine with it. Would be great to also get some feedback from @benpate, as for me, I approve this PR

1cg:
@fhp-mec can you confirm that this change is backwards compatible?  (I don't know this code well enough to be sure)

1cg:
merged per @Renerick's endorsement, @fhp-mec would still like your comments on backwared compat.

thank you for your work on this!

fhp-mec:
@1cg

I'm relatively confident in saying its backwards compatible with 1.9.6, I dropped it into a thing that I had deployed a little while ago and it just worked. 
That said it only used the swapping functionality, though I think it may have pretty much covered everything that was working in 1.9.6.

I'm not super familiar with the javascript support on ie11 so I tried to copy the syntax used in the rest of htmx and use as many of the existing api endpoints as possible. I'm not however setup to test on ie11, so that's the only danger point I think.

test/attributes/hx-sse.js
```diff
@@ -4,43 +4,53 @@ describe("hx-sse attribute", function() {
...
-                    event.data = data;
-                    listener(event);
+            sendEvent: function(eventName, data) {
+                var listeners = listeners[eventName];
+                if (listeners) {
+                    listeners.forEach(function(listener) {
+                        var event = htmx._("makeEvent")(eventName);
+                        event.data = data;
+                        listener(event);
+                    }
```
rbri, line 28: closing ) missing -> syntax error

test/ext/sse.js
```diff
@@ -0,0 +1,232 @@
...
+                listeners[message].push(l)
+            },
+            sendEvent: function(eventName, data) {
+                var listeners = listeners[eventName];
+                if (listeners) {
+                    listeners.forEach(function(listener) {
+                        var event = htmx._("makeEvent")(eventName);
+                        event.data = data;
+                        listener(event);
+                    }
```
rbri, line 31: and again closing ')' missing

PR #2039, opened 2023-11-23
Removed IE-specific tests

Telroshan:
As discussed with @1cg, removed special checks in tests that were specific to IE, as we're going to drop support for it


alexpetros:
You can just push directly to the branch, no need to PR.

PR #2045, opened 2023-11-26
Improve performance of lib during initialization phase

mihalikv:
## Description
If I have following page:
```
<div hx-boost="true">
    <a href="https://htmx.org/">asdasdasd</a>
</div>
<a href="https://www.google.com/">test</a>
... thousand of same link
<a href="https://www.google.com/">test</a> 
```
htmx on load will init all of the links on the page. However, it's necessary to init htmx only on one link inside hx-boost wrapper. This can result in poor performance on pages with lot of "plain" links.

Performance on page with 1K links (6 x slowdown):
Before:
image

After:
image

## Testing
I have test the performance and simple boosted link.

## Checklist

  source changes)
  approved via an issue - **I hope that it's bugfix**

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1894,7 +1894,7 @@ return (function () {
             });
         }
 
-        function hasChanceOfBeingBoosted() {
+        function getBoostedWrappers() {
             return document.querySelector("[hx-boost], [data-hx-boost]");
         }
 
@@ -1934,10 +1934,17 @@ return (function () {
 
         function findElementsToProcess(elt) {
             if (elt.querySelectorAll) {
-                var boostedElts = hasChanceOfBeingBoosted() ? ", a" : "";
-                var results = elt.querySelectorAll(VERB_SELECTOR + boostedElts + ", form, [type='submit'], [hx-sse], [data-hx-sse], [hx-ws]," +
+                var boostedWrappers = getBoostedWrappers();
+                var boostedElts = [];
+                boostedWrappers.forEach(boostWrapper => {
+                    boostedElts.push(...boostWrapper.querySelectorAll('a'))
+                });
+                var results = elt.querySelectorAll(VERB_SELECTOR + ", form, [type='submit'], [hx-sse], [data-hx-sse], [hx-ws]," +
                     " [data-hx-ws], [hx-ext], [data-hx-ext], [hx-trigger], [data-hx-trigger], [hx-on], [data-hx-on]");
-                return results;
+                return [
+                    ...results,
+                    ...boostedElts,
+                ];
             } else {
                 return [];
             }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1894,10 +1894,6 @@ return (function () {
             });
         }
 
-        function getBoostedWrappers() {
-            return document.querySelector("[hx-boost], [data-hx-boost]");
-        }
-
         function shouldProcessHxOn(elt) {
             var attributes = elt.attributes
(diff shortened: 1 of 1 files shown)


fercomunello:
I also noticed this performance drop in the node processing phase, your contribution is a very relevant improvement for me and I think it should be included in the next version! =)

src/htmx.js
```diff
@@ -1923,10 +1919,13 @@ return (function () {
 
         function findElementsToProcess(elt) {
             if (elt.querySelectorAll) {
-                var boostedElts = hasChanceOfBeingBoosted() ? ", a" : "";
-                var results = elt.querySelectorAll(VERB_SELECTOR + boostedElts + ", form, [type='submit'], [hx-sse], [data-hx-sse], [hx-ws]," +
+                var boostedElts = document.querySelectorAll("[hx-boost] a, [data-hx-boost] a");
```
Telroshan, line 1922: Shouldn't we also query for `a[hx-boost]` and `a[data-hx-boost]` there, i.e. query the links that define themselves a `hx-boost` attribute instead of only links that are children of a boosted element?

src/htmx.js
```diff
@@ -1923,10 +1919,13 @@ return (function () {
...
+                var results = elt.querySelectorAll(VERB_SELECTOR + ", form, [type='submit'], [hx-sse], [data-hx-sse], [hx-ws]," +
                     " [data-hx-ws], [hx-ext], [data-hx-ext], [hx-trigger], [data-hx-trigger], [hx-on], [data-hx-on]");
-                return results;
+                return [
+                    ...results,
+                    ...boostedElts,
```
Telroshan, line 1927: Note that the spread operator is [not IE11 compatible](https://caniuse.com/?search=spread%20operator), and htmx (1, IE11 support will be dropped for htmx 2) strives for IE11 compatibility

src/htmx.js
```diff
@@ -1923,10 +1919,10 @@ return (function () {
 
         function findElementsToProcess(elt) {
             if (elt.querySelectorAll) {
-                var boostedElts = hasChanceOfBeingBoosted() ? ", a" : "";
-                var results = elt.querySelectorAll(VERB_SELECTOR + boostedElts + ", form, [type='submit'], [hx-sse], [data-hx-sse], [hx-ws]," +
+                var boostedElts = document.querySelectorAll("[hx-boost] a, [data-hx-boost] a, a[hx-boost], a[data-hx-boost]");
+                var results = elt.querySelectorAll(VERB_SELECTOR + ", form, [type='submit'], [hx-sse], [data-hx-sse], [hx-ws]," +
                     " [data-hx-ws], [hx-ext], [data-hx-ext], [hx-trigger], [data-hx-trigger], [hx-on], [data-hx-on]");
-                return results;
+                return toArray(results).concat(toArray(boostedElts));
```
Telroshan, line 1925: Now that I see it, couldn't you just add the hx-boost selectors to `result`'s querySelectorAll, and avoid having to concat the arrays altogether as you would retrieve all elements with a single `querySelectorAll` call?

Telroshan:
Btw, is the performance still good with the requested changes @mihalikv ?
Just to make sure it still matches your initial intent, as the numbers above come from the first version of this PR if I'm not mistaken

alexpetros:
Needs to be rebased but this is a good change

mihalikv:
> Btw, is the performance still good with the requested changes @mihalikv ? Just to make sure it still matches your initial intent, as the numbers above come from the first version of this PR if I'm not mistaken

Yeah it looks good.

PR #2053, opened 2023-11-28
Use faster `regexp.test` over `string.match` for testing against regex

Encephala:
## Description
As was mentioned by @xhaggi [in this code review](https://github.com/bigskysoftware/htmx/pull/1913/files/2389b43a399ff62b14ec9c9fa71bab6d46ad5fe9#r1388211677), `regex.test` is faster than `string.match` if the goal is to check if a string matches a regular expression. `string.match` is only necessary if the goal is to extract a substring.

Out of the six places in HTMX where `string.match` is used, four of them can be replaced by `regexp.test`.

Corresponding issue: None, suppose that's not necessary for this small change, but correct me if I'm wrong.

## Testing
I ran the testing suite and everything came out green.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -274,7 +274,7 @@ return (function () {
         }
 
         function aFullPageResponse(resp) {
-            return resp.match(/<body/);
+            return /<body/.test(resp)
         }
 
         /**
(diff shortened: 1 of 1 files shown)


Telroshan:
To back up your proposal, just linking [this SO thread](https://stackoverflow.com/a/10940138/12742595) where a user posted some performance measurements between those 2

PR #2068, opened 2023-12-04
Responsive layout example

jkmnt:
## Description

This brings the simple but useful responsive layout example to the website.
The english is not my mother tongue, so treat it as a rough sketch. 
I just can't write it correctly, but the urge to share the little trick is huge )

## Checklist

  source changes)
  approved via an issue


alexpetros:
I don't think this example really fits with what we're looking for in this section, which is primarily simple examples of htmx core functionality.

Also, css really is the first thing you should reach for when doing responsive layouts.

PR #2074, opened 2023-12-07
Update _index.md

elabx:
## Description
Update docs to reflect the behavior of an extension that wants to leverage handleSwap.

### DIFF
--- a/www/content/extensions/_index.md
+++ b/www/content/extensions/_index.md
@@ -114,3 +114,4 @@ Extensions can override the following default extension points to add or change 
     encodeParameters : function(xhr, parameters, elt) {return null;}
 }
 ```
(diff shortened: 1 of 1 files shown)


alexpetros:
I think the general point you're trying to make is that the extension version of any function can be overwrriten on the element, after the extension is loaded, right? It has nothing to do with `handleSwap` in particular?

elabx:
I was only working with handleSwap so it's the only one I could argue for. But yes, it was confusing that the extension's handleSwap was never reached, couldn't wrap my around around the fact that this is not like hooking into the events before/after but an actual replacement for what hx-swap configures.-

I'll think a bit more on how to make clear what I want to communicate in the docs. 

Thanks for reading!

alexpetros:
I'll think on it too.

Issue #997, opened 2022-08-04
Add Shadow dom support


OnurGumus:
The previous issue is closed but I think it should have been open. So I create this issue
[ Does HTMX work within shadow dom?](https://github.com/bigskysoftware/htmx/issues/647)

Primarily the problem is htmx use document.querySelector to access other elements but that approach won't work when your component is in a shadow dom. Hence the way we access elements should be adjusted.

rdehuyss:
I agree - I would love to use htmx with [Shoelace](https://github.com/shoelace-style/shoelace).

But if I do a `hx-put` or `hx-post` on an html form, then the html `input` fields within the Shoelace elements is not found and thus not posted.

OnurGumus:
In case anyone want it, I have a fork here that works with shadow dom. But you need to use alpine as hyperscript doesn't support retriggering
https://github.com/OnurGumus/htmx-shadow

saolof:
@OnurGumus  Neat. To be clear, do you mean that it needs the alpine morph plugin to work? Or just that it does not support hyperscript, and that nothing changes for basic use?

OnurGumus:
Well the thing is all htmx, hyperscript and alpine rely on observing dim mutations when new code is injected. htmx has processNode and alpine has initTree to manually initiate dom parsing. Since neither can observe behind shadow dom. hyperscript doesn’t have anything like that.

saolof:
I see that there is also a PR for the fork at #1191 ?

OnurGumus:
Yes that PR is mine.

zapling:
Would love to see this supported, any input on what we need to do in order to get this implemented? @alexpetros

alexpetros:
> I agree - I would love to use htmx with [Shoelace](https://github.com/shoelace-style/shoelace).
> 
> But if I do a `hx-put` or `hx-post` on an html form, then the html `input` fields within the Shoelace elements is not found and thus not posted.

@rdehuyss Check out the shoelace discussion here if you haven't seen it: https://github.com/bigskysoftware/htmx/issues/1323

> Would love to see this supported, any input on what we need to do in order to get this implemented?

This has come up before, I think it's just a sufficiently substantial change that it requires thorough review, tests, etc. #997 is out of date right now.

rdehuyss:
Hi @alexpetros - thanks for following up!

Tbh, I've switched to Bulma so it's not a problem for me anymore.

OnurGumus:
[...]

PR #2075, opened 2023-12-07
Add support for shadow DOM

kgscialdone:
## Description
Adds basic support for using HTMX inside of shadow DOM. HTMX attributes will not be automatically sought out inside of shadow roots, but can be enabled by calling `htmx.process(elt.shadowRoot)` on desired elements. All basic functionality should work, and will treat the containing shadow root as if it were the entire document unless otherwise specified.

This also expands the extended selector syntax used by `hx-target` et al. to include two new options:
- `root` - the current shadow root, or `document` if not applicable.
- `global <extended selector>` - selects from `document` rather than the current shadow root.

Note that this is not a fully tested feature; for example, I anticipate history support potentially running into issues with it. However, per 1cg's instructions, these issues have been left for later resolution.

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/997

## Testing
Added the `core/shadowdom.js` test suite, which replicates all tests from `core/ajax.js` with minor adjustments to function in shadow DOM, alongside basic tests for the new additions to the extended selector syntax.

## Checklist

* [x] I have targeted this PR against the correct branch (`v2.0v2.0` as requested on Discord)
* [x] This is a new feature that has been explicitly approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -186,6 +186,13 @@ return (function () {
         function getDocument() {
             return document;
         }
+        
+        /**
+         * @returns {Document | ShadowRoot}
+         */
+        function getRootNode(elt, global) {
+            return elt.getRootNode ? elt.getRootNode({ composed: global }) : getDocument();
+        }
 
         /**
          * @param {HTMLElement} elt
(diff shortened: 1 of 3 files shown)


alexpetros:
This is sick

andryyy:
She crazy.

emfax:
Was just having a play around with this and it doesn't seem to work with OOB swaps at the moment. Is this planned or should I try to provide an implementation?

PR #2090, opened 2023-12-12
Prevent references to elements swapped out of the dom accumulating as detached elements (memory leak).

croxton:
## Description
I noticed that detached nodes accumulated in a project. The memory heap contained references to `htmx-internal-data` and a `replacedWith` property caused by this line:

Ref:
https://github.com/bigskysoftware/htmx/blob/206912b71d355477377aade60d06ea013c8e5455/src/htmx.js#L990

Important notes: 
* I do not understand the purpose of the `replacedWith`  property, there are no other references to it in the source code.
* I am also deleting the element's `htmx-internal-data` in `cleanUpElement()` because some elements would retain content in this structure (with references to elements no longer in the dom) _despite_ the preceding cleanup - I do not know why they sometimes do, or if it is actually wise to delete it at this point.

## Testing
Observed manually using Chrome's developer tools to take heap snapshots while performing a sequence of typical user tasks on a complex htmx application. Verified with Microsoft Edge which has a useful 'Detached Elements' feature.

Passes htmx automated tests.

## Checklist

  source changes)
  approved via an issue


croxton:
Sorry, going to do it again on the dev branch.

PR #2091, opened 2023-12-12
Prevent references to nodes swapped out of the dom accumulating as detached elements in memory

croxton:
## Description
I noticed that detached nodes accumulated in a project after multiple swaps. Memory heap snapshots contained detached elements with references to `htmx-internal-data` and a `replacedWith` property caused by this line:

Ref:
https://github.com/bigskysoftware/htmx/blob/206912b71d355477377aade60d06ea013c8e5455/src/htmx.js#L990

Important notes: 
* I do not understand the purpose of the `replacedWith`  property, there are no other references to it in the source code.
* I am also deleting the element's `htmx-internal-data` in `cleanUpElement()` because some elements would retain content in this structure (with references to elements no longer in the dom) _despite_ the preceding cleanup - I did not investigate why, or determine if there is ever a reason to retain data in `htmx-internal-data` when an element is "cleaned up".

## Testing
Observed manually using Chrome's developer tools to take heap snapshots while performing a sequence of typical user tasks on a complex htmx application. And checked with Microsoft Edge which has a useful 'Detached Elements' feature. These changes prevented detached nodes accumulating.

Passes htmx automated tests.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -972,6 +972,9 @@ return (function () {
             if (element.children) { // IE
                 forEach(element.children, function(child) { cleanUpElement(child) });
             }
+            if (element.hasOwnProperty('htmx-internal-data')) {
+                delete element['htmx-internal-data'];
+            }
         }
 
         function swapOuterHTML(target, fragment, settleInfo) {
@@ -987,7 +990,6 @@ return (function () {
                 } else {
                     newElt = eltBeforeNewContent.nextSibling;
                 }
-                getInternalData(target).replacedWith = newElt; // tuck away so we can fire events on it later
                 settleInfo.elts = settleInfo.elts.filter(function(e) { return e != target });
                 while(newElt && newElt !== target) {
                     if (newElt.nodeType === Node.ELEMENT_NODE) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -960,10 +960,8 @@ return (function () {
                     }
                 });
             }
-            if (internalData.initHash) {
-                internalData.initHash = null
-            }
             deInitOnHandlers(element);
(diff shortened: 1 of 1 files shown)


1cg:
Thank you @croxton!

grey4owl:
@croxton @1cg 
Well, that's what broke [my extension](https://github.com/grey4owl/htmx-preserve-attr) I was using this extension in my project and it worked really well but I didn't update htmx in a long time until now but after update it doesn't work anymore.. So i started to investigate and I found out that `replacedWith` doesn't exist anymore :disappointed: My project depends heavily on this extension and the reason why it worked so well is because I could replace the attributes of next target element before the request was executed, which I really needed so that it would not have problems later with alpinejs. So is there any workaround of this?

croxton:
@grey4owl Looking at your extension it should be possible to store the element attributes you need before the swap happens on the` htmx:beforeSwap` event? Alternatively, there is an Alpine Morph extension for preserving the component state that you could try (I use this myself, it works really well): https://github.com/bigskysoftware/htmx-extensions/blob/main/src/alpine-morph/README.md

grey4owl:
@croxton And yes I know that you can get new target after swapping and you can set new attributes but then you get alpinejs problems.. So I need a way to set custom attributes to new element that is gonna be swapped but before request is finished.

grey4owl:
@croxton I mange to make it work like you explained but i used `event.detail.serverResponse` since `event.detail.xhr.response` is "read-only". But is this the only way to make this? Somehow it feels like wrong way of doing this..

croxton:
Great! Htmx's events are designed to allow listeners to mutate the objects passed to them, so I don't think it's wrong. You're extending htmx, exactly as intended.

PR #2095, opened 2023-12-14
Remove legacy WS/SSE code

Renerick:
## Description

I've removed all mentions of WS and SSE from core htmx.js, from the test server and also removed tests

Corresponding issue: 2.0 development 2.0

Probably could use a revision after merging #2025 

## Testing

I've checked that all tests (excluding removed ones) are 100% parsing. I also manually tested that all extension demos on the test server are working

## Checklist

  source changes)
  approved via an issue


Renerick:
After merging updated SSE implementation, it appears that it was a bit more reliant on legacy sse handling of triggers in the core, so I had to clean it up a bit, as well as removing legacy attributes handling from it. I think it's worth it to port this clean ups to 1.0 to make extension a bit more stable.

PR #2098, opened 2023-12-15
Clarified docs on hx-disable

feakuru:
## Description
Clarified docs on hx-disable according to changes in #1567 

## Testing
N/A

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/attributes/hx-disable.md
+++ b/www/content/attributes/hx-disable.md
@@ -10,4 +10,10 @@ The value of the tag is ignored, and it cannot be reversed by any content beneat
  
 ## Notes
 
-* `hx-disable` is inherited
+* `hx-disable` is inherited.
+* When dynamically adding hx-disable to an element from your custom code:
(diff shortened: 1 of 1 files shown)


alexpetros:
This is a little too weeds-y for the documentation, I'm afraid. I think we could maybe discuss consolidating documentation about what attributes can or cannot be added dynamically at some point, but this isn't the way to do it.

feakuru:
> This is a little too weeds-y for the documentation, I'm afraid.

I was mainly trying to add info on the necessity to call `htmx.process` right after adding and deleting the attribute, since I encountered this problem in a project and found the info on the function I needed to call only after going through commits and finding it in a PR description. I believe it should be made obvious in the docs in some way.

>  I think we could maybe discuss consolidating documentation about what attributes can or cannot be added dynamically at some point, but this isn't the way to do it.

This would be a good way to do it in my opinion. Right now there's no obvious way to tell from the docs.

PR #2038, opened 2023-11-23
fix src/htmx.js(3842,25): error TS2769: No overload matches this call.

JLarky:
## Description

This fixes one of the typescript issues that I see when I run `test-types` command

```
src/htmx.js(3842,25): error TS2769: No overload matches this call.
  Overload 1 of 2, '(this: WindowEventHandlers, ev: PopStateEvent): any', gave the following error.
    The 'this' context of type 'void' is not assignable to method's 'this' of type 'WindowEventHandlers'.
  Overload 2 of 2, '(this: Window, ev: PopStateEvent): any', gave the following error.
    The 'this' context of type 'void' is not assignable to method's 'this' of type 'Window'.
```
 
image

Corresponding issue:

I couldn't find it

## Testing

I was running `test-types` before and after this change I saw that issue is not there anymore

## Checklist

* [x] I have read the contribution guidelines
* [x] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [ ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
  - this PR has very strong `Refactors that do not make functional changes will be automatically closed, unless explicitly solicited. Imagine someone came into your house unannounced, rearranged a bunch of furniture, and left.` vibe, I'm doing this just to gauge if I can help with type improvements in this repo :)
* [x] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #2107, opened 2023-12-19
introduce types.ts; improve types for internal functions with focus on hx-swap

JLarky:
## Description

> previous [PR](https://github.com/bigskysoftware/htmx/pull/2038)

- Add `HtmxExtension.init type` (src/htmx.d.ts)
- create new file `src/types.ts` based on code form @trvswgnr https://gist.github.com/trvswgnr/7239fa78cd12cee986384623a29b9b89
  - it has a lot of `any`s but it's better than nothing :)
- improve types for hx-swap, this brings the number of errors in `test-types` from 21 to 15
- add types to `internalAPI` (the ones that @trvswgnr wrote)

This PR will introduce a new file `dist/types.ts` into the mix, I'm not 100% sure if that's good or bad :)
So let's talk a bit about alternatives:
 - inline everything in `.js` file with jsdoc syntax --- that's what we were doing so far, not the best ergonomics
 - write types inside `htmx.d.ts` --- that works great for external types like `HtmxConfig` or `html.addClass`, but you don't want to put `splitOnWhitespace` there
 - write types inside `types.d.ts` --- this is what [SvelteKit](https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/client/types.d.ts) is doing and I first started with that approach when I started doing JsDoc projects, but I found it a bit clumsy, the biggest issue for me was how confusing it would work in case if you have both `something.js` and `something.d.ts` file
 - write types inside `types.ts` --- that's what someone recommended to me to try instead of doing `.d.ts` and I was happy ever since :D (but I'm not married to that)

So most of this PR is done, there are two outstanding things that I wanted to discuss before I turn this PR from Draft to Ready:
- `HtmxExtension.init` --- this method takes `HtmxInternalApi` as first argument, so that means that `HtmxInternalApi` kind of have to be public API now, and if we are adding types of `HtmxInternalApi` to `htmx.d.ts` then you have to add all additional types like `HtmxTriggerSpecification` or `TriggerHandler` to it and so on.
  - so we could just use the type `init(apiRef: import("./types").HtmxInternalApi): void;` which works, but sends a signal that "there will be dragons"
  - I'm looking for feedback in terms of do we expect 3rd party extensions to consume those types (as in an npm module that will be written in TS that imports internal types to work) or even rewrite existing extensions to use those types, right now there's basically no type checkin in src/ext 
- how much of the stuff from `src/types.ts` should actually be in `htmx.d.ts`? In the original gist it feels like it was intended to be a part of public api https://gist.github.com/trvswgnr/7239fa78cd12cee986384623a29b9b89

## Testing
     npm run test-types

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,0 +1,151 @@
+interface HtmxTriggerSpecification {
+  sseEvent?: string;
+  trigger: string;
+  root?: Element | Document | null;
+  threshold?: string;
+  delay?: number;
+  pollInterval?: number;
+}
+
+interface HtmxSwapSpecification {
+  swapStyle: SwapStyle;
+  swapDelay: number;
+  settleDelay: number;
+}
+
+interface HtmxSettleInfo {
+  title?: string;
+  tasks?: (() => void)[];
+  elts?: Element[];
+}
+
+interface ListenerInfo {
+  listener: EventListener;
+  on: HTMLElement;
+  trigger: string;
+}
+
+/** the http verb used in the request (lowercase) */
+type Verb = "get" | "post" | "put" | "delete" | "patch";
+
+interface KnownInternalData {
+  initHash?: number | null;
+  listenerInfos?: ListenerInfo[];
+  path?: string;
+  streamPaused?: boolean;
+  streamReader?: ReadableStreamDefaultReader;
+  verb?: Verb;
+  lastButtonClicked?: Element | null;
+  timeout?: number;
+  webSocket?: WebSocket;
+  sseEventSource?: EventSource;
+  onHandlers?: { event: string; listener: EventListener }[];
+  xhr?: XMLHttpRequest;
+  requestCount?: number;
+}
+
+type InternalData = KnownInternalData & Record<PropertyKey, any>;
+
+interface InputValues {
+  errors: any[];
+  values: Record<string, string>;
+}
+
+interface Pollable {
+  polling: boolean;
+}
+
+interface TriggerHandler {
+  (elt: Element, evt: Event): void;
+  (): void;
+}
+
+type SwapStyle =
+  | "innerHTML"
+  | "outerHTML"
+  | "afterbegin"
+  | "beforebegin"
+  | "afterend"
+  | "beforeend"
+  | "none"
+  | "delete";
+
+interface HtmxInternalApi {
+  addTriggerHandler(
+    elt: Element,
+    triggerSpec: HtmxTriggerSpecification,
+    nodeData: Pollable,
+    handler: TriggerHandler
+  ): void;
+  bodyContains(elt: Node): boolean;
+  canAccessLocalStorage(): boolean;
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -0,0 +1,174 @@
...
+export type SwapStyle =
+  | "innerHTML"
+  | "outerHTML"
+  | "beforebegin"
+  | "afterbegin"
+  | "beforeend"
+  | "afterend"
+  | "delete"
+  | "none";
+
```
trvswgnr, line 88: we should probably add `(string & NonNullable<unknown>)` like this:
```ts
export type SwapStyle =
    | "innerHTML"
    | "outerHTML"
    | "afterbegin"
    | "beforebegin"
    | "afterend"
    | "beforeend"
    | "none"
    | "delete"
    | (string & NonNullable<unknown>);
```

because there are other swap styles (like `morph:innerHTML` when using idiomorph)

this way, editor autocomplete will still be there for known values, but any string *can* be passed.

trvswgnr:
you can add me as co-author on this 😘

```


Co-authored-by: trav <trvswgnr@gmail.com>
```

JLarky:
> you can add me as co-author on this 😘

will do :) do you have any thoughts on `types.ts` alternatives? :)

trvswgnr:
> > you can add me as co-author on this 😘
> 
> will do :) do you have any thoughts on `types.ts` alternatives? :)

i think that devs authoring their own extensions will usually do so in their own environment and not as part of the htmx project (at least initially), so it would make sense that those types are exposed somehow. i think "internal api" is a little misleading, as it gets used throughout extensions, making it more of a "developer api".

the only reason i would be for .d.ts instead of .ts is that the latter might lead people to believe that there is logic in there.

src/types.ts
```diff
@@ -0,0 +1,174 @@
...
+ * into the JsDoc comments in the htmx.js file. */
+
+interface HtmxTriggerSpecification {
+  sseEvent?: string;
+  trigger: string;
+  root?: Element | Document | null;
+  threshold?: string;
+  delay?: number;
+  pollInterval?: number;
+}
```
trvswgnr, line 14: looked through and updated this with keys found in source:

```ts
export interface HtmxTriggerSpecification {
    trigger: string;
    sseEvent?: string;
    eventFilter?: string;
    changed?: boolean;
    once?: boolean;
    consume?: boolean;
    from?: string;
    target?: string;
    throttle?: number;
    queue?: string;
    root?: string;
    threshold?: string;
    delay?: number;
    pollInterval?: number;
};
```

alexpetros:
Hey all, we're about to do a pretty significant refactor and unfortunately now isn't the right time to introduce this. The htmx typing story is at the moment incomplete—if you want to update the API types you're more than welcome to.

JLarky:
@alexpetros I need a bit more guidance here :)

alexpetros:
Haha the guidance is, for the moment, please hold off on type changes.

PR #2120, opened 2023-12-21
Support encapsulating oob swap elements in templates in response

Telroshan:
As discussed in the Discord, here's a template-based suggestion to work around the annoying scenario of oob-swapping the lone `tr` or `td` to, which by default won't work if there are any elements before it in the response, because of the way the DOM parser works.

Updated the docs to mention that scenario and the way we would now recommend to address it.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -829,6 +829,15 @@ var htmx = (function() {
     return oobValue
   }
 
+  function findAndSwapOobElements(fragment, settleInfo) {
+    forEach(findAll(fragment, '[hx-swap-oob], [data-hx-swap-oob]'), function (oobElement) {
+      const oobValue = getAttributeValue(oobElement, "hx-swap-oob");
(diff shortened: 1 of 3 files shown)


src/htmx.js
```diff
@@ -865,10 +865,22 @@ return (function () {
...
+
+            function findAndSwapOobElements(fragment) {
+                forEach(findAll(fragment, '[hx-swap-oob], [data-hx-swap-oob]'), function (oobElement) {
+                    var oobValue = getAttributeValue(oobElement, "hx-swap-oob");
+                    if (oobValue != null) {
+                        oobSwap(oobValue, oobElement, settleInfo);
+                    }
+                });
+            }
+
```
1cg, line 877: request to pull this out to same level as other functions

1cg (changes requested):

PR #2129, opened 2023-12-24
Update `htmx.ajax` type signature to make the `element` argument optional

PizzasBear:
I updated the type signature of `htmx.ajax` to allow the user to not provide a value to the `element` argument and just use the default documented value, `body`.


alexpetros (approved):
Cool, looks good

Issue #423, opened 2021-03-17
Let oob use innerHTML for source (and outerHTML for target)


guettli:
I want to use [hx-swap-oob](https://htmx.org/attributes/hx-swap-oob/) to replace a table row of the existing page "out of band".

HTML already in browser (target):
```
<table>
 <tr id="offer_1">....</tr>
 <tr id="offer_2">....</tr> (old)
 <tr id="offer_3">....</tr>
</table>
```

I want to add this new table row (source coming from server)
```
 <tr id="offer_2" class="some-new-css-class"> .... </tr> (new)
```

Up to now hx-swap-oob does take the outerHTML from the source.

I would like to take the innerHTML, because of one reason:

On the server I have a method to create the table row. This method should not know the fact that result of the method gets used in hx-swap-oob. 

I am willing to provide a pull request.

Up to now I am unsure about how solution should look like at the end.

Here is one of many options:
```
<table hx-swap-oob="innerHTML-source">
 <tr id="offer_2" class="some-new-css-class"> .... </tr> (new)
</table>
```

`innerHTML-source` tells htmx to ignore the <table> tag and take the innerHTML as source.

Would you accept a PR?

BTW, #422 would be nice, too (Docs about how to create a PR).

1cg:
I'm open to the idea.  Maybe use the https://htmx.org/attributes/hx-select/ attribute in conjunction w/ the oob-swap?  Or a `select:` modifier in `hx-oob-swap`?

I'm good w/ any pull request, just add tests :)

adicco:
Hey @guettli, I was actually looking into this myself. Have you found anything else in the meantime?

guettli:
@adicco sorry, I don't recall how I solved this. I know that there was an issue with sending `<tr>...</tr>` out-of-band.

But that was solved by carlson: https://github.com/bigskysoftware/htmx/issues/469

Maybe you get a better answer if you ask in the discord channel or at Stackoverflow.

adicco:
Thanks for the quick reply @guettli 

In my case I am using a Bun/Elysia stack with JSX templating, and it would be neat to be able to do something like

```jsx
<div hx-swap-oob="outerHTML:#my-target-component" hx-swap-oob-source="innerHTML">
<MyComponent />
</div>
```

Because presently the ```<div>``` itself replaces the target's outerHTML. Of course I can get by by doing:

```jsx
<MyComponent hx-swap-oob="..." />
```

[...]

PR #2169, opened 2024-01-09
Add support for hx-swap-oob-source as "firstChild" (or "innerHTML")

adicco:
## Description

Added support for a new `hx-swap-oob-source` attribute which presently can only take one value, `firstChild` (in line with the DocumentFragment API.) If set to `firstChild`, the swapped node is not the node with the `hx-swap-oob` attribute but its first child.

At the moment this silently fails if no child node is passed, but it might be desirable to throw an error, which I'd be happy to implement.

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/423

## Testing

I added one test to the hx-swap-oob suite.

## Checklist

  source changes)
  approved via an issue **No, but the change is so minimal that I thought I'd offer an actual PR** 

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -819,7 +819,7 @@ return (function () {
                         var oobElementClone = oobElement.cloneNode(true);
                         fragment = getDocument().createDocumentFragment();
                         fragment.appendChild(oobElementClone);
-                        if (!isInlineSwap(swapStyle, target)) {
+                        if (!isInlineSwap(swapStyle, target) || getAttributeValue(oobElement, "hx-swap-oob-source") === "firstChild") {
                             fragment = oobElementClone; // if this is not an inline swap, we use the content of the node, not the node itself
                         }
 
--- a/test/attributes/hx-swap-oob.js
+++ b/test/attributes/hx-swap-oob.js
@@ -128,5 +128,16 @@ describe("hx-swap-oob attribute", function () {
         this.server.respond();
         should.equal(byId("d1"), null);
     });
+
+    it('handles hx-swap-oob-source properly', function()
+    {
+        this.server.respondWith("GET", "/test", '<div hx-swap-oob="true" hx-swap-oob-source="firstChild" id="d1"><div id="d2">Swapped</div></div>');
+
+        var div = make('<div id="d1" hx-get="/test">Foo</div>')
+        div.click();
+        this.server.respond();
+        should.equal(byId("d1"), null);
+        byId("d2").innerHTML.should.equal("Swapped");
+    });
 });
 


adicco:
Actually, `firstChild` might not be the right terminology, as several children inside would all be swapped. Perhaps `innerHTML` then?

Atmos4:
Why would you want to not swap inline? Can you provide more details about the use case you are trying to cover here?

The reason I am asking is that I think this functionality is not usable nor scalable. The persistance of OOB swap target across renders is what ensures that the same code can be run again.
In your example, running the test twice will fail because the target will have been swallowed by the swap.

OOB swaps are very useful for things like toasts or alerts that you would want to display somewhere else as a side effect of a request. In most (all?) cases where an OOB swap is useful, you would want to keep the target element so that it can be reused for following swaps.

If you come up with a proper use case, I think it will help understanding what you are trying to do.

adicco:
Hi @Atmos4 

You make a good point regarding repeatability. My use case is particularly related to component-based templates (say with a JavaScript backend or even Laravel Blade components and such like.)

What I wanted to avoid was to have any kind of oob-related attributes as part of my components' definition. Say I have the following Toast component in JSX (with an Elysia backend for instance)

```tsx
type ToastProps = {
  title: string;
  content: string;
  type: 'info' | 'success' | 'alert';
}

function Toast({ title, content, type }: ToastProps) {
return (
  <div class="toast">
    <h3>${title}</h3>
    <p>${content}</p>
  </div>
);
}
```

It would be neat to be able to do something like:

```html
<div hx-swap-oob="beforeend:#toasts" hx-swap-oob-source="firstChild">
  <Alert .../>
</div>
```

Rather than having to add extra, purely oob-related  complexity to the Toast component (assuming I have defined an extra `MaybeOobProps` type:)

function Toast({ title, content, type, ...oobProps }: ToastProps & MaybeOobProps) {
return (
  <div class="toast" {...oobProps}>
    <h3>${title}</h3>
    <p>${content}</p>
  </div>
);
}
```

Admittedly this is not a huge increase in complexity, but I appreciated the fact that the Toast component - or any other that might be swapped out of band - could define their own logic without having to know whether they would be swapped OOB or not.

Am I missing something? Do you have any tips regarding component-based architectures for views?

**Comment by Atmos4** [comment]
I would suggest against reasoning with framework-specific logic. All you need to be concerned about is the HTML you render, because this is what will be swapped by HTMX.

You may want it to always swap your toasts at the same place, e.g. `#toast-root`.
Take this example structure:
```html
<body>
  <main>
    <div hx-get="/my-content" hx-trigger="load" hx-swap="innerHTML"></div>
  </main>
  <div id="toast-root">
     <!-- your toasts -->
  </div>
</body>
```

Let's say that you want to OOB swap a toast to `#toast-root` when requesting `GET /my-content`. Then you can return this HTML:
```html
<!-- example content of the request -->
<p>Hello world</p>

<!-- oob swap -->
<div id="toast-root" hx-swap-oob="innerHTML">
  <!-- your toast -->
</div>
```

If you want to learn more about out-of-band swaps and how it works, I suggest taking a look at [the docs](https://htmx.org/attributes/hx-swap-oob/) and experimenting by yourself as it is the best way to learn!

adicco:
Okay, thank you for taking the time to write this. I understand the framework-agnostic approach, and wouldn't have made the PR if I thought this would somehow bloat the core in a significant way. However, based on the scalability and repeatability issues you mentioned, it makes sense to no want to include this in the core. I'll figure out a way to keep things clean in my templates. 

I also understand that the extension API is evolving in 2.0 - we might have access to more internal methods? - so I might be able to do something with that.

Atmos4:
Is my approach solving your problem though? Let me know if you need more help, but it sounds like what I described above is what you wanted to do!

adicco:
Dear @Atmos4,

Although I am grateful you took the time to pitch in, I didn't learn anything new from your hints, in the sense that I was well aware of how oob works, of their purpose, and of the documentation. 

My concerns were not why and how to "spit them out" in the final HTML, but rather how to not bloat up my components with purely oob-related logic. Admittedly this is somewhat framework-dependent, but many frameworks - not only JS ones - do use the component paradigm for their views (Laravel Blade templates, for one.)

In any case, I think I have found an approach I am satisfied with which wouldn't require adding support for a new oob attribute and wouldn't require my components to handle oob logic, which is basically the way  ["Click to Load" is implemented.](https://htmx.org/examples/click-to-load/)

It looks like this:

```html
<body>
  <main>
    <div hx-get="/my-content" hx-trigger="load" hx-swap="innerHTML"></div>
  </main>
  <div id="toast-root">
     <!-- your toasts -->
     <div id="toasts-next"></div>
  </div>
</body>
```

And then on `GET /my-content` I could return:

```html
<!-- example content of the request -->
<p>Hello world</p>

<!-- oob swap with embedded toasts-next-->
<div hx-swap-oob="outerHTML:#toasts-next">
 <Toast />
 <div id="toasts-next"></div>
</div>
```

Atmos4:
> I didn't learn anything new from your hints

I think you did learn something new :) otherwise you would not have:
- closed your PR
- reused my code in your own example

I am glad I could help in some small way then!

It can be further simplified though, similar to how I have done in my example. Your `GET /content` response can look like this:

```html
<div id="toast-root" hx-swap-oob="innerHTML"> <!-- or afterbegin -->
 <Toast />
</div>
```

adicco:
I think I would still need to embed a new `#toast-next` container in my response, wouldn't I? Otherwise it will break subsequent oobs (in the same way that the [Click To Load implementation](https://htmx.org/examples/click-to-load/) has a 'special' `#replaceMe` final row, which is nested/reincluded on each further load more call)

Atmos4:
> click to load

This example isn't about OOB at all 🙈 instead you should check out [Updating other content](https://htmx.org/examples/update-other-content/)

I took some time to make [a CodePen](https://codepen.io/atmos4/pen/bGZJBRe), hopefully it helps!

adicco:
The example is not about oob *per se*, but the 'embedding trick' used (bringing along a nested container that acts as the recipient for subsequent swaps)  is definitely helpful for what I was trying to accomplish.

adicco:
I have basically come full circle on this one. My original implementation of simply using

```html
<div hx-swap-oob="beforeend:#toasts">
   <div class="toast toast-info">Toast content</div>
</div>
```

Works as expected and results in the following DOM:

```html
<body>
  <div id="toasts" class="fixed top flex etc">
      <div class="toast toast-info">Toast content</div>
  </div>
</body>
```

For whatever reason, I had earlier had the parent container, which holds the oob attribute, swapped in as well, resulting in the following DOM after the oob swap:

```html
<body>
  <div id="toasts" class="fixed top flex etc>
      <div hx-swap-oob="beforeend:#toasts">
          <div class="toast toast-info">Toast content</div>
      </div>
  </div>
</body>
```

Not exactly sure why this is now working as expected, and why it wasn't before, but I will post if I discover anything more.

Atmos4:
Instead of `beforeend:#toasts`, maybe you should just use `id=toasts hx-swap-oob=beforeend`. I have never tried using a selector, but this is maybe what is wrong with your code.

Have you even tried some of the solutions I had? For example [this CodePen](https://codepen.io/atmos4/pen/bGZJBRe).

You seem to be very resistant to learning 😅

Issue #2167, opened 2024-01-09
Missed syntax errors in test/ext/sse.js


Renerick:
https://github.com/bigskysoftware/htmx/pull/2025#pullrequestreview-1809784832

Aside from fixing the tests, need to ensure that CI will not pass if similar problem occur in the future

hawi19952:
Hi, 
I think adding eslint to the project can be a viable solution and coupling that with the CI pipeline, so the pipeline doesn't approve unless the linter rules are passed. 

I'm not sure if that aligns with the philosophy of the maintainer, however, I'll give it a shot.

hawi19952:
Something I found out, the code is failing 7 tests, all of which are relevant to the syntax error. 

I'll make the PR now for lint, and will pick up the failing tests later @Renerick

Issue #844, opened 2022-02-26
Add `hx-swap` support to `hx-ext="sse"`


rajasimon:
Hi there, I've testing sse extension in my project and it's working as expected for swapping the content. However, I want to add new message below like chat message. And when I tried `hx-swap` with `hx-target` but that also seems to update the target instead of appending it. 

```
<div hx-ext="sse" sse-connect="/search/{{ query.id }}/" sse-swap="message" hx-target="#result" hx-swap="beforeEnd">
  Hang tight we are fetching your results.
  <div id="result"></div>
</div>
```
<img width="1181" alt="Screenshot 2022-02-26 at 1 04 16 PM" src="https://user-images.githubusercontent.com/6974101/155834642-4980fb7a-1390-45d3-b7b5-cec4fe8f6330.png">

I need the end result should be like append into each box. 

```
<div hx-ext="sse" sse-connect="/search/{{ query.id }}/" sse-swap="message" hx-target="#result" hx-swap="beforeEnd">
  Hang tight we are fetching your results.
  <div id="result">
    <div class="box></div>
    <div class="box></div>
    <div class="box></div>
    <!--  As soon as new message arrives this result get appended.-->
  </div>
</div>
```
Let me know if this possible with sse or any alternatives thanks.

ocharles:
I think this would also solve https://github.com/bigskysoftware/htmx/issues/784

benpate:
Yes.  Agreed.  Without duplicating *a whole lot* of code, this would require that html publish a workable `swap()` method for external scripts to use.  So this depends on #255

tdb-alcorn:
+1! I have exactly the same use case. Perhaps using htmx for a chat app is not in scope, but it feels like this would be a reasonable new application of existing htmx concepts.

benpate:
I've gotten htmx to work for for a chat app, but it takes some extra tinkering.  The easiest thing for me to do is was to use the SSE as a `hx-trigger` to load additional items whenever they were posted to the server.  It's an extra round-trip, but it made my server logic a lot easier.

To add a little more info to this issue, we specifically need access to the internal `swap()` function that I mentioned above.  I'm still holding out hope that we can get this included in the htmx 2.0 design.  This would solve outstanding issues with A LOT of extensions.

thallada:
[...]

PR #2171, opened 2024-01-10
Fix SSE tests and emit new "beforeMessage" event for SSE Extension

neelrr1:
## Description
I'm building a toy chat app with HTMX + the SSE extension. I have some custom JS that I want to run **before and after** new SSE messages are swapped in from the server.

For a regular request with HTMX, I could add listeners to the `htmx:beforeSwap` and `htmx:afterSwap` events. Similarly, the WS extension emits `htmx:wsBeforeMessage` and `htmx:wsAfterMessage`, but there is no corresponding ability to listen to "before swap" messages with the SSE extension.

This PR introduces a new event, `htmx:sseBeforeMessage`, that is emitted before the SSE extension swaps in content from the server. Open to feedback on naming convention!

## Testing
I tested this change manually by applying this patch to my local copy of the SSE extension.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/sse.js
+++ b/src/ext/sse.js
@@ -148,6 +148,7 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 					}
 
 					// swap the response into the DOM and trigger a notification
+					api.triggerEvent(elt, "htmx:sseBeforeMessage", event);
 					swap(child, event.data);
 					api.triggerEvent(elt, "htmx:sseMessage", event);
 				};
--- a/src/ext/sse.js
+++ b/src/ext/sse.js
@@ -148,7 +148,9 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 					}
 
 					// swap the response into the DOM and trigger a notification
-					api.triggerEvent(elt, "htmx:sseBeforeMessage", event);
+					if(!api.triggerEvent(elt, "htmx:sseBeforeMessage", event)) {
+						return;
+					}
 					swap(child, event.data);
 					api.triggerEvent(elt, "htmx:sseMessage", event);
 				};
--- a/test/attributes/hx-sse.js
+++ b/test/attributes/hx-sse.js
@@ -13,15 +13,15 @@ describe("hx-sse attribute", function() {
                 })
             },
             addEventListener: function(message, l) {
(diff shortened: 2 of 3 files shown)


neelrr1:
It may also make sense to rename the `sseMessage` event to `sseAfterMessage` but that would be a breaking change.

fhp-mec:
Hi,

Given that the naming is at least consistent with the ws version it seems like the best choice for the name of the event if thats what is decided to be the best solution.

That said, I do think that the future of the sse extension would probably be to use htmx's built in swap function rather than the bespoke one that is currently being used, and therefore emit both before and after swap events consistent with the existing htmx behavior. (see https://github.com/bigskysoftware/htmx/issues/844 for context)

If having `beforeSwap` and `afterSwap` events be fired by the swapping function would give you the functionality you need I would suggest that instead of adding a new event. 

I would also suggest that you add a regression test to the testing suite no matter what the solution is.

src/ext/sse.js
```diff
@@ -148,6 +148,7 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 					}
 
 					// swap the response into the DOM and trigger a notification
+					api.triggerEvent(elt, "htmx:sseBeforeMessage", event);
```
Renerick, line 151: This probably should provide cancellation support by wrapping this call in an if statement, like so 
https://github.com/bigskysoftware/htmx/blob/master/src/ext/ws.js#L131-L136

Also, would be nice to add tests for this event (and its cancellation)

neelrr1:
@Renerick, thanks so much for the first review and your comments. I've updated the code to add cancellation support and added tests to verify that the events are being emitted correctly. Can you take another look?

In the process, I also fixed our test suites for the SSE extension and `hx-sse` attribute. These were not running due to syntax errors in the test files, (#2167) which hid the fact that the tests were broken.

Renerick (approved):
LGTM

Especially thank you for fixing syntax errors in SSE tests

Renerick (approved):
LGTM

Special thank you for fixing syntax errors in SSE tests 🙂

neelrr1:
@Renerick are you able to run CI so we can get this merged?

PR #2199, opened 2024-01-16
Added Handlebars implementation

RayyanNafees:
## Description

Handlebars says itself to be "Mustache compatible" on their [website](https://handlebarsjs.com) but its implementation with HTMX [client-side-template](https://htmx.org/extensions/client-side-templates/) extension is slightly different and breaks when migrating to Handlebars from Mustache

I've added the Different implementation from mustache, needed to be followed when using handlebars

### Significantly two explicit changes are needed
1. Use the CDN upkg.com `/handlebars@latest/dist/handlebars.min.js` instead of `/handlebars@latest`
2. Use `{#each this}` for looping instead of `{#each data}` or `{#data}` inspired from mustache example

Corresponding issue:

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/extensions/client-side-templates.md
+++ b/www/content/extensions/client-side-templates.md
@@ -123,6 +123,79 @@ Here's a working example using the `mustache-array-template` working against an 
 </html>
 ```
 
+
+### Full Handlebars HTML Example
+
+To use the client side template, you will need to include htmx, the extension, and the rendering engine.
+Here is an example of this setup for Mustache using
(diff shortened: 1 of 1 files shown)


alexpetros:
This looks reasonable, can you rebase against the most recent version?

Telroshan:
Hey, we moved extensions to a [separate repository](https://github.com/bigskysoftware/htmx-extensions) alongside the htmx 2 release.
If you are still interested, please open a PR against' the extension's [README in that new repo](https://github.com/bigskysoftware/htmx-extensions/blob/main/src/client-side-templates/README.md)

PR #2229, opened 2024-01-21
Improve active-search example by not using chrome-only event "search"

### LINKED ISSUE #1569: Active-search example partially incompatible with Firefox

### LINKED ISSUE #2126: [BUG] Active Search Demo Initial State Is Blank
I was checking out htmx as I have heard great things but when I went to this demo:

https://htmx.org/examples/active-search/

I saw that the initial list is filtered out already. This isn't exactly what I'd expect from a framework I want to use.

MikeMoolenaar:
Fixes #1569
Fixes #2126

## Description
The active search example uses [the chrome-only event "search"](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/search_event) which is not supported in Safari and Firefox. This PR solves that  by using the `keyup[key=='Enter']` trigger. 

I also fixed #2126 so it won't show an empty table on load, it now shows all data initially. I think this better reflects how a real table search would work, but please let me know if you don't agree with that change and I'll remove the `load` trigger.

Corresponding issue: #1569 #2126

## Testing
Tested on latest the latest version of Firefox and Chromium.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/examples/active-search.md
+++ b/www/content/examples/active-search.md
@@ -41,7 +41,7 @@ we add the `changed` modifier to the trigger to ensure we don't send new queries
 value of the input (e.g. they hit an arrow key, or pasted the same value).
 
 We can use multiple triggers by separating them with a comma, this way we add 2 more triggers:
-- `keyup[key=='Enter']` triggers once enter is pressed. We use [event filters](/attributes/hx-trigger#standard-event-filters) here to check for [key property in the KeyboardEvent object](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key).
+- `keyup[key=='Enter']` triggers once enter is pressed. We use [event filters](/attributes/hx-trigger#standard-event-filters) here to check for [the key property in the KeyboardEvent object](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key).
 - `load` in order to show all results initially on load.
 
 Finally, we show an indicator when the search is in flight with the `hx-indicator` attribute.
--- a/www/content/examples/active-search.md
+++ b/www/content/examples/active-search.md
@@ -5,20 +5,20 @@ template = "demo.html"
 
 This example actively searches a contacts database as the user enters text.
 
-We start with a search input and an empty table:
+We start with a text input and an empty table:
 
 ```html
-<h3> 
-  Search Contacts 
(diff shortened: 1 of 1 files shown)


Telroshan:
The `search` input type is supported [even by IE11](https://caniuse.com/?search=input%20search), why change it to a simple text input here?

The `search` event on the other hand [is indeed non standard](https://caniuse.com/?search=search%20event) and you're right about adding a keyup trigger, though shouldn't we use _both_ events here, as even though it's non standard, it provides a better UX on the browsers that support it?

MikeMoolenaar:
> The `search` input type is supported [even by IE11](https://caniuse.com/?search=input%20search), why change it to a simple text input here?

You are right, I mistakenly though it wasn't supported on other browsers than Chrome. I've restored the input to "search".

> houldn't we use _both_ events here, as even though it's non standard, it provides a better UX on the browsers that support it?

Honestly, I don't know about the advantage of adding the search event here. I've tested on Chrome without the search event and the requests still gets sent when clearing the input by pressing `Escape`.  also, the input also works on Android by pressing the search button on the keybaord (it just send the keyup event with Enter code).

alexpetros:
> shouldn't we use both events here, as even though it's non standard, it provides a better UX on the browsers that support it?

The purpose of this page is to showcase how htmx works for simple, common use-cases, so using browser-specific features is almost categorically out of scope. It's much more important that the examples work everywhere; when people identify that they don't, we always fix it.

PR #2237, opened 2024-01-23
use wildcard (*) to match "revealed" trigger when scrolling

### LINKED ISSUE #2233: "revealed" event conflicts with other events, causing the revealed event to stop working
When using `hx-trigger="revealed, click"` or any other event or custom event, **AND** the element is below the fold (so that you have to scroll into it) the "revealed" event won't be triggered, but the other event still works.

I've found this bug trying to trigger the request with both "revealed" and some custom event from the server 
```HTML
<span hx-get="{% url 'contacts_count' %}" hx-trigger="updatecount from:body, revealed">
        loading...
</span>
```

I've created a minimal example that reproduces this problem:

https://jsfiddle.net/danielnieto/a6gwjL5u/

I can confirm that if you remove the "click" event from the hx-trigger attribute in the example I posted then "revealed" starts working again, also, when the page loads if the element is already on screen then "revealed" works as expected even if it's combined with other events.

Try for yourself, you can edit the JSFiddle link above and click on the play button top left

**Steps to reproduce:**

1. Create HTML with lots of content, so you have to scroll down to get to the bottom
2. Use `hx-trigger="revealed, click"` on an element at the bottom of the page

danielnieto:
## Description
Fixes #2233, using wildcard to match "revealed" trigger when collecting elements to execute `maybeReveal` on while scrolling allows using more events in the same hx-trigger attribute (i.e. click, mouseenter, custom events, etc)

A minimal example to reproduce the bug is included in the original issue.

Corresponding issue: #2233 

## Testing
The only way to test this is to have an element with `hx-trigger="revealed, anyOtherEvent"` below the fold (so that the initial `maybeReveal` function would not trigger the request)

I've updated `scroll-test-eventHandler.html` manual test to include 3 cases (at the very bottom), when the reveal event appears first, when it's last and when it's in the middle. All of them working after applying the fix.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1578,7 +1578,7 @@ return (function () {
                 setInterval(function() {
                     if (windowIsScrolling) {
                         windowIsScrolling = false;
-                        forEach(getDocument().querySelectorAll("[hx-trigger='revealed'],[data-hx-trigger='revealed']"), function (elt) {
+                        forEach(getDocument().querySelectorAll("[hx-trigger*='revealed'],[data-hx-trigger*='revealed']"), function (elt) {
                             maybeReveal(elt);
                         })
                     }
--- a/test/manual/scroll-test-eventHandler.html
+++ b/test/manual/scroll-test-eventHandler.html
@@ -102,6 +102,9 @@
 <div class="panel" hx-get="/more_content" hx-trigger="revealed"></div>
 <div class="panel" hx-get="/more_content" hx-trigger="revealed"></div>
 <div class="panel" hx-get="/more_content" hx-trigger="revealed"></div>
+<div class="panel" hx-get="/more_content" hx-trigger="revealed, click"></div>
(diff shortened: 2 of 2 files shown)


carducci (commented):
Grateful for this pull request. I've been running this patched version in production for a couple of months. Many thanks to the authors and contributors in this community!

1cg:
hey @danielnieto would like to merge this but it has a big conflict w/ the 2.x codebase for some reason (even though the change is so simple).  Can you redo it against a fresh copy of the 2.0 code and I'll merge it?

danielnieto:
@1cg oh! I just realized it's conflicting because you already fixed this bug 6 months ago in commit 8c61e71 by making the exact same changes. Therefore, you can close this PR and mark [#2233](https://github.com/bigskysoftware/htmx/issues/2233) as resolved.

PR #2241, opened 2024-01-24
Add trailing slash to manual tests link

danielnieto:
## Description

When trying to contribute, [docs suggest](https://github.com/bigskysoftware/htmx/?tab=readme-ov-file#hacking-guide) to use `npx serve` and then go to http://0.0.0.0:3000/test/ to see all tests, in this file there's a link to the manual tests, but if you have followed the steps on the guide you will realize that the links for the individual manual tests are not working, all files will throw a 404, the first time I saw this I didn't know if I had something misconfigured, then I realized that the links were relative and since the link to "manual" didn't have a trailing slash, all individual manual tests were pointed to the root of tests folder (i.e. `http://localhost:3000/test/aborting-request` instead of `http://localhost:3000/test/manual/aborting-request`)

The "correct" way to fix this would be to add a `serve.json` as stated here: https://github.com/vercel/serve/issues/628, but not sure about adding a new config file and also, on my first test that approach worked for the test files links but broke relative image paths so I would need to review all tests and fix all affected. (not to mention that we need to make sure that paths works when open the files "locally" without `npx serve`)

Instead, I'm submitting this PR which only adds a trailing slash to the manual tests link on `http://0.0.0.0:3000/test/` this way a newcomer would not get immediately confused when trying to run and explore tests. 

I've made sure that this approach also works when opening the files "locally".

**Screenshot of the issue:**
image

Corresponding issue:

## Testing
NA

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -24,7 +24,7 @@
 </ul>
 
 <h2>Manual Tests</h2>
-<a href="manual">Here</a>
+<a href="manual/">Here</a>
 
 <h2>Mocha Test Suite</h2>
 <a href="index.html">[ALL]</a>


alexpetros (approved):
Yup, I know exactly what you're talking about and this fixes it—thank you! If there were a way to specify this config with an environment variable we could put in the package.json and that would be fine, but you're right that we don't want to add another config.

Issue #2148, opened 2024-01-02
Broken TypeScript types in 1.9.10


nwalters512:
There are two issues with the TypeScript types that prevent `tsc` from typechecking a project that uses HTMX:

- The syntax at the following line is invalid (the `?` should come before the `:`: https://github.com/bigskysoftware/htmx/blob/145627a5792520b10f35e86eea23ce5c388e0fb0/src/htmx.d.ts#L432
- The type referenced on the following line (`HtmxTriggerSpecification`) does not exist: https://github.com/bigskysoftware/htmx/blob/145627a5792520b10f35e86eea23ce5c388e0fb0/src/htmx.d.ts#L438

The first is trivial to fix, but I'm not sure what the type for `HtmxTriggerSpecification` is supposed to be. If someone can point me in the right direction, I can open a PR.

It'd be great to have CI check these types to prevent future regressions!

scrhartley:
Looking at this recent PR: #2107
> Hey all, we're about to do a pretty significant refactor and unfortunately now isn't the right time to introduce this. The htmx typing story is at the moment incomplete—if you want to update the API types you're more than welcome to.

and

> Haha the guidance is, for the moment, please hold off on TS changes.
> 
> EDIT: unless they're to the type declarations in the API interface.

Given that's the current stance, I question whether the `triggerSpecsCache` property should have been added (which was done as part of #2094).

nwalters512:
I imagine they shouldn't have been added. What's the best course of action here: remove `triggerSpecsCache` and fix the syntax error in `ignoreTitle`?

nwalters512:
I opened https://github.com/bigskysoftware/htmx/pull/2261 to fix this.

1cg:
no longer relevant

PR #2261, opened 2024-01-30
Fix TypeScript types

nwalters512:
## Description

This PR fixes the TypeScript types in `src/htmx.d.ts`. There were two separate issues:

- Line 432 contained invalid syntax.
- `triggerSpecsCache` referenced a `HtmxTriggerSpecification` type the doesn't exist

Given the guidance in https://github.com/bigskysoftware/htmx/pull/2107#issuecomment-1866499625, I opted to remove the types for `triggerSpecCache` entirely.

Interestingly, there are other parts of the code (JSDoc comments) that reference `HtmxTriggerSpecification`:

- https://github.com/bigskysoftware/htmx/blob/7f3d114269d423ac8e77c64e988d6f4cf60325b1/src/htmx.js#L1273
- https://github.com/bigskysoftware/htmx/blob/7f3d114269d423ac8e77c64e988d6f4cf60325b1/src/htmx.js#L1364

This leads me to believe that this type did exist at one point, but I couldn't immediately find it in the Git history. It seems like a type that should exist though!

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/2148

## Testing

I installed these types into my project and verified that `tsc` could now successfully parse the types.

## Checklist

* [x] I have targeted this PR against the correct branch (`master` for website changes, `dev` for source changes)

### DIFF
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -429,13 +429,7 @@ export interface HtmxConfig {
      * If set to true htmx will not update the title of the document when a title tag is found in new content
      * @default false 
      */
-    ignoreTitle:? boolean;
-    /**
-     * The cache to store evaluated trigger specifications into.
-     * You may define a simple object to use a never-clearing cache, or implement your own system using a [proxy object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
-     * @default null
-     */
-    triggerSpecsCache?: {[trigger: string]: HtmxTriggerSpecification[]};
+    ignoreTitle?: boolean;
 }
 
 export type HtmxEvent = "htmx:abort"


alexpetros:
We're doing a big re-work for this for htmx 2.0!

nwalters512:
@alexpetros thats exciting! Is v2 going to be released shortly? The latest version of `htmx` is unusable in TypeScript projects because of this, so if v2 is not right around the corner, I'd very much like to see this fixed in a v1 release.

alexpetros:
The beta should be out in a couple weeks. 

I'm not sure I agree that htmx is unsuable in TypeScript projects, as there are a lot of ways to enable htmx in TS projects that don't require typed DOM attributes or APIs (with templates, for instance). Broadly speaking, htmx enables you to have minimal logic (and therefore minimal scripting) in the frontend, drastically diminishing the need for TypeScript's guarantees to extend past the backend. Many people use statically-typed backends with htmx happily.

I'm sure it's a problem for whatever your rendering setup is, and I apologize for that. I know this response is sort of annoying, but I don't want anyone to find this comment and think it applies to TypeScript projects generally.

nwalters512:
Sorry, to be clear, I don't mean that it's unusable because of any lack of type declarations or such. I don't actually rely on any of that, and I don't have a "rendering setup" that necessitates any type information from HTMX at all.

I have an `import 'htmx.org/dist/htmx.js'` statement I use to get HTMX into my client-side bundle; that bundle root is a `.ts` file. Because of the invalid syntax in the HTMX type declarations, TypeScript errors when typechecking my bundle, even though I don't actually rely on any of HTMX's types:

```
../../node_modules/htmx.org/dist/htmx.d.ts:432:17 - error TS17020: '?' at the start of a type is not valid TypeScript syntax. Did you mean to write 'boolean | null | undefined'?

432     ignoreTitle:? boolean;
                    ~~~~~~~~~

../../node_modules/htmx.org/dist/htmx.d.ts:438:45 - error TS2304: Cannot find name 'HtmxTriggerSpecification'.

438     triggerSpecsCache?: {[trigger: string]: HtmxTriggerSpecification[]};
                                                ~~~~~~~~~~~~~~~~~~~~~~~~


Found 2 errors.
```

This isn't silenceable with a `// @ts-ignore` or a `// @ts-expect-error` on the `import ...` statement. So when I say "HTMX is unusable in a TypeScript project", I mean "importing HTMX into a `.ts` file produces a compiler error that I cannot silence or work around".

alexpetros:
Much appreciated on the explanation, I had indeed misunderstood what you were saying. I do think we're going to need to do a 1.9.11, so I'll reopen this and we'll get it looked at.

alexpetros (approved):

PR #2280, opened 2024-02-05
Add htmx security essay

alexpetros:
New essay about security

### DIFF
--- a/www/content/essays/web-security-basics-with-htmx.md
+++ b/www/content/essays/web-security-basics-with-htmx.md
@@ -1,0 +1,321 @@
++++
+title = "Web Security Basics (with htmx)"
+date = 2024-02-06
+[taxonomies]
+author = ["Alexander Petros"]
+tag = ["posts"]
++++
+
+As htmx has gotten more popular, it's reached communities who have never written server-generated HTML before. Dynamic HTML templating was, and still is, the standard way to use many popular web frameworks—like Rails, Django, and Spring—but it is a novel concept for those coming from Single-Page Application (SPA) frameworks—like React and Svelte—where the prevalence of JSX means you never write HTML directly.
+
+But have no fear! Writing web applications with HTML templates is a slightly different security model, but it's no harder than securing a JSX-based application, and in some ways it's a lot easier.
+
+## Who is guide this for?
+
+These are web security basics with htmx, but they're (mostly) not htmx-specific—these concepts are important to know if you're putting *any* dynamic, user-generated content on the web.
+
+For this guide, you should already have a basic grasp of the semantics of the web, and be familiar with how to write a backend server (in any language). For instance, you should know not to create `GET` routes that can alter the backend state. We also assume that you're not doing anything super fancy, like making a website that hosts other people's websites. If you're doing anything like that, the security concepts you need to be aware of far exceed the scope of this guide.
+
+We make these simplifying assumptions in order to target the widest possible audience, without including distracting information—obviously this can't catch everyone. No security guide is perfectly comprehensive. If you feel there's a mistake, or an obvious gotcha that we should have mentioned, please reach out and we'll update it.
+
+## The Golden Rules
+
+Follow these four simple rules, and you'll be following the client security best practices:
+
+1. Only call routes you control
+2. Always use an auto-escaping template engine
+3. Only serve user-generated content inside HTML tags
+4. If you have authentication cookies, set them with `Secure`, `HttpOnly`, and `SameSite=Lax`
+
+In the following section, I'll discuss what each of these rules does, and what kinds of attack they protect against. The vast majority of htmx users—those using htmx to build a website that allows users to login, view some data, and update that data—should never have any reason to break them.
+
+Later on I will discuss how to break some of these rules. Many use applications can be built under these constraints, but if you do need more advanced behavior, you'll be doing so with the full knowledge that you're increasing the conceptual burden of securing your application. And you'll have learned a lot about web security in the process.
+
+## Understanding the Rules
+
+### Only call routes you control
+
+This is the most basic one, and the most important: **do not call untrusted routes with htmx.**
+
+In practice, this means you should only use relative URLs. This is fine:
+
+```html
+<button hx-get="/events">Search events</button>
+```
+
+But this is not:
+
+```html
+<button hx-get="https://google.com/search?q=events">Search events</button>
+```
+
+The reason for this is simple: htmx inserts the response from that route directly into the user's page. If the response has a malicious `<script>` inside it, that script can steal the user's data. When you don't control the route, you cannot guarantee that whoever does control the route won't add a malicious script.
+
+Fortunately, this is a very easy rule to follow. Hypermedia APIs (i.e. HTML) are [specific to the layout of your application](https://htmx.org/essays/hypermedia-apis-vs-data-apis/), so there is almost never any reason you'd *want* to insert someone else's HTML into your page. All you have to do is make sure you only call your own routes.
+
+Though it's not quite as popular these days, a common SPA pattern was to separate the frontend and backend into different repositories, and sometimes even to serve them from different URLs. This would require using absolute URLs in the frontend, and often, [disabling CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
+
+With htmx (and, to be fair, modern React with NextJS) this is an anti-pattern. Instead, you simply serve your HTML frontend from the same server (or at least the same domain) as your backend, and everything else falls into place: you can use relative URLs, you'll never have trouble with CORS, and you'll never call anyone else's backend.
+
+htmx executes HTML; HTML is code; never execute untrusted code.
+
+### Always use an auto-escaping template engine
+
+When you send HTML to the user, all dynamic content must be escaped. Use a template engine to construct your responses, and make sure that auto-escaping is on.
+
+Fortunately, all template engines support escaping HTML, and most of them enable it by default (Jinja is a notable exception). Below are just a few examples.
+
+| Language | Template Engine | Escapes HTML by default? |
+| ---- | ---- | ---- |
+| JavaScript | Nunjucks | Yes |
+| JavaScript | EJS | Yes, with `<%= %>` |
+| JavaScript | Handlebars | Yes, with `{{ }}` |
+| Python | Jinja | **No** |
+| PHP | Blade | Yes |
+| Go | html/template | Yes |
+| Java | Thymeleaf | Yes |
+| Rust | Tera | Yes |
+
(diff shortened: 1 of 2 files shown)


megatux:
Just skimming the text, looks great to me.
Sorry if my English is wrong here, is "Many use applications can be built" right? or "useR applications" was the intended string?
Also, I'd like to add Ruby ERB example to the templates escaping table:
```markdown
| Ruby | ERB | Yes, with `<%= %>` |
```

alexpetros:
I meant "useful" actually, thanks for the catch, and for reading! I'll add the Ruby templates, I'm kinda surprised I didn't have them already

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+<a {{ user.attribute }}></a>
+
+<!-- User-defined attribute VALUES are sometimes okay, it depends -->
+<a class="{{ user.class }}"></a>
+
+<!-- Escaped content is always safe inside HTML tags (this is fine) -->
+<a>{{ user.name }}</a>
+```
+
+CSS, JavaScript, and HTML attributes are ["dangerous contexts,"](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#dangerous-contexts) places where it's not safe to allow arbitrary user input, even if it's escaped. Escaping will protect you from some vulnerabilities here, but not all of them; the vulnerabilities are varied enough that it's safest to default to not doing *any* of these.
```
yawaramin, line 170: It would be better to keep the comma on the right of the link markup.

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+Set-Cookie: token=asd8234nsdfp982; Secure; HttpOnly; SameSite=Lax
+
+[HTML content]
+```
+
+So what do the options do?
+
+The first one, `Secure`, ensures that the browser will not send the cookie over an insecure HTTP connection, only a secure HTTPS connection. Sensitive info, like a user's login token, should *never* be sent over an insecure connection.
+
+The second option, `HttpOnly`, means that the browser will not expose to the cookie to JavaScript, ever (i.e. it won't be in [`document.cookie`](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie)). Even if someone is able to insert a malicious script, like in the `evilwebsite.com` example above, that malicious script cannot access the user's cookie or send it to `evilwebsite.com`. The browser will only attach the cookie when the request is made to the website the cookie came from.
```
yawaramin, line 218: `s/the browser will not expose to the cookie/the browser will not expose the cookie/`

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+
+You can fix this in a couple of ways. The simplest, and safest, trick is to let users customize these values, but don't let them define the literal text. In the image example, you might upload the image to your own server (or S3 bucket, or the like), generate the link yourself, and then include it, unescaped. In nunjucks, you use the [safe](https://mozilla.github.io/nunjucks/templating.html#safe) function:
+
+```html
+<img src="{{ user.fav_img_s3_url | safe }}">
+```
+
+Yes, you're including unescaped content, but it's a link that you generated, so you know it's safe.
+
+You can handle custom CSS in the same way. Rather than let your users specific the color directly, give them some limited choices, and set the choices based on their input.
```
yawaramin, line 273: `s/specific/specify/`

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+{% if user.favorite_color === 'red' %}
+h1 { color: 'red'; }
+{% else %}
+h1 { color: 'blue'; }
+{% endif %}
+```
+
+In that example, the user can set `favorite_color` to whatever they like, but it's never going to be anything but red or blue. A less trivial example might ensure that only properly-formatted hex codes can be entered, using a regex. You get the idea.
+
+Depending on what kind of customization you're supporting, securing it might be relatively easy, or quite difficult. Some attributes are ["safe sinks,"](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#safe-sinks) which means that their values will never be interpreted as code; these are quite easy to secure. If you're going to include dynamic input in ["dangerous contexts,"](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#dangerous-contexts) you need to research *what* is dangerous about those contexts, and ensure that that kind of input won't make it into the document.
```
yawaramin, line 285: Same as before, recommend moving the commas to the right of the link markup ie `[...](),`

yawaramin (commented):
Overall looks good. Few comments.

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+
+Instead, you simply serve your HTML frontend from the same server (or at least the same domain) as your backend, and everything else falls into place: you can use relative URLs, you'll never have trouble with CORS, and you'll never call anyone else's backend.
+
+htmx executes HTML; HTML is code; never execute untrusted code.
+
+### Always use an auto-escaping template engine
+
+When you send HTML to the user, all dynamic content must be escaped. Use a template engine to construct your responses, and make sure that auto-escaping is on.
+
+Fortunately, all template engines support escaping HTML, and most of them enable it by default (Jinja is a notable exception). Below are just a few examples.
```
merwok, line 66: Maybe it should be noted that any web framework that uses Jinja2 will enable auto-escaping for its HTML templates: https://jinja.palletsprojects.com/en/3.0.x/api/#autoescaping

alexpetros: This still says it's disabled by default I'm afraid.

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+| Go | html/template | Yes |
+| Java | Thymeleaf | Yes |
+| Rust | Tera | Yes |
+
+The kind of vulnerability this prevents is often called a Cross-Site Scripting (XSS) attack, a term that is [broadly used](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#introduction) to mean the injection of any unexpected content into your webpage. Typically, an attacker uses your APIs to store malicious code in your database, which you then serve to your other users who request that info.
+
+For example, let's say you're building a dating site, and it lets users share a little bio about themselves. You'd render that bio like this, with `{{ user.bio }}` being the bio stored in the database:
+
+```html
+<p class=bio>
```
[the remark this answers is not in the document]
alexpetros: These are all legal and fine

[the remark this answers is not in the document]
yawaramin: True, but (1) it's inconsistent because other places in the article are using quotes and (2) this is a security-focused article and specifically recommends using quotes, so why not get into the habit everywhere?

www/content/essays/web-security-basics-with-htmx.md
alexpetros: I went back and forth on this, but I'm sticking with the American English style style.

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+
+But this is not:
+
+```html
+<button hx-get="https://google.com/search?q=events">Search events</button>
+```
+
+The reason for this is simple: htmx inserts the response from that route directly into the user's page. If the response has a malicious `<script>` inside it, that script can steal the user's data. When you don't control the route, you cannot guarantee that whoever does control the route won't add a malicious script.
+
+Fortunately, this is a very easy rule to follow. Hypermedia APIs (i.e. HTML) are [specific to the layout of your application](https://htmx.org/essays/hypermedia-apis-vs-data-apis/), so there is almost never any reason you'd *want* to insert someone else's HTML into your page. All you have to do is make sure you only call your own routes.
```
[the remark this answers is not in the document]
alexpetros: Good idea.

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Who is guide this for?
+
+These are web security basics with htmx, but they're (mostly) not htmx-specific—these concepts are important to know if you're putting *any* dynamic, user-generated content on the web.
+
+For this guide, you should already have a basic grasp of the semantics of the web, and be familiar with how to write a backend server (in any language). For instance, you should know not to create `GET` routes that can alter the backend state. We also assume that you're not doing anything super fancy, like making a website that hosts other people's websites. If you're doing anything like that, the security concepts you need to be aware of far exceed the scope of this guide.
+
+We make these simplifying assumptions in order to target the widest possible audience, without including distracting information—obviously this can't catch everyone. No security guide is perfectly comprehensive. If you feel there's a mistake, or an obvious gotcha that we should have mentioned, please reach out and we'll update it.
+
+## The Golden Rules
```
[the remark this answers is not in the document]
alexpetros: Yup, I'll fix that

www/content/essays/web-security-basics-with-htmx.md
```diff
@@ -0,0 +1,322 @@
...
+
+Follow these four simple rules, and you'll be following the client security best practices:
+
+1. Only call routes you control
+2. Always use an auto-escaping template engine
+3. Only serve user-generated content inside HTML tags
+4. If you have authentication cookies, set them with `Secure`, `HttpOnly`, and `SameSite=Lax`
```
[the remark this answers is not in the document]
alexpetros: Good question... I think authentication is appropriate here though because it's really about credentials.

alexpetros:
Thanks so much for the review @yawaramin :)

www/content/essays/web-security-basics-with-htmx.md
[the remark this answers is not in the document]
alexpetros: You're right, I was just going to come back and say that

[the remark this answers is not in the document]
alexpetros: I stand by my no-quoting style!!! But it's not appropriate for this article specifically

www/content/essays/web-security-basics-with-htmx.md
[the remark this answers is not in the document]
alexpetros: Went with sentence-case, except on this one specifically where I'm using Golden Rules like a proper noun.

www/content/essays/web-security-basics-with-htmx.md
merwok: Well yes, but that’s not the point I was making 🙂 

The jinja2 library does not escape HTML by default, because it’s a generic template system for text files. But any web framework will configure jinja2 to escape in its HTML templates.

alexpetros: Oh, I see your point! Thanks for clarifying. I will update the text to clarify.

PR #2310, opened 2024-02-13
Sync configuration options from reference

bencroker:
This syncs the configuration options from the reference page to the main docs page, adding new and updating some modified values.


alexpetros (approved):
Sweet, thank you Ben!

PR #2319, opened 2024-02-15
Fixes issue 1537 - OOB does not escape query selector

FraserChapman:
## Description

This change alters how the oobSwap function builds the query selector string. Rather than simply prefixing it with a `#` it uses the  `'[id="raw_id"]'` format. Where `raw_id` is the value obtained from `getRawAttribute(oobElement, "id")`

Ideally this could be done more cleanly with template literals (backtick strings) - but as I understand it you want to support IE11. 

This fix allows the call querySelectorAll to return the nodelist of all elements with the id "raw_id" - without getting "Failed to execute 'querySelectorAll' on 'Document': 'raw_id' is not a valid selector, and whilst not limiting the selection to the first matching id in the document.

Corresponding issue:
https://github.com/bigskysoftware/htmx/issues/1537

## Testing

I ran the htmx.js test suite Version: 1.9.10 and got passes: 679 failures: 0

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -799,7 +799,7 @@ return (function () {
          * @returns
          */
         function oobSwap(oobValue, oobElement, settleInfo) {
-            var selector = "#" + getRawAttribute(oobElement, "id");
+            var selector = '[id="' + getRawAttribute(oobElement, "id") + '"]';
             var swapStyle = "outerHTML";
             if (oobValue === "true") {
                 // do nothing


src/htmx.js
```diff
@@ -799,7 +799,7 @@ return (function () {
          * @returns
          */
         function oobSwap(oobValue, oobElement, settleInfo) {
-            var selector = "#" + getRawAttribute(oobElement, "id");
+            var selector = '[id="' + getRawAttribute(oobElement, "id") + '"]';
```
MAW016, line 802: Wouldn't be nicer to use CSS.escape()?

```javasctipt
let selector = '#' + CSS.escape(getRawAttribute(oobElement, 'id'))
```

FraserChapman: Yes it would be nicer...but htmx maintains support for Internet Explorer 11 so `CSS.escape` (and any other ES6) isn't supported.

MAW016: hmm I am curious how this fix will be propagated to v2... Contribution rules says PR only into dev.

FraserChapman: Err, this is into bigskysoftware:dev

Telroshan:
Hey @FraserChapman , following up late on this (sorry about that!)
If you are still interested in this change, could you resolve conflicts?

> as I understood it htmx v1 maintains support for Internet Explorer 11 and CSS.escape isn't supported.

Since then, we released htmx 2 which dropped IE11 support, so feel free to use that method if you want 😄

src/htmx.js
MAW016: Sorry I was not sure if dev branch is for v2 or v1... I quite sure that is for v2 so IE support is dropped... @FraserChapman

MAW016:
Hi @FraserChapman, could you, please, finish this PR? It just needs to resolve conflicts, or should I take over?

FraserChapman:
@MAW016 absolutely happy to, my understanding it was it wasn't wanted/required based on the previous messages. I raised it way back on version 1. Apologies for the misunderstanding.

MichaelWest22:
a newer version of this PR just merged so we can close this duplicate now

FraserChapman:
Cool, just a shame this sat for over a year then project moved on and it was out of date  :/

FraserChapman:
@MichaelWest22 Many thanks for picking this up, sorry it seemed to slip through the cracks for so long!

PR #2322, opened 2024-02-15
Solve wrapping issues for small content area.

gnat:
Before / After

image

### DIFF
--- a/www/content/migration-guide-hotwire-turbo.md
+++ b/www/content/migration-guide-hotwire-turbo.md
@@ -24,8 +24,8 @@ addEventListener("htmx:afterOnLoad", (event) => {
 })
 ```
 * Or, [hx-on](@/attributes/hx-on.md) may be used:
-  * `hx-on:submit='event.target.querySelectorAll("button").forEach(node => { node.disabled = true })'`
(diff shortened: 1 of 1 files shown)


www/content/migration-guide-hotwire-turbo.md
```diff
@@ -50,13 +50,13 @@ document.body.addEventListener('htmx:configRequest', (event) => {
 ```
 
 * Or, use an [hx-trigger](@/attributes/hx-trigger.md) condition: `hx-trigger="submit[action(target)]"`
-  * Does not currently resolve async calls such as `fetch`. See: https://github.com/bigskysoftware/htmx/issues/912
```
alexpetros, line 53: I think the "async" to "await" here is not correct. Yes, `await` is the statement but they're "asynchronous functions".

gnat:
@Atmos4 What's the problem? This is a revision of a previous commit.

Anyways, done. The narrow content area is a challenge to layout content for.

Atmos4:
No problem here 🙂
Same comment with the await, the rest LGTM!

alexpetros (approved):

Issue #1597, opened 2023-07-17
Rework "find" extended CSS selector


Telroshan:
I must admit I'm not very satisfied with the current `find` keyword
`closest` seems obvious because it's going to retrieve the closeST element, so obviously only one.

But when it comes to `find`, maybe is it because I'm not a native English speaker, but I "feel" it differently depending on the context
- For hx-target, I think of it as retrieving a single element _(which is correct)_
- But for hx-trigger's `from` keyword, and `hx-include`, I would expect it to return all elements matching the selector _(which is not correct, current implementation only returns the first matching child)_
Examples for that second point:
```html
<div hx-get="/search" hx-trigger="keyup from:find input[type='search']">
    ...
    <label>Some filter <input type="search"></label>
    <label>Some other filter <input type="search"></label>
</div>
```
Here, I would instinctively expect the div to listen for `keyup` on both inputs, or more if there were more of them.
Not only does it only retrieve the first matching child, the only way to accomplish such behaviour as of today would be using a standard CSS selector such as `#result input[type='search']`, which requires adding an ID / class to be able to select on that parent div specifically.

I find it confusing because the word `find` itself doesn't indicate if it's going to be one, or many elements

So, would it make sense to add a new keyword to make that clearer ? Which would btw add the feature of listening on multiple elements using `from` as mentioned above
Examples that come to mind:
- `find-first` and `find-all` 
```html
<div hx-trigger="from:find-first input"></div>
<div hx-trigger="from:find-all input"></div
```
- `find-child` and `find-children`
```html
<div hx-trigger="from:find-child input"></div>
<div hx-trigger="from:find-children input"></div
```
- `closest-child` and `children`
```html
<div hx-trigger="from:closest-child input"></div>
<div hx-trigger="from:children input"></div
```

Feel free to suggest other ideas, especially given that my English isn't that great

For retro-compatibility, I would then suggest to keep `find` being an alias of whatever new keyword that would replace it, but deprecate it / undocument it

Let me know

xhaggi:
Calling `find` in `querySelectorAllExt` seems wrong to me, because the function name `querySelectorAllExt` suggests something else. For `next` and `previous`, `querySelectorAll` is also used. What if we use `findAll` as a quick fix here? This way you could use `:first-child` if you are only interested in the first element e.g. `from:find input:first-child`.

xhaggi:
[...]

PR #2326, opened 2024-02-16
support this: extended css selector

gbourant:
hx-trigger now supports `this` as extended css selector, so we can do something like this:

```html
<form method="GET" hx-trigger="input delay:500ms changed from:(this input), submit">
        <input type="email" name="email">
        <input name="password">
</form>
```

Also we might want to do `thisFirst` (or something like that) which will use `.querySelector`.

If this is about to be added i will the tests later.


Telroshan:
What is the difference / added value compared to the already existing [`find` keyword](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L619)?

gbourant:
The form has two input fields, if i use `from:(find input)` the find method returns only the first. That is happening because the find method uses `.querySelector` and not `.querySelectorAll`

In order to get the all input fields i use `from:(this input)` which uses `.querySelectorAll`.

Telroshan:
Oh ok indeed I see, I also complained back then about the inability to target multiple elements with the `from` modifier (#1597)
- I must admit I'm not fond of the `this` keyword in this case, doesn't sound that intuitive to me.
- Please note that we're holding off feature PRs for now as we work on htmx 2, which we're hoping to release in beta by early March
- As per the contribution guidelines, we expect people to respect the PR template, including referencing an associated issue (we don't take unsollicited feature PRs - note that this relates directly to #1597 so you don't need to create a new issue don't worry, but I'll have to have 1cg look at the issue first before a PR can be accepted on the matter), and tests (so don't hesitate adding tests to your PRs directly to speed up the back-and-forth)

Until then, as a workaround, note that you can put an ID or use any specific-enough selector to target elements within your form, such as
```html
<form id="whatever" method="GET" hx-trigger="input delay:500ms changed from:(#whatever input), submit">
        <input type="email" name="email">
        <input name="password">
</form>
```

As said above, we aren't going to move forward with feature PRs until we get htmx 2.0 out so don't expect movement from our side on this before at least early April I'd say

1cg:
maybe should support a `from any:` modifier?

gbourant:
Alright, I've added a test for this feature. If this is going to be added i will update the docs accordingly.

> maybe should support a `from any:` modifier?

I think that `from any:` means 'any element of that selector in this page' which does not describe exactly what we want to do.

Telroshan:
Hey, are you still interested in this change @gbourant ?
If so, please update & resolve conflicts
I have to say I'm still not fond of the `this` keyword, maybe `find-all` would be more in tune with the current `find` that only returns 1 result? Just an idea out loud though, don't take this as a change request

bennadel:
Just want to say +1 to using `this` to scope a CSS selector!

alexpetros:
It looks like this PR is broken, but I think the possibility of adding a selector like this is still open. Closing this for inactivity, but feel free to open an issue if this is interesting to you and link to it here. Extending the extended css selectors is a little tricky but we are open to it.

PR #2336, opened 2024-02-19
JSdoc in htmx.js + generated TypeScript definition

Telroshan:
# Highlights
- Added JSDoc basically everywhere
- Used parent types whenever possible _(i.e. Node instead of Element, Element instead of HTMLElement and so on)_
- Typescript types definition is now generated, using `npm run type-declarations`
- Inlined public API's functions within the `htmx` object itself as `tsc` would otherwise generate invalid TypeScript code
- I put all custom `typedef`s in the global scope so that `tsc` properly exports them, and at the bottom of the file as I thought it would be pretty annoying to have to scroll past 150~ lines of JSDoc to start reading the code
- To avoid making _too many_ changes to the code, I added a few `// @ts-ignore` in places where code _is_ valid but tsc is too strict

Tests pass, also ran the tests in the extensions repo, all clear

# Going further
If this is accepted, I would recommend going further with the following:
- Call `tsc` in the `npm t` script to have the pipeline check the code is correctly typed before merging PRs
- Make a custom JS script _(called with node in the npm command along tsc)_ to copy the JSDoc descriptions and default values hints _(as all JSDoc is lost along tsc's generation)_

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,14 +1,19 @@
 {
-    "compilerOptions": {
-        "target": "es5",
-        "noEmit": true,
-        "checkJs": true,
-        "jsx": "react",
-        "strict": false,
-        "lib": [ "dom" ]
-    },
-    "include": [
-        "./src/htmx.js"
-    ],
-    "verbose": true
+  "compilerOptions": {
+    "allowJs": true,
+    "declaration": true,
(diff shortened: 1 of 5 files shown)


1cg (approved):
lgtm

Issue #2308, opened 2024-02-12
hx-swap-oob should support all hx-swap attributes


aral:
`hx-swap-oob` should support all `hx-swap` atttributes (`swap`, `settle`, `delay`, etc.) but fails with "is not a valid selector" error (support for this is not present currently in hx-swap-oob in the main branch or in v2.0v2.0).

e.g.,

```html
<p hx-swap-oob='true settle:0.5s'>…</p>
```

Results in:

```
Uncaught DOMException: Document.querySelectorAll: '0.5s' is not a valid selector
```

This is an important missing use case for WebSocket workflows where you want to allow CSS transitions time to play between updating nodes.

Add this failing test to /test/attributes/hx-swap-oob.js to recreate:

```js
  it('works with a swap delay', function(done) {
    this.server.respondWith('GET', '/test', "<div id='d2' hx-swap-oob='innerHTML swap:10ms'>Clicked!</div>")
    var div = make("<div id='d1' hx-get='/test'></div>")
    var div2 = make("<div id='d2'></div>")
    div.click()
    this.server.respond()
    div2.innerText.should.equal('')
    setTimeout(function() {
      div2.innerText.should.equal('Clicked!')
      done()
    }, 30)
  })
```

As far as I can see at the moment, the main issue appears to be that `swap()` calls `oobSwap()` whereas each out-of-band swap should really be calling a generic `swap()` function with a reference to the element to affect that applies the full set of swap functionality. Then regular swaps would become a special case where the element in question is also the one that triggered the event.

The only other way I can see of implementing this is to repeat the functionality that’s already in `swap()` in `oobSwap()` which isn’t ideal by any means. 

## Discovery

I’ll update this section as I look into this issue further with an eye towards implementing this functionality for hx-swap-oob in general and for the ws extension in particular (from a cursory glance, it looks like that requires work also to implement this).

- `getSwapSpecification()` in htmx.js only reads values defined in `hx-swap`. This is an easy fix:

  ```js
  function getSwapSpecification(elt, swapInfoOverride) {
    var swapInfo = swapInfoOverride ? swapInfoOverride : getClosestAttributeValue(elt, "hx-swap");
    if (!swapInfo) swapInfo = getClosestAttributeValue(elt, "hx-swap-oob");
      var swapSpec = {
        "swapStyle" : getInternalData(elt).boosted ? 'innerHTML' : htmx.config.defaultSwapStyle,
        //…
  ```

[...]

PR #2367, opened 2024-03-02
2.0 Swap API Normalization + OOB swap modifier support

jeffturcotte:
## Description

This is a bugfix refactor for v2.0 that does the following:

- Normalizes the new `swap()` api to handle all modifiers internally such as view transitions, swap delay, and settle delay. Previously these features were limited to the XHR handler and would not work via other contexts, such as SSE/WS.
- Brings OOB swaps closer to being "first class swaps" as support for swap modifiers has been added to them.

```
# OOB Swap Example
<div hx-swap-oob="true:.selector transition:true swap:3s settle:2s"></div>
```

IMO These are both pretty important to ensure that all swaps work the same, no matter the type or context from which they are called. There are a couple issues about this specifically and maybe many tangential ones.

I didn't open an issue myself or start a discussion beforehand, but hoping to use this PR as the forum to determine the best way forward on this. Thanks!

## Testing

Added new tests for hx-swap-oob modifiers, swap: and settle:

```
npm run test
```

## Checklist

* [ X ] I have read the contribution guidelines
* [ X ] FOR 2.0...  I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [ X ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [ X ] I ran the test suite locally (`npm run test`) and verified that it succeeded

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1449,12 +1449,12 @@ var htmx = (function() {
               maybeDelay(doSettle, swapSpec.settleDelay)
             }
 
-            doSwap = wrapWithTransition(doSwap, swapSpec)
+            doSwap = wrapWithTransition(doSwap, swapSpec, swapOptions)
 
             try {
               maybeDelay(doSwap, swapSpec.swapDelay)
             } catch (e) {
-              triggerErrorEvent(elt, 'htmx:oobSwapError', swapOptions.eventInfo)
+              triggerErrorEvent(swapOptions.contextElement, 'htmx:oobSwapError', swapOptions.eventInfo)
             }
           }
         }
@@ -1846,6 +1846,10 @@ var htmx = (function() {
     if (!swapOptions) {
       swapOptions = {}
     }
+    // default contextElement for logging
+    if (!swapOptions.contextElement) {
+      swapOptions.contextElement = document.body
+    }
 
     target = resolveTarget(target)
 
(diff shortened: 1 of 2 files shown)


src/htmx.js
```diff
@@ -1400,19 +1400,20 @@ var htmx = (function() {
...
-    } else if (oobValue.indexOf(':') > 0) {
-      swapStyle = oobValue.substr(0, oobValue.indexOf(':'))
-      selector = oobValue.substr(oobValue.indexOf(':') + 1, oobValue.length)
-    } else {
-      swapStyle = oobValue
+    let swapStyle = swapSpec.swapStyle
+
+    if (swapStyle.indexOf(':') > 0) {
+      selector = swapStyle.substr(swapStyle.indexOf(':') + 1, swapStyle.length)
+      swapStyle = swapStyle.substr(0, swapStyle.indexOf(':'))
```
zimmi, line 1413: Random drive by optimization / modernization:
```suggestion
    const colonIdx = swapStyle.indexOf(':')
    if (colonIdx > 0) {
      selector = swapStyle.substring(colonIdx + 1)
      swapStyle = swapStyle.substring(0, colonIdx)
```

alexpetros:
Hi! 

> I didn't open an issue myself or start a discussion beforehand, but hoping to use this PR as the forum to determine the best way forward on this. Thanks!

As noted in the contribution guidelines and your description, issues are the appropriate forum for proposing new behavior like this. Please open an issue with your proposed changes and we can discuss them. We will be triaging issues once 2.0 launched.

aral:
Discussion is happening in #2308.

PR #2376, opened 2024-03-04
More explicit explanation of htmx.config.getCacheBusterParam

lookbusy1344:
## Description

The documentation for `htmx.config.getCacheBusterParam` can be confusing. Colleagues and I expected an ephemeral timestamp or random number, not the target element id. I have tried to clarify its behaviour in the docs:

Original text is "defaults to false, if set to true htmx will include a cache-busting parameter in GET requests to avoid caching partial responses by the browser"

New text is "defaults to false, if set to true htmx will append the target element to the GET request in the format `org.htmx.cache-buster=targetElementId`"

For reference the behaviour is implemented here
https://github.com/bigskysoftware/htmx/blob/ba2c30b648d64be50a127ca4767eae41a4bc266f/src/htmx.js#L3204

Note the tests are well phrased in this regard
https://github.com/bigskysoftware/htmx/blob/ba2c30b648d64be50a127ca4767eae41a4bc266f/test/attributes/hx-get.js#L113


alexpetros (approved):
Great clarification, thank you!

PR #2391, opened 2024-03-10
WS.js fix for Safari on iOS 17.4

Renerick:
## Description

Seems like Safari on iOS 17.4 doesn't correctly set `event.target` from custom events dispatched from JS. Earlier same issue occurred in my htmx-signalr extension: https://github.com/Renerick/htmx-signalr/pull/11

Corresponding issue: https://discord.com/channels/725789699527933952/1216185723229704222

## Testing

I've ran the test suite and verified that test server still works. However, I don't have iOS 17.4 device at hand, so will have to ask the user to verify if it works.

## Checklist

  source changes)
* [x] This is either a **bugfix**, a documentation update, or a new feature that has been explicitly
  approved via an issue

### DIFF
--- a/src/ext/sse.js
+++ b/src/ext/sse.js
@@ -37,10 +37,11 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 		 */
 		onEvent: function(name, evt) {
 
+			var parent = evt.target || evt.detail.elt;
 			switch (name) {
 
 				case "htmx:beforeCleanupElement":
-					var internalData = api.getInternalData(evt.target)
+					var internalData = api.getInternalData(parent)
 					// Try to remove remove an EventSource when elements are removed
 					if (internalData.sseEventSource) {
 						internalData.sseEventSource.close();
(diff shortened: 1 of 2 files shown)


andryyy:
Why are they doing this again and again and again… 

Nice catch!

soyangel:
> By chance are you based in the EU? Because the biggest changes in 17.4 are likely to affect EU users in particular - despite the biggest changes being rescinded due to our advocacy work.
> 
> https://open-web-advocacy.org/blog/apple-backs-off-killing-web-apps/

Not OP but had the same bug but when using SSE. I'm in the EU and I'm wondering if this also happens for users outside the EU.

kxzk:
> > By chance are you based in the EU? Because the biggest changes in 17.4 are likely to affect EU users in particular - despite the biggest changes being rescinded due to our advocacy work.
> > https://open-web-advocacy.org/blog/apple-backs-off-killing-web-apps/
> 
> Not OP but had the same bug but when using SSE. I'm in the EU and I'm wondering if this also happens for users outside the EU.

In the US, happening for me as well.

Issue #2393, opened 2024-03-10
Gracefully close SSE connection initiated by server


roharvey:
There are use cases where the server should deliver a discrete number of events to the client (e.g. a progress bar). It is not intuitive how to stop this using the SSE extension. I propose this is formalized via a `sse-close` attribute, which identifies an event name that can trigger the client to close the eventSource. This is simpler and more intuitive than other solutions that involve swapping elements (see #1234 and #2149).

KianYang-Lee:
This should be closed?

Telroshan:
Indeed, solved in https://github.com/bigskysoftware/htmx-extensions/pull/13

PR #2394, opened 2024-03-10
SSE Close function

roharvey:
## Description
Added sse attribute `sse-close`, which identifies the event type to trigger an eventSource close.

Corresponding issue: #2393

## Testing
Added a test to ensure the connection is closed when the close event is sent.
image

## Checklist

  source changes)
  approved via an issue


package.json
```diff
@@ -28,7 +28,7 @@
     "test": "mocha-chrome test/index.html",
     "test-types": "tsc --project ./jsconfig.json",
     "dist": "cp -r src/* dist/ && npm run-script uglify && gzip -9 -k -f dist/htmx.min.js > dist/htmx.min.js.gz && exit",
-    "ws-tests": "cd ./test/ws-sse && node ./server.js",
+    "ws-tests": "cd ./test/ws-sse && node ./server.mjs",
```
roharvey, line 31: This was failing. I realized afterwards it wasn't necessary for testing (was running off master initially, which was quietly skipping the sse tests), kept it anyway.

roharvey:
> * [ ]  This is either a bugfix, a documentation update, or a new feature that has been explicitly
>   approved via an issue

New feature, however several related bug fixes were resolved via obscure workarounds.

> * [ ]  I ran the test suite locally (`npm run test`) and verified that it succeeded

Doesn't seem to work? I ran `npx serve` and verified the fixes instead.

src/ext/sse.js
```diff
@@ -196,6 +196,17 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
 				}
 			}
 		});
+
+		var closeAttribute = api.getAttributeValue(elt, "sse-close");
```
Renerick, line 200: Using this selector is not correct. `registerSSE` is not necessarily processes elements with `sse-connect` on the root.

The following test case, for example, is failing.

```js
    it('is closed after close message from server in nested content', function() {
        var div = make('<div hx-ext="sse"><div sse-connect="/foo" sse-close="close">' +
            '<div id="d1" sse-swap="e1"></div>' +
            '</div></div>');
        this.eventSource.wasClosed().should.equal(false);
        this.eventSource.sendEvent("close");
        this.eventSource.wasClosed().should.equal(true);
    })
```

I would either move this piece of code to `ensureEventSource` or use `queryAttributeOnThisOrChildren` and then look through the hierarchy upwards for the SSE element

src/ext/sse.js
```diff
@@ -196,6 +196,17 @@ This extension adds support for Server Sent Events to htmx.  See /www/extensions
...
+		if (closeAttribute) {
+			// close eventsource when this message is received
+			var source = api.getInternalData(elt).sseEventSource;
+			if (source != undefined) {
+				source.addEventListener(closeAttribute, function() {
+					source.close()
```
Renerick, line 206: Should also invoke `maybeCloseSSESource` like other event handlers

Renerick (changes requested):
I like the idea, but there were some nuances in the implementation that need addressing. 

Thank you for the contribution!

src/ext/sse.js
roharvey: I see, and thanks for the clear explanation of the problem. It does seem most appropriate inside `ensureEventSource` so I moved it to there, plus included the test you referenced above.

src/ext/sse.js
roharvey: I'm unsure about how to implement this change. The other functions that call this seem to all use swap, but since I'm not adding an additional element it appears that `maybeCloseSSESource()` finds the element in the body and thus doesn't close the source. Should I manufacture an extra element when setting up the event listener? Or perhaps I'm just passing the wrong element to search for in `maybeCloseSSESource()`?

Renerick: Thinking about it, it was a silly suggestion on my side. When this event handler is triggered, the event source will be closed anyway, there's no need to double check. Sorry 😅

roharvey:
> Doesn't seem to work? I ran `npx serve` and verified the fixes instead.

I don't know what I did before... but I was able to run this now.

Renerick:
Hi again, sorry for the late response. I was a bit busy over past week or so, hoping to find the time to review the updates this weekend. Thank you for your patience

Renerick (approved):
All right, current implementation looks good to me, it's clean and simple, so think we can give this the green light. One thing though is that since my last review we merged 2.0 into dev, so you'll need to retarget this PR against `v1` and also open matching PR in [bigskysoftware/htmx-extensions](https://github.com/bigskysoftware/htmx-extensions) for 2.0, which I'll also happily merge

roharvey:
@Renerick thanks, I've retargeted this and added https://github.com/bigskysoftware/htmx-extensions/pull/13

Renerick:
Status update, since the team is currently focused on 2.0, I think I'll leave this PR open until next 1.0 patch. I'll mark it ready for review so we can merge it as next 1.0 patch approaches

roharvey:
> I think I'll leave this PR open until next 1.0 patch

It might be simpler & clearer to just leave it as a 2.0 option only? The timing seems ideal.

Renerick:
Well, if you don't mind leaving it out of 1.0, I would agree that leaving it for 2.0 only is a decent option. I'm going to close the PR in that case, but feel free to ping or reopen if you got any questions!

Renerick:
@roharvey just wanted to let you know, I published sse-close as 2.1.0 on NPM https://www.npmjs.com/package/htmx-ext-sse/v/2.1.0

GrimalDev:
> Well, if you don't mind leaving it out of 1.0, I would agree that leaving it for 2.0 only is a decent option. I'm going to close the PR in that case, but feel free to ping or reopen if you got any questions!

https://github.com/bigskysoftware/htmx-extensions/issues/4#issuecomment-2109457084

Issue #2407, opened 2024-03-17
Out of date mustache example for client-side-templates extension in the docs.


airblast-dev:
The linked jsbin in the [docs](https://htmx.org/extensions/client-side-templates/#full-mustache-html-example) is quite out of date, so I will focus on the full example on the doc page itself. 

- The json from `https://jsonplaceholder.typicode.com/todos/1` returns a UserId, and not a UserID. So in the result we get a blank string. Simply making the last character lowercase fixes this issue.
- Clicking the button on the full example will keep in some artifacts from the template. We end up with a bunch of "{% raw %}", and "{% endraw %}". Simply removing those from the template fixes the issue.

PR #2409, opened 2024-03-17
Fix example for mustache in client-side-templates extension.

airblast-dev:
## Description
The unnecessary escape sequences were removed from the [mustache full example](https://htmx.org/extensions/client-side-templates/#full-mustache-html-example) as they broke the output.
The template key "UserID" was renamed to "UserId" as the [test API](https://jsonplaceholder.typicode.com/todos/1) uses "UserId". The old key name would result in a blank string as the keys wouldnt match.

Corresponding issue:
https://github.com/bigskysoftware/htmx/issues/2407

## Testing
I have tested the changed example in the docs. It is working correctly.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/extensions/client-side-templates.md
+++ b/www/content/extensions/client-side-templates.md
@@ -81,7 +81,7 @@ a [`<template>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/t
     <p id="content">Start</p>
 
     <template id="foo">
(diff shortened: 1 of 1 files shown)


airblast-dev:
There is one more thing that should be fixed. The jsbin example linked right under the example in the docs is really out of date, and also uses the incorrect key name "UserID" instead of "UserId". It would be really sweet if the jsbin example could be updated by one of the maintainers with the new example, or just remove the jsbin example altogether.

alexpetros (approved):
Thank you!

PR #2428, opened 2024-03-23
Switch to default export for htmx 2

alexpetros:


alexpetros:
Looks like we've got two broken tests... they happen on the previous branch for me too

Issue #718, opened 2021-12-15
htmx event listeners set on elements inside shadow DOM return right away


gotcha:
```
var eventListener = function (evt) {
    if (!bodyContains(elt)) {
        eltToListenOn.removeEventListener(triggerSpec.trigger, eventListener);
        return;
    }
```
If I understand properly the code above, this checks if a node is part of body to avoid to run event listener if it is not the case anymore.

However, if elements from inside a shadow DOM are processed with code like

```
htmx.process(elt.shadowRoot)
```

`bodyContains(elt)` at the begin of eventListener always returns `false` because `elt` is part of shadow DOM.

I propose the following fix to take care of shadow DOM nodes.

```         
function bodyContains(elt) {
    if (elt.getRootNode() instanceof ShadowRoot) {
        return getDocument().body.contains(elt.getRootNode().host);
    } else {
        return getDocument().body.contains(elt);
    }
}
```

Opinions ?

1cg:
Thank you for the change.  Can you write a test that exercises this functionality?

gotcha:
@cg Tests are why I started the CI PR. I'll do my best to add one.

PR #2434, opened 2024-03-26
bodyContains: Walk up the nested shadow root chain to check if the body contains the element

andrejota:
## Description

`bodyContains()` currently assumes that shadow root host element is always contained in the root body. That may not be the case and I experienced problem when nesting web components that rely on HTMX (e.g. `hx-get`. This change walks up the chain of shadow roots to find the topmost host element and verify for that one if it exists in the root body.

I did not find a corresponding issue, but I did find an older related issue: #718. I didn't file a new issue as it looks like a fairly straight-forward fix for what I think is just a bug.

## Testing
I ran this with my test site which uses nested web components ('open' shadow DOM) combined with HTMX. This change enabled `hx-get="..."` with `hx-target="click"` to work for me.

## Checklist
  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -764,13 +764,16 @@ var htmx = (function() {
    * @returns {boolean}
    */
   function bodyContains(elt) {
-    // IE Fix
-    const rootNode = elt.getRootNode && elt.getRootNode()
-    if (rootNode && rootNode instanceof window.ShadowRoot) {
-      return getDocument().body.contains(rootNode.host)
-    } else {
-      return getDocument().body.contains(elt)
+    // Shortcut for older browsers missing getRootNode() to avoid checking
+    // the condition on every iteration.
+    if (!elt.getRootNode) {
+      return getDocument().body.contains(elt);
     }
+    // Escape from shadow root.
+    while (elt.getRootNode() instanceof window.ShadowRoot) {
+      elt = elt.getRootNode().host;
+    }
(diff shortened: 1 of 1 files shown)


jackielii:
I'm having this issue in 2.0.0-beta4. I have a simple post button:

```html
<button
	hx-post={ formatUrl("/workflows/%s/runs", props.WorkflowName) }
	hx-target="body"
	hx-push-url="true"
>Run</button>
```

This returns the entire html, thus replacing "body". In version 1.9, there's no error, but in 2.0.0-beta4. I get "htmx:swapError"

image

Telroshan:
Hey, following up late on this @andrejota, could you fix your code's formatting (`npm run format`) to pass the CI? Thanks!

Ideally, adding a test case would greatly help to identify & ensure the issue is indeed fixed and doesn't get broken again in the future

andrejota:
> Hey, following up late on this @andrejota, could you fix your code's formatting (`npm run format`) to pass the CI? Thanks!
> 
> Ideally, adding a test case would greatly help to identify & ensure the issue is indeed fixed and doesn't get broken again in the future

Yup, happy to do it. On it.

Telroshan:
Hey @andrejota , we had a recent PR that addressed the nested shadow roots issue, see #3034 (released alongside htmx 2.0.4)
Could you confirm if that solves the issue here?

1cg:
fixed independently, thank you!

PR #2439, opened 2024-03-27
Fix: htmx-request class removal for parallel requests

tarunbhardwaj:
## Description
This patch addresses a concurrency issue where the same indicator, when used for multiple parallel requests, could result in a negative request count. This issue prevented the proper removal of the `htmx-request` class upon completion of all requests.

The fix ensures accurate tracking of active requests.

## Testing
Test case for multiple requests with same indicator is already covered, but reproducing this race condition in test is hard.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3182,14 +3182,14 @@ var htmx = (function() {
   function removeRequestIndicators(indicators, disabled) {
     forEach(indicators, function(ic) {
       const internalData = getInternalData(ic)
-      internalData.requestCount = (internalData.requestCount || 0) - 1
+      internalData.requestCount = Math.max((internalData.requestCount || 0) - 1, 0)
       if (internalData.requestCount === 0) {
         ic.classList.remove.call(ic.classList, htmx.config.requestClass)
       }
     })
     forEach(disabled, function(disabledElement) {
       const internalData = getInternalData(disabledElement)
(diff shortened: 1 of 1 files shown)


1cg:
is it possible to add a test for this fix?

tarunbhardwaj:
> is it possible to add a test for this fix?

There is already a test, but not for the race condition. Any suggestion on how to reproduce this race condition in test?

Telroshan:
@tarunbhardwaj In tests, we manually trigger the server answer, which means you _should_ be able to run multiple client actions and only then simulate the server response.
Could that work in your case? if so, you'd want to reproduce the issue with the current lib's state and confirm the test reproduces the bug, then ensure your PR fixes it

See this example of the current setup _(picked a random test there)_:
https://github.com/bigskysoftware/htmx/blob/c6a89b315b6f964f6d69fa53252c2eb3e63aa8c8/test/core/ajax.js#L15-L18

nikalexis:
Is this related to #2860?
Although #2860 is merged, I can't see the problem fixed in version 2.0.3.

Telroshan:
Hey @nikalexis,
If you're still seeing the issue and are interested in contributing, feel free to open a PR with a test case as described above for that race condition
I'll be closing this PR as it looks very similar to #2860 that was merged, so it's likely that if the issue persists, it lies elsewhere than what this PR addresses

PR #2458, opened 2024-04-03
Additional note for `hx-select` in documentation

devinMcArthur:
Explain how to set hx-select behavior to the default, useful when behavior is inherited by a parent.

## Description
Spent a good hour trying to reset an `hx-select` attribute that was being inherited from a parent element. I knew the inheritance was causing the issue as that is mentioned here, but was unsure how to properly deal with it's implications, I hope this note will help others deal with this moving forward.

If you feel inclined, feel free to adjust my wording. 

Cheers.

## Testing
N/A

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/attributes/hx-select.md
+++ b/www/content/attributes/hx-select.md
@@ -21,3 +21,4 @@ which will replace the entire button in the DOM.
 ## Notes
 
 * `hx-select` is inherited and can be placed on a parent element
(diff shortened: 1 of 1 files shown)


devinMcArthur:
After further review, `hx-select="*"` does not work as expected. It takes all elements from the response and places them at the top level of the tree, removing all parent child relationships.

PR #2532, opened 2024-05-05
make bodyContains() handle possibly nested shadow root's

spez:
## Description
make bodyContains() handle the case where elt might be inside nested shadow dom's

## Testing
Tested manually with and without nested shadow dom's

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -764,13 +764,16 @@ var htmx = (function() {
    * @returns {boolean}
    */
   function bodyContains(elt) {
-    // IE Fix
-    const rootNode = elt.getRootNode && elt.getRootNode()
-    if (rootNode && rootNode instanceof window.ShadowRoot) {
-      return getDocument().body.contains(rootNode.host)
-    } else {
-      return getDocument().body.contains(elt)
+    // if elt is inside a possibly-nested shadow root, we need to check the host
+    while (true) {
+      const rootNode = elt.getRootNode && elt.getRootNode()
+      if (rootNode && rootNode instanceof window.ShadowRoot) {
+        elt = rootNode.host
+      } else {
+        break
+      }
     }
+    return getDocument().body.contains(elt)
   }
 
   /**


1cg:
no opinion on this @kgscialdone do you have one?

Definitely needs tests if it is to be accepted though.

kgscialdone:
@innovation-stack I'm unclear why solving this with an extension would be preferable to simply solving it, even if the solution is a little gross. Could you elaborate on why?

innovation-stack:
@kgscialdone - For performance reasons. 
In my case, I am building a query-builder web component, which can be contain deeply nested components. 
Here's an illustration:

deeply-nested-wcs

**Few notes:**
1. Card component can be deeply nested, upto any level of hierarchy 
2. In my query-builder, any deeply nested child component can refer its root via `builder` property. 
3. As per solution proposed in PR, calling `bodyContains` on any deeply nested child web component have to iterate over all parent web components, which will incur performance. `bodyContains` is being called multiple times during render and API calls.

**Proposed solution:**
1. Using `body-contains-elem` extension, I can completely bypass all parent node traversals and simply use the `builder` property to refer to root component.

kgscialdone:
@innovation-stack I see where you're coming from, thank you for the explanation. However, I don't think that relegating a fix for this to extensions is the right approach from a logistical point of view.

That said, on further inspection of the use cases of `bodyContains`, it may be worth looking into whether it actually needs to be "does the `<body>` contain this element", or if all use cases would be fine also including elements in the `<head>`. I can't determine that offhand at the moment, so will need to inquire of @1cg for more information, but if it's the case that it doesn't matter whether it's in the `<body>` or the `<head>` in all use cases, the whole function could be replaced with the following and avoid the whole issue:
```js
function documentContains(elt) {
  return elt.getRootNode({ composed: true }) === getDocument()
}
```

innovation-stack:
@kgscialdone : Thank you for the explanation. I agree with your solution. Its elegant.
Let me know what's the final verdict on the approach.

PR #2544, opened 2024-05-13
Docs: Mention all extended selector keywords in `hx-disabled-elt` docs

imbolc:
I've also updated the example to disable inner form fields only


1cg:
agree the docs could be expanded but please keep the reference in to the `closest` method and mention that all the other keywords are followed by CSS selectors

imbolc:
> agree the docs could be expanded but please keep the reference in to the `closest` method and mention that all the other keywords are followed by CSS selectors

Sure, copied the attribute value options from `hx-target` instead of linking it

PR #2583, opened 2024-06-01
Fix typo in hx-ws.md

JennaHuff:
## Description
Fixed typo


1cg:
hx-ws is removed in 2.0 so i'm closing this out

new docs are here: https://github.com/bigskysoftware/htmx-extensions/blob/main/src/ws/README.md

PR #2585, opened 2024-06-02
Allow non-id selectors in `hx-select-oob` to fix #2561

### LINKED ISSUE #2561: Weird behaviour with `hx-select-oob` and non-id selectors
The issue in short, is that `hx-select-oob` only supports id selectors, and when non id selectors are passed in, it causes really strange and unintiutive behavior. This behavior is not documented anywhere and can cause really hard to debug bugs.

The offending code is below
```javascript
const oobSelectValue = oobSelectValues[i].split(':', 2)
var id = oobSelectValue[0].trim(); // Input
if (id.indexOf("#") === 0) {
    id = id.substring(1);
}
var oobValue = oobSelectValue[1] || "true";
var oobElement = fragment.querySelector("#" + id); // Output
```
[code in master](https://github.com/bigskysoftware/htmx/blob/4bc7c9710e99369a8c659f9b865e75866fd7a446/src/htmx.js#L867-L873)
[code in dev](https://github.com/bigskysoftware/htmx/blob/601d16c5ef2e94decc0d4189864e986a652c4bb3/src/htmx.js#L1838-L1844) 

Let's try some values:
- `#alert` produces `#alert`. As expected
- `alert-box` produces `#alert-box`. You'd expect it to produce an element by name(in this case, a Web Component) but it instead gets an element by id. The documentation suggests all types of selectors are accepted. But it is still acceptable if the documentation is updated
- `aside div` produces `#aside div`. WTF 
- `div[id="alert"]` produces `#div[id="alert"]`. Now we are in insanity territory

Furthermore, Attempting to use `:` in a css selector also completely breaks the parsing because we are using `String.split`. Something like `#id:has(~ #something-after):beforeend` will parse as `#id` for element and `:has(~#something-after)` for swap style. `beforeend` is completely dropped.

This selector should select the element, and swap it by its id, *even if we don't use an id selector*. **It should act like `hx-swap-oob` without a selector**. Instead, it has this weird behavior of forcing all selectors to be id selectors.

There are a few ways to solve this issue
## 1. Change the code to follow the logical behavior
```diff
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1835,13 +1835,11 @@ var htmx = (function() {
       if (swapOptions.selectOOB) {
         const oobSelectValues = swapOptions.selectOOB.split(',')
         for (let i = 0; i < oobSelectValues.length; i++) {
-          const oobSelectValue = oobSelectValues[i].split(':', 2)
-          let id = oobSelectValue[0].trim()
-          if (id.indexOf('#') === 0) {
-            id = id.substring(1)
-          }
-          const oobValue = oobSelectValue[1] || 'true'
-          const oobElement = fragment.querySelector('#' + id)
+          const oobSelectValue = oobSelectValues[i]
+          // Support colon in css selectors
+          let colon = oobSelectValues[i].lastIndexOf(':')
+          const oobValue = oobSelectValue.substring(colon + 1) || 'true'
+          const oobElement = fragment.querySelector(oobSelectValue.substring(0, colon - 1))
           if (oobElement) {
             oobSwap(oobValue, oobElement, settleInfo)
           }
```
This is technically a breaking change. However, the feature in question is unintuitive, undocumented, and very likely to cause more bugs than its fix. Even the [tests](https://github.com/bigskysoftware/htmx/blob/4bc7c9710e99369a8c659f9b865e75866fd7a446/test/attributes/hx-select-oob.js) do not mention this behavior. Obviously we can never be certain no one is using this, there is likely someone who hasn't read the documentation and assumes the argument is always an id but I see that as an unlikely possibility.

It also makes the total code 3 lines shorter so I see this as a win-win

## 2. Keep the current behavior, but document it
This is the only solution that is non-breaking. However, even if it is documented, it feels like a footgun waiting to trigger for some beginner.

## 3. Partially fix the issue, by explicitly only allowing id selectors
```diff
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1835,13 +1835,15 @@ var htmx = (function() {
       if (swapOptions.selectOOB) {
         const oobSelectValues = swapOptions.selectOOB.split(',')
         for (let i = 0; i < oobSelectValues.length; i++) {
-          const oobSelectValue = oobSelectValues[i].split(':', 2)
-          let id = oobSelectValue[0].trim()
+          const oobSelectValue = oobSelectValues[i]
+          let colon = oobSelectValues[i].lastIndexOf(':')
+
+          var id = oobSelectValue.substring(0, colon - 1).trim()
           if (id.indexOf('#') === 0) {
             id = id.substring(1)
           }
-          const oobValue = oobSelectValue[1] || 'true'
-          const oobElement = fragment.querySelector('#' + id)
+          const oobValue = oobSelectValue.substring(colon + 1) || 'true'
+          const oobElement = fragment.querySelector('#' + CSS.escape(id))
           if (oobElement) {
             oobSwap(oobValue, oobElement, settleInfo)
           }
```
This is still technically a breaking change. However. Even though there is likelihood of people using id selectors without `#` and expecting it to work, *no one* is using stuff like `alert[data-enabled="true"]` and expecting it to become `#alert[data-enabled="true"]`. If anything, they'd expect it to match `<div id='alert[data-enabled="true"]'>`. No one will encounter any issue.

I'd ideally also like some kind of warning when I try to use complex CSS selectors inside `hx-select-oob` but that has been left an exercise to the readers

If this fix is chosen, the documentation will also need to be updated

## Why is this an issue rather than a PR?
There are multiple ways to solve this problem. And I want a discussion of how this should be solved before attempting to make a pull request.

cadrgtsecond:
## Description
This pull request fixes #2561. I retargeted it against `v1` as suggested on discord. I found this issue with the `hx-select-oob` attribute while trying to help a beginner on discord. This technically breaks an obscure behavior of adding back `#` to a selector if it does not begin with a `#`. This behavior is neither documented nor tested. In fact, the documentation suggests you can use this with any CSS selector, not just id selectors(though it does not give an example). 

> Each value in the comma separated list of values can specify any valid [hx-swap](https://htmx.org/attributes/hx-swap/) strategy by separating the selector and the swap strategy with a :.

As I have explained in the issue, this can cause really strange behavior when combined with non-id selectors by reinterpreting them as id selectors(and that too not even properly).

## Testing
I added tests for non-id selectors and tests for non-id selectors with colons. I also added missing tests for `hx-select-oob` with different swap styles

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -864,13 +864,12 @@ return (function () {
             if (oobSelects) {
                 var oobSelectValues = oobSelects.split(",");
                 for (var i = 0; i < oobSelectValues.length; i++) {
-                    var oobSelectValue = oobSelectValues[i].split(":", 2);
-                    var id = oobSelectValue[0].trim();
-                    if (id.indexOf("#") === 0) {
-                        id = id.substring(1);
-                    }
-                    var oobValue = oobSelectValue[1] || "true";
-                    var oobElement = fragment.querySelector("#" + id);
+                    const oobSelectValue = oobSelectValues[i]
+                    // Support colon in css selectors
+                    const colon = oobSelectValue.lastIndexOf(':')
+                    const split_at = colon == -1 ? oobSelectValue.length : colon
+                    const oobValue = oobSelectValue.substring(split_at + 1) || 'true'
+                    const oobElement = fragment.querySelector(oobSelectValue.substring(0, split_at))
                     if (oobElement) {
                         oobSwap(oobValue, oobElement, settleInfo);
                     }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -864,12 +864,12 @@ return (function () {
             if (oobSelects) {
                 var oobSelectValues = oobSelects.split(",");
                 for (var i = 0; i < oobSelectValues.length; i++) {
-                    const oobSelectValue = oobSelectValues[i]
+                    var oobSelectValue = oobSelectValues[i]
                     // Support colon in css selectors
-                    const colon = oobSelectValue.lastIndexOf(':')
-                    const split_at = colon == -1 ? oobSelectValue.length : colon
-                    const oobValue = oobSelectValue.substring(split_at + 1) || 'true'
-                    const oobElement = fragment.querySelector(oobSelectValue.substring(0, split_at))
+                    var colon = oobSelectValue.lastIndexOf(':')
+                    var split_at = colon == -1 ? oobSelectValue.length : colon
+                    var oobValue = oobSelectValue.substring(split_at + 1) || 'true'
+                    var oobElement = fragment.querySelector(oobSelectValue.substring(0, split_at))
                     if (oobElement) {
                         oobSwap(oobValue, oobElement, settleInfo);
                     }
--- a/test/attributes/hx-select-oob.js
+++ b/test/attributes/hx-select-oob.js
@@ -49,6 +49,38 @@ describe("hx-select-oob attribute", function () {
         div2.innerHTML.should.equal("");
     });
 
-
+  it('supports different swap styles', function()
+  {
+      this.server.respondWith('GET', '/test', "<div id='d1'>foo</div><ul id='d2'><li>baz</li></ul>")
+      var div = make('<div hx-get="/test" hx-select="#d1" hx-select-oob="#d2:beforeend"></div>')
+      make('<ul id="d2"><li>bar</li></ul>')
+      div.click()
+      this.server.respond()
+      div.innerHTML.should.equal('<div id="d1">foo</div>')
+      var list = byId('d2')
+      list.innerHTML.should.equal('<li>bar</li><li>baz</li>')
+  })
+  it('supports non-id selectors', function()
+  {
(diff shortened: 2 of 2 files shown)


src/htmx.js
```diff
@@ -864,13 +864,12 @@ return (function () {
...
-                        id = id.substring(1);
-                    }
-                    var oobValue = oobSelectValue[1] || "true";
-                    var oobElement = fragment.querySelector("#" + id);
+                    const oobSelectValue = oobSelectValues[i]
+                    // Support colon in css selectors
+                    const colon = oobSelectValue.lastIndexOf(':')
+                    const split_at = colon == -1 ? oobSelectValue.length : colon
+                    const oobValue = oobSelectValue.substring(split_at + 1) || 'true'
+                    const oobElement = fragment.querySelector(oobSelectValue.substring(0, split_at))
```
1cg, line 872: unfortunately in v1 we can't use const, since it needs to be IE compatible.  Can you revert to var?

cadrgtsecond:
Sorry I'm a bit late. I was a bit busy

Issue #1349, opened 2023-04-07
intersect trigger: add rootMargin option


bidoule:
Hello, 

The `rootMargin` property for the Intersection Observer API allows to trigger the event before the element intersect into the view. It leads to a smoother web experience because the element is inserted before the user scroll into, and so it doesn't appear in front of the user.

It seems simple to add this option in the same way the options `root` and `threshold` are treated:
https://github.com/bigskysoftware/htmx/blob/192df25b25060f41abdc72e93472fbc796f0db93/src/htmx.js#L1718

References:
https://developer.mozilla.org/fr/docs/Web/API/Intersection_Observer_API
https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin
https://caniuse.com/mdn-api_intersectionobserver_rootmargin

RogierdeRuijter:
Would love to see this

deltour-hugo:
I need this option too please 😉

deltour-hugo:
`js
if (triggerSpec.rootMargin) {
    observerOptions.rootMargin = triggerSpec.rootMargin;
}
`

MichaelWest22:
This is now included in the latest beta release of 4

PR #2593, opened 2024-06-05
feat(hx-trigger): add support for rootMargin on intersect

believer:
## Description

Add support for [`rootMargin`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin) in `intersect` triggers. This allows us to extend the size of the intersect area. I've selected to separate multiple values using underscores, similar to how [Tailwind](https://tailwindcss.com/docs/content#using-spaces-and-underscores) solves it.

Corresponding issue: #1349 

## Testing

I've tested it locally using one of my own projects. I didn't find any tests for the `intersect` modifiers, so I couldn't find anything to base new tests on.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2179,7 +2179,7 @@ var htmx = (function() {
               triggerSpec[token] = consumeUntil(tokens, WHITESPACE_OR_COMMA)
             } else if (token === 'rootMargin' && tokens[0] === ':') {
               tokens.shift()
-              triggerSpec[token] = consumeUntil(tokens, WHITESPACE).replace(/_/g, ' ')
+              triggerSpec[token] = consumeUntil(tokens, /,|(root)|(threshold)/).replace(/_/g, ' ')
             } else {
               triggerErrorEvent(elt, 'htmx:syntax:error', { token: tokens.shift() })
             }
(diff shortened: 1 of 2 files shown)


1cg:
i like everything about this except the underscores.  What if we consumed until EOF or the next token followed by a colon?  (this may be a horrible idea idk)

believer:
I wasn't too excited about the underscores either, but I didn't have a better idea at the time. Two other options could be:

- `/,|(root)|(threshold)/` to match comma or one of the other two intersect options
- `/[,rt]/` for comma and the starting character of the options

I think the first option is easier to understand though

believer:
@1cg any thoughts on the updated parsing proposal above?

believer:
I came to need this again, so I updated the API without the need for underscores. Now it works with all these variations:

- `intersect rootMargin:10%`
- `intersect rootMargin:10px 20px 30px 40px`
- `intersect rootMargin:10px 10px threshold:0.1`
- `intersect rootMargin:10px 10px threshold:0.1 root:#test`

gustafnk:
Hi! Given there might be many elements with the need for lazy loading scroll with the same margin value, it would be nice to support setting a default value for all elements instead of specifying it for each element (not sure exactly how though).

believer:
@1cg What do you think about the updated API? I found a need for this again and remembered I had this PR 😅

believer:
Could this be something to add in v4?

MichaelWest22:
refactoring onTrigger so added rootMargin as an option as it is only one line in v4

believer:
Ohh nice!

believer:
I've tested the latest beta (`4.0.0-beta4`) with the changes from https://github.com/bigskysoftware/htmx/pull/3805 and it works perfectly. Thanks for the addition @MichaelWest22!

Issue #2599, opened 2024-06-07
BUG: head-support race condition


stukennedy:
I am trying to lazy-load a JS script using the new head-support extension, but getting race conditions.

Example:
```html
<head>
  <script src="https://cdn.jsdelivr.net/npm/chart.js"></script>
</head>
<canvas id="myChart"></canvas>
<script>
  htmx.on('htmx:afterHeadMerge', function () {
    var ctx = document.getElementById('myChart').getContext('2d');
    new Chart(ctx, { ...config});
  });
</script>
```

There are 2 issues going on here:
1. the head-support script initiates on an htmx:afterSwap event, but this can happen before the script in the DOM has run, so the DOM script in some cases won't ever see the `htmx:afterHeadMerge` event.
2. even when the `htmx:afterHeadMerge` is caught there is no guarantee that the JS script has finished initialising

I can fix 2 in my own code ... using this
```js
htmx.on('htmx:afterHeadMerge', function () {
  (async () => {
    while (!window.Chart) {
      await new Promise((resolve) => setTimeout(resolve, 100));
    }
    drawChart();
  })();
});
```
It feels a bit hacky, but it works ... I can abstract that away into a `executeWhen(drawChart, Chart)` method or something

But 1 needs fixing in the head-support script.

I've figured that if we change `htmx:afterSwap` to `htmx:afterSettle` that means that the script has loaded, and that fixes this issue so I'm going to punt in a PR.

1cg:
fixed in 4.0

PR #2600, opened 2024-06-07
fix: head-support race condition #2599

stukennedy:
## Description
currently the head-support feature starts doing its merging based off an `htmx:afterSwap` event. This means that any script being loaded into the DOM from the htmxSwap that wants to run some initialisation code may run after the `htmx:afterHeadMerge` event has triggered, and never see it.

An example for this would be if you want to return a ChartJS from an endpoint along with the script to display the chart.

```html
    <head>
      <script src="https://cdn.jsdelivr.net/npm/chart.js"></script>
    </head>
    <canvas id="myChart"></canvas>
    <script>
      htmx.on('htmx:afterHeadMerge', function () {
        (async () => {
          while (!Chart) {
            await new Promise((resolve) => setTimeout(resolve, 100));
          }
          var ctx = document.getElementById('myChart').getContext('2d');
          var myChart = new Chart(ctx, { <CONFIG> });
        })();
      });
    </script>
```
The polling is necessary to check that the `Chart` class has been declared (i.e. the new head script has loaded)
(and yes this could do with a timeout so it doesn't try for ever, but this is the basic solution)
However, this code won't work every time if we have `htmx:afterSwap` in `head-support.js` as the trigger.

Changing `htmx:afterSwap` to `htmx:afterSettle` ensures the DOM has settled (i.e. our event listener has been registered) before the head swap functionality is performed and our code works as expected.

## Testing
This was tested with the above code by calling this code on a lazy-loaded hx-post.
The race condition seems to be gone.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/dist/ext/head-support.js
+++ b/dist/ext/head-support.js
@@ -3,144 +3,170 @@
 //
 // An extension to htmx 1.0 to add head tag merging.
 //==========================================================
-(function(){
+(function () {
+	if (htmx.version && !htmx.version.startsWith('1.')) {
+		console.warn(
+			'WARNING: You are using an htmx 1 extension with htmx ' +
+				htmx.version +
+				'.  It is recommended that you move to the version of this extension found on https://extensions.htmx.org'
+		);
+	}
 
-    if (htmx.version && !htmx.version.startsWith("1.")) {
-        console.warn("WARNING: You are using an htmx 1 extension with htmx " + htmx.version +
-            ".  It is recommended that you move to the version of this extension found on https://extensions.htmx.org")
-    }
+	var api = null;
 
-    var api = null;
+	function log() {
+		//console.log(arguments);
+	}
 
-    function log() {
-        //console.log(arguments);
-    }
+	function mergeHead(newContent, defaultMergeStrategy) {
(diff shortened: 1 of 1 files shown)


Telroshan:
Hello, please note that for htmx 2 (aka the `dev` branch here), we moved extensions to [a separate repository](https://github.com/bigskysoftware/htmx-extensions).
So you'll want to either retarget this PR to the `v1` branch, in the extension's [`src` directory](https://github.com/bigskysoftware/htmx/tree/v1/src/ext), or port this PR to the new extensions repo (you may want to check that htmx 2 + the v2 extension doesn't already solve this btw)

1cg:
Yep, please retarget against v1 and the new extensions repo.  Thank you!

1cg:
Hey stu checking in here, can you retarget the `v1` branch and port the change over to the new extensions repo?

stukennedy:
> Hey stu checking in here, can you retarget the `v1` branch and port the change over to the new extensions repo?

Apologies, totally forgot to see this through.

Opened a new PR on the other branch
https://github.com/bigskysoftware/htmx/pull/2704

Issue #2601, opened 2024-06-08
BUG: hx-target with special characters fails 


pragmatta:
If you have a hx-target that is an id that has special characters like the colon fails because they need to be escaped for `document.querySelectorAll`.

![image](https://github.com/bigskysoftware/htmx/assets/29479033/dc4738ee-f4be-4e6d-8644-673feabb920d)
![image](https://github.com/bigskysoftware/htmx/assets/29479033/f0dd5a40-d67b-44b1-b380-d6d611ce67c1)
![image](https://github.com/bigskysoftware/htmx/assets/29479033/98e47829-61ab-487d-98a0-ba852c540410)

Same happens on console and you can fix it by adding a `\\`.
![image](https://github.com/bigskysoftware/htmx/assets/29479033/a3ffdd23-1f72-4c7b-9f02-120fb8201acf)

pragmatta:
Created a fix and a pull request for this:
https://github.com/bigskysoftware/htmx/pull/2602/files

PR #2602, opened 2024-06-08
Fixed escaping on normalizeSelector [2601]

pragmatta:
## Description
Colons and dots are valid id characters but need to be escaped. This PR adds escaping to the normalizeSelector-function.

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/2601

## Testing
I've tested that adding this escaping locally to htmx.js fixes the issue. I've not tested others scenarios that use normalizeSelector, but all code paths are feeding to the querySelectorAll so seems solid.

## Checklist

  source changes)
  approved via an issue
All tests were green but there was one mention of an issue elsewhere in code that seemed unrelated
`  1) hx-indicator attribute
       multiple requests with same indicator are handled properly:`

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1129,7 +1129,7 @@ var htmx = (function() {
     if (startsWith(trimmedSelector, '<') && endsWith(trimmedSelector, '/>')) {
       return trimmedSelector.substring(1, trimmedSelector.length - 2)
     } else {
-      return trimmedSelector.replaceAll(':', '\\:').replaceAll('.', '\\.'); // Issue 2601: colons and dots are valid id characters but need to be escaped
+      return trimmedSelector.replace(/\:/g, '\\:').replace(/\./g, '\\.')
     }
   }
 
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1128,8 +1128,10 @@ var htmx = (function() {
     const trimmedSelector = selector.trim()
     if (startsWith(trimmedSelector, '<') && endsWith(trimmedSelector, '/>')) {
       return trimmedSelector.substring(1, trimmedSelector.length - 2)
-    } else {
+    } else if (startsWith(trimmedSelector, '#')) {
       return trimmedSelector.replace(/\:/g, '\\:').replace(/\./g, '\\.')
+    } else {
+      return trimmedSelector
     }
   }
 
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1129,7 +1129,7 @@ var htmx = (function() {
     if (startsWith(trimmedSelector, '<') && endsWith(trimmedSelector, '/>')) {
       return trimmedSelector.substring(1, trimmedSelector.length - 2)
     } else {
-      return trimmedSelector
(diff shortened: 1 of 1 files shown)


Telroshan:
Hey, seems like reverting your changes gets the tests to all pass but with it, this hx-indicator always fails. So this must be related in some way even though it may not look obvious at first glance.
It seems that you replace ALL dots by escaped dots, which could maybe mess up the standard CSS class selector ?
I see that the failing test uses the following selectors:
https://github.com/bigskysoftware/htmx/blob/d6afc5b8dbd7213037d0bc4213aa0b7b469bcd62/test/attributes/hx-indicator.js#L107-L109
Just an idea but it really looks related to those dots replacement.

I hope you'll understand that this has to be fixed for this PR to be considered!

pragmatta:
Now there are no failing tests (still says "3 pending", not sure what that means) and it fixes the original issue.

Telroshan:
Now that I think about it, 3 things come to my mind :
- I recall an issue mentioning [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static), and as this PR targets htmx 2, it should be safe to use now (we refused it for v1 as it's IE11 incompatible). It probably does the escaping you want here and maybe covers more edge cases ? (like `#` within the id selector)
- I feel like your current implementation would still conflict with selectors such as `#container .some-child-class` ; here the dot would be escaped when it shouldn't, right?
- Finally, to avoid any doubt / confusion, we would need to add tests for this, and check as many selectors syntaxes as possible, the basic ones as well as the intricate ones

pragmatta:
You are correct. You could solve that by first testing if # exists, splitting by space and then doing the startsWith-test and replace for each word separately. But even then it's a good question if this covers (enough of) the edge cases. How deep into CSS semantics you'd need to go on this..

pragmatta:
Unfortunately `CSS.escape` is not an automatic solution as it will escape # also and break things.

```
CSS.escape('#OINOApiRow-databases-1:1')
'\\#OINOApiRow-databases-1\\:1'
```
The correct place/context is really in my code in the place where I concatenate an id to the target because then you what needs escaping :/ 

Should I close the PR?

Telroshan:
That's a bummer, it seems intricate to solve at the htmx level 😕 
Maybe we should just leave it to the responsibility of the user to properly format/escape their selectors, either manually in the hx attributes or escaping themselve beforehand for their JS calls? When you think about it, you also have to handle it yourself as the developer using the function, when calling the standard `document.querySelector`, which htmx is using internally. Maybe we shouldn't even try to handle those from htmx itself after all... Let me know your thoughts on this!

pragmatta:
Yeah I don't this is a correct path. But the problem is kinda fundamental in HTMX so let's store some context in case someone tries to solve it in the future or lands here based on googling the error. The case where this happens for me is a HTMX-template that get's filled with database column values. BUT it depends on the context how that escaping needs to be done.
image

Now you want to keep your templates as data and code as code. I.e. while editing the template you'd need to differentiate the cases in a way that you don't add code to data (e.g. JS-eval stuff) or make your code template specific (somehow parse from the context of the replace how to escape). 

I think on the short term I'll switch from colon to something selector safe, it'll just be a problem over and over again. Also I think I'll split the `###`-tags to different variants based on the context. I can add more of those as different needs pop up.

Issue #251, opened 2020-12-02
Document mutual incompatibility of hx-{delete,get,patch,post,put}


adamchainz:
These attributes all change the HTTP method. They're presumably incompatible - but what does this do? Does it cause an error, is there a specificity order?

1cg:
I have no idea what that will do.  :/  Well, I imagine it will do all of them, but that's just a guess.  

We can update the docs to say so, for sure.

aburd:
So I did some testing on this and if multiple attributes are specified at the same time, one method will take precedence and only one request is launched, where hx-get has highest priority and hx-patch the lowest.

GET,POST,PUT,PATCH,DELETE => GET
POST,PUT,PATCH,DELETE => POST
PUT,PATCH,DELETE => PUT
PATCH,DELETE => DELETE

I can make a PR with some docs if necessary.

samfurr:
I believe the order is defined [here](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L126), where downstream functions eventually parse for the hx elements. @bencroker do we still want this noted in the docs? PR to follow.

PR #2606, opened 2024-06-09
Docs: Add note on attribute execution priority

samfurr:
## Description
Filing to resolve an older issue on the repo, feel free to reject this if it is believed not necessary.
Priority information identified [here](https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L126) in the code, which is handled downstream by the parsers etc.

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/251

## Testing

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/docs.md
+++ b/www/content/docs.md
@@ -202,6 +202,8 @@ This tells the browser:
 
 > When a user clicks on this div, issue a PUT request to the URL /messages and load the response into the div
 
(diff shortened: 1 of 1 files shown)


Telroshan:
My 2 cents: while I totally understand the curiosity of "what would happen if", I feel like this doesn't belong to the docs, as it's an unsupported feature and htmx shouldn't be used that way at all.
My opinion would be that we don't have to and we _shouldn't_ document such things as I'm afraid it would introduce more confusion to the reader than anyting else.

samfurr:
Agree, thank you @Telroshan . Feel free to reject/close the associated issue.

1cg:
agree, don't think this should be in the docs it will just cause confusion

PR #2647, opened 2024-06-19
Doc: Update oobBeforeSwap

amrojjeh:
## Description
Updated oobBeforeSwap since it's out of date for two reasons:
- `shouldSwap` always defaults to true
- `xhr` and `requestConfig` are not properties of `detail`
- `fragment` is not documented

Regarding the `shouldSwap`, I've run tests returning 300 status codes and they did not set the `shouldSwap` property to false, as was previously specified. I've also reviewed the [source code](https://github.com/bigskysoftware/htmx/blob/f9b3f883576c979b533ed657b8a7556c3b1982e4/src/htmx.js#L2933) with a debugger to ensure that the `shouldSwap` in fact never changes and is always true unless modified by the event listener.

## Checklist

  source changes)
  approved via an issue
* [ ] I ran the test suite locally (`npm run test`) and verified that it succeeded :cry:

### DIFF
--- a/www/content/events.md
+++ b/www/content/events.md
@@ -310,7 +310,7 @@ This event is triggered as part of an [out of band swap](@/docs.md#oob_swaps) an
 
 * `detail.elt` - the element that dispatched the request
 * `detail.fragment` - the response fragment
-* `detail.shouldSwap` - if the content will be swapped (defaults to `false` for non-200 response codes)
+* `detail.shouldSwap` - if the content will be swapped (defaults to `true`)
 * `detail.target` - the target of the swap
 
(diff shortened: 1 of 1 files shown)


bencroker:
I additionally removed `detail.elt` and matched the event details in `htmx:oobAfterSwap`.

https://github.com/bigskysoftware/htmx/blob/f9b3f883576c979b533ed657b8a7556c3b1982e4/src/htmx.js#L1443

amrojjeh:
The detail.elt is added by the `triggerEvent`. Shouldn't it be in the docs? @bencroker

bencroker:
You’re right, that was an oversight on my part. I’ve reverted that change in https://github.com/bigskysoftware/htmx/pull/2649.

Issue #2629, opened 2024-06-17
2.0.0 - Missing dist/htmx.d.ts


jacob-ebey:
`package.json` has the the types field defined as `dist/htmx.d.ts` but that file is not published to npm.

1cg:
Sorry, will be in the next release (early next week) @alexpetros is fixing!

jacob-ebey:
Please re-open. Type issues still exist due to use of a ".js" MJS file being pointed to by a CJS package.json. You have two options:
- rename the `dist/htmx.esm.js` to `dist/htmx.esm.mjs` and update the main file accordingly
- or add `"type": "module"` to the package.json to ".js" means MJS.

alexpetros:
I think I added `"type": "module"` to the `package.json` and it caused some issues a while back

alexpetros:
Regardless, that is a separate issue, so please open a new one.

PR #2653, opened 2024-06-19
Generate .d.ts file in build script

alexpetros:
## Description

Corresponding issue: #2629

## Testing
* Ran the dist script and verified that it output the types.

## Checklist

  source changes)
  approved via an issue


scripts/dist.sh
```diff
@@ -2,35 +2,38 @@
...
 
 # Clean the dist directory
 rm -rf dist/*
 
 # Regular IIFE script
 cp $HTMX_SRC dist/htmx.js
 
-# Minified script
+# Generate .d.ts file
+tsc dist/htmx.js --declaration --allowJs --emitDeclarationOnly --outDir dist
```
Telroshan, line 14: Would you be okay to use a configuration file here instead, to avoid hardcoding the tsc options in 2 different places?
So that the `src` .d.ts file is generated just like the `dist` one and we don't take the risk of forgetting to update the other if we were to tweak that tsc configuration

For example we could have `jsconfig_dist.json`
```json
{
  "extends": "./jsconfig.json",
  "compilerOptions": {
    "outFile": "dist/htmx.d.ts"
  },
  "include": [
    "./dist/htmx.js"
  ]
}
```

Using it with `tsc --project ./jsconfig_dist.json`

alexpetros: I don't want to add any more config files unless there is a very good reason to and I don't think this is one of them. None of these are options where forgetting to change them could cause subtle breakages. They simply specify the input and output—which are both highly unlikely to change and unlikely to cause subtle bugs if they do.

Telroshan: Ok, then you'll still want to update that tsc command in the dist script I think as it's likely to be missing some options of the current tsc config, see `jsconfig.json`:
https://github.com/bigskysoftware/htmx/blob/fa2dc6c02e96786d7daaae7c165daef229829250/jsconfig.json#L1-L19
More specifically, `target`, `checkJs` and `lib` might differ if not specified I guess?

alexpetros: Although we could eliminate the jsconfig file, and just run `npm run check-types` from the dist script, which seems reasonable to me

alexpetros: Resolved this by removing the jsconfig and putting it all into the CLI. Turns out you only need all these options while checking, not generating (I verified that they produce the same output).

Issue #2659, opened 2024-06-21
htmx.logAll is not a function


anentropic:
Using Vite...

I have:
```js
import htmx from 'htmx.org';
window.htmx = htmx;
```
in my `index.js`

at the bottom of my html I have:
```html
      <script>
        document.body.addEventListener('htmx:load', function (evt) {
          htmx.logAll();
        });
      </script>
```
and I get:
```
Uncaught TypeError: htmx.logAll is not a function
    at HTMLBodyElement.<anonymous> (...)
    at he (htmx__org.js?v=a64430c1:1658:21)
    at htmx__org.js?v=a64430c1:3038:11
```
so HTMX is initialised because the `htmx:load` event fired, and `window.htmx` is something, but apparently not a thing that has `logAll`

the docs here https://htmx.org/docs/#webpack say to do this `window.htmx = require('htmx.org');` in a separate file and then import that into `index.js`

but then I get `ReferenceError: require is not defined`

TBH I am lost in the confusing mess of JS build tools and module formats and just want some instructions that work

anentropic:
so if I log what `htmx` is in the event handler...

when imported as `import htmx from 'htmx.org';` then:
![Screenshot 2024-06-21 at 14 22 39](https://github.com/bigskysoftware/htmx/assets/147840/5873ae37-66cb-4bc5-abe7-5c6a94f7f978)

as `import * as htmx from 'htmx.org';`:
![Screenshot 2024-06-21 at 14 23 49](https://github.com/bigskysoftware/htmx/assets/147840/3ac17e7b-402c-4317-8580-96f3c8cf7b6b)

as `import { htmx } from 'htmx.org';`:
`undefined`

if I just `import 'htmx.org';` then VS Code inteliisense will prompt me with `htmx.logAll()` as if it would work, but at runtime it gives `ReferenceError: htmx is not defined`

anentropic:
ah..! here https://github.com/bigskysoftware/htmx/issues/2628

if I do this in `index.js`:
```js
import htmx from "htmx.org/dist/htmx.esm";
window.htmx = htmx;
```

then the body script works:
```html
      <script>
        document.body.addEventListener('htmx:load', function (evt) {
          htmx.logAll();
        });
      </script>
```

Telroshan:
Hey, sorry for the inconvenience, as 1cg stated in the issue you linked
> entirely possible that those instructions are wrong now, unfortunately, since no one on the core team that I'm aware of uses webpack

[...]

Issue #2657, opened 2024-06-20
2.0: TypeError: import_htmx.default.config is undefined when defining config programmatically


danjac:
This was working in **1.9x**:

```javascript
import htmx from "htmx.org";

// global HTMX configuration
// https://htmx.org/docs/#config

htmx.config.historyCacheSize = 0;
htmx.config.refreshOnHistoryMiss = false;
htmx.config.scrollBehavior = "smooth";
htmx.config.scrollIntoViewOnBoost = false;
htmx.config.useTemplateFragments = true;
```

In **2.0.0** this causes the following error:

**TypeError: import_htmx.default.config is undefined**

The workaround of course is to use the `<meta>` tag, but has this behavior changed in 2.0?

danjac:
I also tried the workaround described here for webpack: https://htmx.org/docs/#webpack however this did not work, same result (note: I'm using esbuild, not webpack).

restless:
Have you seen this: https://github.com/bigskysoftware/htmx/issues/2628?

TL;DR;
Try `import htmx from "htmx.org/dist/htmx.esm";`

danjac:
@restless yes, that appears to work, thanks. FYI if updating the docs for webpack, the solution also works for esbuild.

davidlesieur:
Using Parcel here. `import htmx from "htmx.org/dist/htmx.esm";` works in that case too.

1cg:
fixed in 2.0

Issue #2628, opened 2024-06-17
[2.0] Webpack installation, global HTMX properties undefined


Jesmaster:
I just tried updating to 2.0 in my Webpack project and it looks like issue bigskysoftware/htmx#1964 came back. Having the exact same problem as then.

Switching back to version 1.9.12 fixes the issue.

1cg:
may be related to https://github.com/bigskysoftware/htmx/issues/2629

not familiar w/ webpack does it use any of the javascript modules?  if so you might need to switch to the new module-specific files in the distribution

Jesmaster:
The webpack installation documentation hasn't been updated so I assumed the setup is the same. None of the other js libraries I use in webpack need to specifiy any special module types.

https://htmx.org/docs/#webpack

1cg:
entirely possible that those instructions are wrong now, unfortunately, since no one on the core team that I'm aware of uses webpack.  We will get the .d.ts file fixed in the next release and then we can try again, and then maybe try by referencing one of the module files (`/dist/htmx.esm.js` ?) and then update the docs if necessary.  Can you try referencing the esm file now just to see if it does anything different?

Jesmaster:
I figured out how to set it up correctly. I'll create a pull request to fix the documentation.

For anyone who needs it

```
window.htmx = require('htmx.org');
```

becomes

```
import htmx from "htmx.org/dist/htmx.esm";
window.htmx = htmx;
```

restless:
Thanks @Jesmaster you've helped me to resolve my issue.

I use Vite and I had to do a similar change there. In the vite.config.js file i had to replace `htmx.org` with the `htmx.org/dist/htmx.esm`:

```
    plugins: [
      inject({
        htmx: 'htmx.org/dist/htmx.esm',
      }),
    ],
```

This is related to: https://github.com/bigskysoftware/htmx/issues/1690

cdmistman:
fyi using htmx 2.0.2 `htmx.org/dist/htmx.esm` doesn't seem to exist (via `node_modules`). With Astro (which uses Vite) all I had to do in my `<head>` was:

```html
<script>
	import htmx from 'htmx.org';
	window.htmx = htmx;
</script>
```

and i confirmed presence in the browser by printing `window.htmx` in the js console

PR #2668, opened 2024-06-24
Update webpack and Vite istallation instructions for htmx 2.0

restless:
## Description

The webpack installation instructions are no longer valid for htmx 2.0. There are also no installation instructions for Vite. 
This is affecting multiple people, judging from the issue tracker.

Corresponding issue:
- https://github.com/bigskysoftware/htmx/issues/2628
- https://github.com/bigskysoftware/htmx/issues/2657
- https://github.com/bigskysoftware/htmx/issues/2659

## Testing
It's a documentation update and the only way to test it is to follow the new guidelines. These new guidelines are what works for me and other people (see issues listed above).

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/docs.md
+++ b/www/content/docs.md
@@ -162,7 +162,7 @@ import 'htmx.org';
 ```
 
 However, you will probably need the global `htmx` variable, which gives your code access to methods like
-`htmx.onLoad` and allow you to use npm-installed htmx plugins. You can use one of the solutions following solutions
+`htmx.onLoad` and allow you to use npm-installed htmx plugins. You can use one of the following solutions
 to achieve this.
 
 
--- a/www/content/docs.md
+++ b/www/content/docs.md
@@ -154,6 +154,12 @@ For npm-style build systems, you can install htmx via [npm](https://www.npmjs.co
 npm install htmx.org@2.0.4
 ```
 
+and for extensions:
+```sh
+npm install htmx-ext-debug@2.0.1
+```
+
+
 After installing, you’ll need to use appropriate tooling to use `node_modules/htmx.org/dist/htmx[.esm, .cjs, .amd].js` (or `.min.js`).
 
 In the simplest case all you need to do is to add this import to your entry point (like `index.js`)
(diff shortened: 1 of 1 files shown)


Darkproduct:
I think this should include

```js
import inject from '@rollup/plugin-inject'
```

Darkproduct:
Also see #1690

restless:
Thanks @Darkproduct I've rewritten the docs to include your suggestions. To summarize there are the following options:

1. simply import htmx - it'll work then but won't allow you to use plugins or `htmx.method`s due to lack of the `htmx` global variable
2. import htmx and manually assign it to the `window.htmx` global variable
3. in case of webpack inject `htmx` via plugin
4. in case of Vite inject `htmx` via plugin

1cg:
I am changing the `main` in `package.json` to be the esm file:

https://github.com/bigskysoftware/htmx/blob/52b357089db95cea13411999a6b7913f0f0bc39b/package.json#L23

Does this change these instructions at all?  (Sorry I know nothing about Webpack or Vite)

jackielii:
Is this possible in esbuild?

www/content/docs.md
```diff
@@ -152,36 +152,81 @@ For npm-style build systems, you can install htmx via [npm](https://www.npmjs.co
...
+to add the following to your [vite.config.js](https://vitejs.dev/guide/backend-integration.html):
 
 ```js
-window.htmx = require('htmx.org');
+import inject from '@rollup/plugin-inject';
+
+export default defineConfig({
+ build: {
+   //[...]
+   plugins: [
```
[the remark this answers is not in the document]
databyte: Correct, `plugins` is a sibling with `build` (same level)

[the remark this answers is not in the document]
restless: Thanks guys. Fixed!

restless:
> Is this possible in esbuild?
> 
> I had to use a work around like so: https://discord.com/channels/725789699527933952/1252928586206154753/1252933134396428348

I don't use esbuild directly, however `define` (https://esbuild.github.io/api/#define) looks promising. Have you tried it?

alexpetros:
This is going to be related too: https://github.com/bigskysoftware/htmx/issues/2629#issuecomment-2226447245

TeachMeHowToUse2FA (commented):
👏

www/content/docs.md
```diff
@@ -152,36 +152,79 @@ For npm-style build systems, you can install htmx via [npm](https://www.npmjs.co
...
-For example, you might bundle htmx with some extensions and project-specific code.
+After installing, you’ll need to use appropriate tooling to use `node_modules/htmx.org/dist/htmx[.esm, .cjs, .amd].js` (or `.min.js`).
 
-### Webpack
+In the simplest case all you need to do is to add this import to your entry point (like `index.js`)
+```js
+import 'htmx.org';
+```
+
+However, you will probably need the global `htmx` variable, which gives your code access to methods like
```
tchoutri, line 162: If I may offer a suggestion:

The use of "probably" is unclear to beginners; You give a very concrete example right after (the use of htmx plugins), but is this the only scenario?

www/content/docs.md
```diff
@@ -152,36 +152,79 @@ For npm-style build systems, you can install htmx via [npm](https://www.npmjs.co
...
+`htmx.onLoad` and allow you to use npm-installed htmx plugins. You can use one of the solutions following solutions
```
tchoutri, line 163:
```suggestion
`htmx.onLoad` and allow you to use npm-installed htmx plugins. You can use one of the following solutions
```

www/content/docs.md
```diff
@@ -152,36 +152,79 @@ For npm-style build systems, you can install htmx via [npm](https://www.npmjs.co
...
 
 ```js
-window.htmx = require('htmx.org');
+import inject from '@rollup/plugin-inject';
+
+export default defineConfig({
+  build: {...},
+  plugins: [
+    inject({
+      htmx: 'htmx.org',
```
plooney81, line 221: Wouldn't this be: `'htmx.org/dist/htmx.esm'`?

restless: @plooney81 As per @1cg comment here: https://github.com/bigskysoftware/htmx/pull/2668#issuecomment-2195557294 `htmx.org` uses `htmx.esm.js` by default now. 
Please see: https://github.com/bigskysoftware/htmx/blob/master/package.json#L23C3-L23C30

In other words `htmx.org` should work (and works for me) well.

databyte (commented):
LGTM. This mimics my setup.

www/content/docs.md
```diff
@@ -152,36 +152,79 @@ For npm-style build systems, you can install htmx via [npm](https://www.npmjs.co
...
 ```
 
-If you want to use the global `htmx` variable (recommended), you need to inject it to the window scope:
+Import this file in your entry point (like `index.js`):
+
+```js
+import './htmx.js';
+
+// now you can import extensions
+import 'htmx-ext-debug/debug.js';
```
marisst, line 182: You are describing a workaround which requires creating `htmx.js` file to inject `htmx` into the `window` scope, so that the extensions can use it. It has to be done in a separate file, because ESM imports are asynchronous.

I have created a [PR](https://github.com/bigskysoftware/htmx-extensions/pull/123) in the extensions repository, which eliminates the need of this workaround. With the new change extensions can be imported like this:
```js
import `htmx.org`;
import `htmx-ext-extension-name`; // Replace `extension-name` with the extension name
```

It relies on the ESM import mechanism supported by most bundlers, including Webpack, Vite and Rollup.

gitblit:
I have not yet had success getting HonoX+Vite+Htmx working with these instructions, but updated guidance is important. :+1:

1cg:
Hello all, does this PR need updating?  I agree on removing "probably" and being firmer in what we recommend.  Unfortunately, my JS build-fu is (intentionally) very weak.  Looking for guidance!

restless:
@1cg I've updated the PR. It now reflects changes implemented by @marisst in [#123](https://github.com/bigskysoftware/htmx-extensions/pull/123) and makes the docs much simpler. 

I tested it with my Vite based setup and it works well. I believe it'll work with other bundlers, like webpack, too as there are no extra steps (providing/injecting htmx) needed anymore.

1cg:
Closing as stale, thank you. If the four docs need an update here, please take a look: https://four.htmx.org/docs

Issue #1701, opened 2023-08-18
Empty class elements left after swap


dmcg:
Inspecting the page after a swap, htmx leaves an empty `class=“”` attribute on the target element. It’s benign, but means that the resulting page is not the same as the a page for the same server state rendered from scratch, which makes testing harder.

 You can see the problem in https://youtu.be/qYuSr__OiHk at 23:15

Telroshan:
Hey, that's because htmx adds temporary [classes](https://htmx.org/reference/#classes) such as `htmx-request` to your elements while the requests are in flight, to let you use some styling when needed, such as a [`hx-indicator`](https://htmx.org/attributes/hx-indicator/) to display say a loading spinner until the request is done.

Since those classes are added, it's then actually simply because of how the JS API works that the empty `class` attribute remains, it's not related to htmx itself.

You can observe that behaviour in [this JSFiddle](https://jsfiddle.net/c9vjuwLh/)
```html
<div id="test">Inspect me in F12</div>

<button type="button" onclick="test.classList.add('test')">Add class</button>

<button type="button" onclick="test.classList.remove('test')">Remove class (classList)</button>
<button type="button" onclick="test.className=''">Remove class (className)</button>
```

Using both `classList` and `className`, you'll see that after removing the class, the `class` attribute doesn't go away, and remains here, empty

dmcg:
I can see how it happens, and if it can’t be fixed then so be it, but it seems a shame that testing is made more complicated than it could be?

duanemoody:
The JSFiddle shows the code leaving behind `class=""` instead of a blank `class` attribute for both RemoveClass buttons, so something different's happening here.

For htmx, the code temporarily creating a `class` attribute for the purpose of handling a trigger could check whether that attribute already exists on the element, and if not set a temporary element like `data-cleanup="class"` for htmx to check for, then have`element.classList.length == 0 && element.removeAttribute('class');` in whatever callback these functions use. 

Assuming a nonboolean attribute with no value is equivalent to no attribute at all is only true to an HTML renderer, but not necessarily to other DOM-aware code running on the page.

duanemoody:
> Hey, that's because htmx adds temporary [classes](https://htmx.org/reference/#classes) such as `htmx-request` to your elements while the requests are in flight, to let you use some styling when needed, such as a [`hx-indicator`](https://htmx.org/attributes/hx-indicator/) to display say a loading spinner until the request is done.
> 
[...]

PR #2683, opened 2024-06-27
Correct hx-verb leaving behind empty `class` attributes on elements

duanemoody:
## Description
Manually added check for empty class to `doSettle()` after it removes its own classes:
```elt.classList.length === 0 && elt.removeAttribute('class')```

## Corresponding issue:
#1701 reporting empty `class` attributes left behind after hx-verb actions. Other functions in htmx.js do `classList.length` checks but `doSettle` doesn't leverage any of them, so per the HTML spec empty class attributes are left behind. 

## Testing
I ran the modified code on elements with and without existing classes, and it behaved as expected: if existing class attributes were present with values, hx-put left them as-is; if no class attribute was present, none was left behind.

## Checklist

  source changes)
  approved via an issue


duanemoody:
Technically I could have just rewritten the line in `doSettle` that removes the utility class via `elt.classList.remove` to instead use htmx's own `removeClassFromElement` method (which does housekeeping on blank `class` attributes) but I suspected there was a reason why `doSettle` doesn't use it.

Telroshan:
Hey,
- It may have been an oversight at the time it was written, by someone who maybe was simply not aware of `removeClassFromElement`'s existence. Looking at the code, I can't think of a reason why it shouldn't be used here
- I see there are other places when `classList.remove` or `classList.add` are called and they're likely to oversights just as much. Would you like to fix those at the same time ?
- As per the [contribution guidelines](https://github.com/bigskysoftware/htmx/blob/master/CONTRIBUTING.md), could you please remove the dist files from the PR ?
> Please avoid sending the dist files along your PR, only include the src ones.

- As per the [contribution guidelines](https://github.com/bigskysoftware/htmx/blob/master/CONTRIBUTING.md), could you please retarget your PR to `dev`  ?
> All PRs must be made against the dev branch, except documentation PRs (that only modify the www/ directory) which can be made against master.

duanemoody:
> It may have been an oversight at the time it was written, by someone who maybe was simply not aware of `removeClassFromElement`'s existence. Looking at the code, I can't think of a reason why it shouldn't be used here
> 
> I see there are other places when `classList.remove` or `classList.add` are called and they're likely to oversights just as much. Would you like to fix those at the same time ?

I should target `htmx.js` only and not any of the other files by the sound of it. I'll get on normalizing the code. 

> Please avoid sending the dist files along your PR, only include the src ones.

Ouch, went overboard there.

> As per the [contribution guidelines](https://github.com/bigskysoftware/htmx/blob/master/CONTRIBUTING.md), could you please retarget your PR to `dev`  ?

Can do as soon as I figure out how to change that in GH.

duanemoody (commented):
Updated commit per requests to 
* Normalize replacement across the codebase
* Target `dev` instead of `master`
* Remove changes to files in `dist/`

duanemoody:
I commented after two lines that the original code used `ic.classList.add/remove.call` without a clear understanding of why that was necessary, but I suspect `addClassToElement` / `removeClassFromElement` already get enough context without needing `call`.

1cg:
Hey @duanemoody I think this PR needs a restart, the `/dist` stuff is checked in but shouldn't be modified, just modify the `/src/htmx.js` file and we'll regenerate those files on release.  Also would like to see a test for this if possible.

1cg:
Already made these fixes, thank you!  Sorry this sat so long!

PR #2696, opened 2024-07-05
Clarify that hx-trigger:from doesn't dynamically update

### LINKED ISSUE #2689: hx-trigger="click from:.class" are not updated
Given a div like:

```
<div id="container" hx-get="/target" hx-trigger="click from:.literal" class="thebox">
...
</div>
```

This works fine (finding all things which match `.literal`), but if I load a new thing via htmx, it is not scanned. Further, even if I run `htmx.process(document.documentElement)`, to (I believe) force a full scan, the things with class `literal` still aren't found, I think because htmx notices the `div` hasn't changed, so it doesn't rescan it.

If I "poke" the div, by giving it a new class (for example), then run `htmx.process(document.documentElement)`, then it will finally find new items with the `.literal` class.

While my actual problem is a htmx problem, I load a div, which contains an svg, where lots of items have the ".literal" class, and they don't become clickable.

I've tried to write a minimal awful HTML+Javascript example to show the problem (attached)

It has 3 buttons:

The first adds the class  'literal' to the text 'hello world', but it doesn't become clickable.

The second runs 'htmx.process', but that doesn't fix it

The third changes the div with the 'hx-get', attaching a new 'class' to it. This still doesn't fix it.

However, now reclicking on the second button now correctly rescans, and we can finally click on 'hello world', and get a result.

[example.zip](https://github.com/user-attachments/files/16070041/example.zip)

I feel like there are two(ish) issues here (but I might be misunderstanding my problem)

1) When HTMX loads something, it should maybe try reapplying any triggers with a 'from', to see if they apply to the new content.

2) htmx.process shouldn't try to cleverly skip things with a 'hx-trigger' with 'from' (if that's what it's doing).

ChrisJefferson:
Make some small tweaks to the docs to mention that hx-trigger:from doesn't update when the page changes.

Fixes #2689 , and gives an example from there.

I also fixed a typo in www/config.toml. This was actually a typo in zola, reported in https://github.com/getzola/zola/issues/2537 and fixed in the latest release. I can take this out if you want to stick to an older version of zola.

### DIFF
--- a/www/config.toml
+++ b/www/config.toml
@@ -4,7 +4,7 @@ theme = "htmx-theme"
 
 compile_sass = false
 build_search_index = false
-generate_feed = true
+generate_feeds = true
 
 taxonomies = [
   { name = "tag", render = false, feed = true },
--- a/www/content/attributes/hx-trigger.md
+++ b/www/content/attributes/hx-trigger.md
@@ -56,6 +56,7 @@ is seen again it will reset the delay.
 is seen again before the delay completes, it is ignored, the element will trigger at the end of the delay.
 * `from:<Extended CSS selector>` - allows the event that triggers a request to come from another element in the document (e.g. listening to a key event on the body, to support hot keys)
   * A standard CSS selector resolves to all elements matching that selector. Thus, `from:input` would listen on every input on the page.
+  * The CSS selector is only evaluated once and is not re-evaluated when the page changes. If you need to detect dynamically added elements use an event filter, for example `click[event.target.matches('input')]`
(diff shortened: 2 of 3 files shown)


ysm-dev:
I think RSS looks broken after this PR, @ChrisJefferson @1cg 
https://htmx.org/atom.xml

ChrisJefferson:
Not sure how the website is actually generated (manually?) I'd guess it was the fix to www/config.toml , if you are using a different version of zola to me?

PR #2701, opened 2024-07-06
HtmxAjaxHelperContext typedef should have all optional fields

anentropic:
## Description
The typedef is currently more strict than the code, and prevents IMHO the most useful way to configure ajax requests.

I also added a clarification to the `htmx.ajax` docs about what specifying `source` in the context object actually does.

See discussion here:
https://github.com/bigskysoftware/htmx/discussions/2698

## Testing

## Checklist

  source changes)
   ...oh, do the docs and code changes need to be in separate PRs?
  approved via an issue

### DIFF
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -113,7 +113,7 @@ type HtmxAjaxHelperContext = {
     source?: Element | string;
     event?: Event;
     handler?: HtmxAjaxHandler;
-    target: Element | string;
+    target?: Element | string;
     swap?: HtmxSwapStyle;
     values?: any | FormData;
     headers?: Record<string, string>;
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -5036,7 +5036,7 @@ var htmx = (function() {
  * @property {Element|string} [source]
  * @property {Event} [event]
  * @property {HtmxAjaxHandler} [handler]
- * @property {Element|string} target
+ * @property {Element|string} [target]
  * @property {HtmxSwapStyle} [swap]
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -35,7 +35,7 @@ npm test
 
 ## Pull Requests
 ### Technical Requirements
-1. Please lint all proposed changes with the `npm lint-fix` command
+1. Please lint all proposed changes with the `npm run lint --fix` command
```
anentropic, line 38: there is no `npm lint-fix` command, there is a `lint` action in package.json scripts... I think it means this?

Telroshan: That command was removed by a recent PR if I recall correctly, I believe it's `npm run format` now

Telroshan:
> do the docs and code changes need to be in separate PRs?

Since the code changes here only affect the JSDoc thus the "code's documentation" + it comes from the same source issue/discussion, I would say it's totally fine.

Though, could you please remove the changes to the duplicated htmx scripts in `www`?

anentropic:
> > do the docs and code changes need to be in separate PRs?
> 
> Since the code changes here only affect the JSDoc thus the "code's documentation" + it comes from the same source issue/discussion, I would say it's totally fine.
> 
> Though, could you please :
> 
> * remove the changes to the duplicated htmx scripts in `www`? Those are generated from `src` or `dist` when we publish a release
> * remove the change to `CONTRIBUTING.md`? I think we'll do that in another PR as it's not directly related to the problem here

Thanks, have made those changes.

Issue #2702, opened 2024-07-06
htmx 2.0.0 breaks <template> elements inside Web Components


luontola:
When a htmx response contains `<template>` elements inside a Web Component element, htmx 1.9.12 adds them to the DOM, but htmx 2.0.0 removes them. This breaks web components which use that method to receive data from the server.

## Example app

1. Save this file as `server.py`
```python
from flask import Flask

app = Flask(__name__)


@app.route("/")
def home():
    # Works with https://unpkg.com/htmx.org@1.9.12/dist/htmx.min.js
    # Doesn't work with https://unpkg.com/htmx.org@2.0.0/dist/htmx.min.js
    return """
<script src="https://unpkg.com/htmx.org@2.0.0/dist/htmx.min.js"></script>
<script type="module">
class HelloWorldElement extends HTMLElement {
  constructor() {
    super();
  }

  connectedCallback() {
    const template = this.querySelector("template");
    console.log("template:", template, template?.content.textContent)
  }
}
customElements.define('hello-world', HelloWorldElement);
</script>

<div hx-get="/lazy" hx-swap="outerHTML" hx-target="this" hx-trigger="load">
</div>

</div>
"""


@app.route("/lazy")
def lazy():
    return """
<hello-world>
<template>Lazy loaded content</template>
</hello-world>
"""

```

2. Run the app with the commands:

```
pip install Flask
flask --app server.py --debug run
```

3. Open the app at http://127.0.0.1:5000/

4. Open the JavaScript console and see what is printed there on page load

### Expected result

- The code prints `template: <template>​…​</template>​ Lazy loaded content`
- This is what happens on htmx 1.9.12

### Actual result

- The code prints `template: null undefined`
- This is what happens on htmx 2.0.0

kubeden:
Hey, after some investigation, I found out the following function added to 2.0.0 which does not exist in 1.9.12:

```
      // oob swaps
      findAndSwapOobElements(fragment, settleInfo)
      forEach(findAll(fragment, 'template'), /** @param {HTMLTemplateElement} template */function(template) {
        findAndSwapOobElements(template.content, settleInfo)
        if (template.content.childElementCount === 0) {
        // Avoid polluting the DOM with empty templates that were only used to encapsulate oob swap
          template.remove()
        }
      })
```

You can see here that there is a comment _"// Avoid polluting the DOM with empty templates that were only used to encapsulate oob swap"_

[...]

PR #2705, opened 2024-07-10
Bugfix/2702: htmx 2.0.0 breaks <template> elements inside Web Components

kubeden:
oob swap to check if a template content contains not only 0 elements but 0 characters too (fixes plaintext templates being removed: #2702 )

## Description

I introduced an additional check `template.content.textContent.trim() === ''` to make sure templates are not removed if they have plaintext inside.

Corresponding issue:
#2702 

## Testing

I ran the sample server.py code provided by the person opening the PR. I then changed the source to htmx to my localhost htmx src. I added the additional check in the if statement and refreshed a few times.

**Without my change:**

with plaintext:
image

without plaintext:
image

**With my change:**

with plaintext:
image

without plaintext:
image

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -113,7 +113,7 @@ type HtmxAjaxHelperContext = {
     source?: Element | string;
     event?: Event;
     handler?: HtmxAjaxHandler;
-    target: Element | string;
+    target?: Element | string;
     swap?: HtmxSwapStyle;
     values?: any | FormData;
     headers?: Record<string, string>;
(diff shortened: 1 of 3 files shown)


kubeden:
Hey. This is what was discussed in the issue and the two questions that emerged:

1. Is it a good practice to have plaintext in <template/>? Probably not but it seems that this generates problems for people.
2. Is HTMX looping through every <template> in the document? If that is the case maybe an "enchantment" that checks if a template has been modified by HTMX, and only then remove it, would solve this?

Other than that, if I did something wrong, tell me and I'll look for a way to do it better.

Telroshan:
Hey, you seem to have some unrelated changes carried by other commits ; could you please rebase against `dev`? So that only your changes appear in the diff

PR #2711, opened 2024-07-11
Extension perf improvments

MichaelWest22:
## Description
I noticed while debugging and investigating how the extension code works that when not using a large range of extensions it still has to traverse up the DOM to search for active extensions on the active element for all the plugin extension points that use  withExtensions(). Even with no extensions installed it performs this additional effort.  Also I found that during the findElementsToProcess() function it was calling all extensions to build a recent extensionSelectors feature on every process when this is static work that could be performed instead once at extension install time to speed things up a little.

So I've made the following performance improvements:

1. During Extension Activation It now generates an extensionActions array of all function names the installed extensions are using which is deduped 
2. Changed getExtensions() to have an action optional argument where you can provide it with the action you need to run on all extensions and it checks if none of the installed extensions implement this action it returns so it doesn't have to recurse up the DOM inspecting element attributes for hx-ext
3. Updated all calls to getExtensions to pass in the required action.
4. Created extensionSelectors string that is filled in once during extension activation and then used as a static string on each node process.

Note that this change does change the definition of the withExtensions() function published to extensions via internalAPI by adding a new action optional parameter.  But all existing extensions will keep working with existing behavior as they will not be passing this new parameter in which means they will not take advantage of the minor performance gain.  Extensions could adopt the use of the action parameter but then they would no longer be compatible with older versions of htmx 2.0 so probably best left as is till this change is well established.

Corresponding issue: #2712

## Testing
Was not able to get new working tests for these internal only changes so to test this I test loading real and manual dummy extensions to confirm the two new objects were getting populated as expected and added some console.logs to check it has reduced the number of calls made to lookup extensions.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1394,7 +1394,7 @@ var htmx = (function() {
    * @returns {boolean}
    */
   function isInlineSwap(swapStyle, target) {
-    const extensions = getExtensions(target)
+    const extensions = getExtensions(target, 'isInlineSwap')
     for (let i = 0; i < extensions.length; i++) {
       const extension = extensions[i]
       try {
@@ -1747,7 +1747,7 @@ var htmx = (function() {
         swapDelete(target)
         return
       default:
-        var extensions = getExtensions(elt)
+        var extensions = getExtensions(elt, 'handleSwap')
         for (let i = 0; i < extensions.length; i++) {
           const ext = extensions[i]
           try {
@@ -2651,20 +2651,8 @@ var htmx = (function() {
   function findElementsToProcess(elt) {
     if (elt.querySelectorAll) {
       const boostedSelector = ', [hx-boost] a, [data-hx-boost] a, a[hx-boost], a[data-hx-boost]'
-
-      const extensionSelectors = []
-      for (const e in extensions) {
-        const extension = extensions[e]
-        if (extension.getSelectors) {
-          var selectors = extension.getSelectors()
-          if (selectors) {
-            extensionSelectors.push(selectors)
-          }
-        }
-      }
-
       const results = elt.querySelectorAll(VERB_SELECTOR + boostedSelector + ", form, [type='submit']," +
-        ' [hx-ext], [data-hx-ext], [hx-trigger], [data-hx-trigger]' + extensionSelectors.flat().map(s => ', ' + s).join(''))
+        ' [hx-ext], [data-hx-ext], [hx-trigger], [data-hx-trigger]' + extensionSelectors)
 
       return results
     } else {
@@ -2896,14 +2884,16 @@ var htmx = (function() {
   /**
    * `withExtensions` locates all active extensions for a provided element, then
    * executes the provided function using each of the active extensions.  It should
-   * be called internally at every extendable execution point in htmx.
+   * be called internally at every extendable execution point in htmx. Optionally 
+   * pass the extension action name you are using to improve lookup performance. 
    *
    * @param {Element} elt
    * @param {(extension:HtmxExtension) => void} toDo
+   * @param {string=} action
    * @returns void
    */
-  function withExtensions(elt, toDo) {
-    forEach(getExtensions(elt), function(extension) {
+  function withExtensions(elt, toDo, action) {
+    forEach(getExtensions(elt, action), function(extension) {
       try {
         toDo(extension)
       } catch (e) {
@@ -2952,7 +2942,7 @@ var htmx = (function() {
     }
     withExtensions(asElement(elt), function(extension) {
       eventResult = eventResult && (extension.onEvent(eventName, event) !== false && !event.defaultPrevented)
-    })
+    }, 'onEvent')
     return eventResult
   }
 
@@ -3630,7 +3620,7 @@ var htmx = (function() {
       if (encodedParameters == null) {
         encodedParameters = extension.encodeParameters(xhr, filteredParameters, elt)
       }
-    })
+    }, 'encodeParameters')
     if (encodedParameters != null) {
       return encodedParameters
     } else {
@@ -4659,7 +4649,7 @@ var htmx = (function() {
 
       withExtensions(elt, function(extension) {
         serverResponse = extension.transformResponse(serverResponse, xhr, elt)
-      })
+      },'transformResponse')
 
       // Save current page if there will be a history update
       if (historyUpdate.type) {
@@ -4782,6 +4772,12 @@ var htmx = (function() {
   /** @type {Object<string, HtmxExtension>} */
   const extensions = {}
 
+  /** @type {string[]} */
+  var extensionActions = []
+
+  /** @type {string} */
+  var extensionSelectors = ''
+
   /**
    * extensionBase defines the default functions for all extensions.
    * @returns {HtmxExtension}
@@ -4811,6 +4807,13 @@ var htmx = (function() {
       extension.init(internalAPI)
     }
     extensions[name] = mergeObjects(extensionBase(), extension)
+    extensionActions = [...new Set(extensionActions.concat(Object.keys(extension)))]
+    if (extension.getSelectors) {
+      var selectors = extension.getSelectors()
+      if (selectors) {
+        extensionSelectors = extensionSelectors + selectors.flat().map(s => ', ' + s).join('')
+      }
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Telroshan:
Hey, while I understand the theoretical performance concern, did you measure the actual performance impact on say a large DOM hierarchy or, more broadly, the worst case scenario? To see what the impact is in this case. 

I see 2 downsides to the suggested approach:
- I don't know all existing extensions but what if an extension has "dynamic selectors" of some sort? What would have been evaluated at every call until now would suddenly be evaluated only at the extension register moment, I'm afraid it could be a breaking change
- It adds more hardcoded strings in the code, which cannot be minified, and we'd like to avoid that as much as possible as it increases the lib size. A minor inconvenience here given the amount of text, I'll agree, but if the performance impact turns out to be even more minor than that... Hence the measurement request!

Note that imho there's no need to address any of the 2 points mentioned above, before a performance test case is produced.
I'd like to first see codesamples and their performance result, both for the current lib state, and the suggested one, before moving on with the retro-compatibility / minification issues!
As after all, if the performance gain is interesting, we'll find a way to work around those constraints anyway.

MichaelWest22:
thanks for looking at this @Telroshan

With regard to the first possible downside I researched when this feature was added and it was very recently only a few months ago in #2505 where it was added and as far as I can see from all my searches only so far used in SSE extension to solve a duplicate processing issue and seems to be only have been intended for a static selector list which is why I thought it could be optimized.

But yeah I had not considered the final minified gziped size.  should probably run that quickly to check the change.

Will see if I can work on some perf tests.  With small virtually empty DOM like all my testing so far the performance difference would be impossible to notice I think so i may have to get a more realistic page and repeat the tasks multiple times.

MichaelWest22:
```
Before
-rwxrwxrwx 1 mwest mwest  49082 Jul 12 20:30 htmx.min.js
-rwxrwxrwx 1 mwest mwest  15761 Jul 12 20:30 htmx.min.js.gz
After
-rwxrwxrwx 1 mwest mwest  49267 Jul 12 20:28 htmx.min.js     + 185
-rwxrwxrwx 1 mwest mwest  15812 Jul 12 20:28 htmx.min.js.gz  + 51
``` 
So it does add a little to the min size as you suggested

Perf testing this change is not 100% easy as the calls i'm optimizing are only a small % of the total time to do real requests so maybe this perf optimization is just not really needed.

One method i've found to get something meaningful so far is to add the following code  before getExtension and rename getExtension to getExtension2
```
  function getExtensions(elt) {
    var ret
    for (var i = 1; i <= 1000; i++) {
      ret = getExtensions2(elt)
    }
    return ret
  }
```
What this does is makes the extension lookup code ~1000 times worse so I can isolate just this portion of the code.  Then i run the full htmx test suite except with the 3 extension test scripts commented out.

I got the following results:

```
Before my perf Optimizations:
    ✅ htmx:sendError is called after a failed request (40ms)
    ✅ htmx:afterRequest is called when replacing outerHTML (31ms)
    ✅ htmx:afterOnLoad is called when replacing outerHTML (19ms)
    ✅ htmx:beforeProcessNode is called when replacing outerHTML (20ms)
    ✅ htmx:beforeProcessNode allows htmx attribute tweaking (23ms)
    ✅ htmx:afterProcessNode is called after replacing outerHTML (29ms)
    ✅ htmx:afterRequest is called when targeting a parent div (30ms)
    ✅ adding an error in htmx:configRequest stops the request (5ms)
    ✅ preventDefault() in htmx:configRequest stops the request (5ms)
    ✅ preventDefault() in the htmx:beforeRequest event cancels the request (5ms)
    ✅ preventDefault() in the htmx:beforeOnLoad event cancels the swap (20ms)
    ✅ htmx:afterRequest event contains 'successful' and 'failed' properties indicating success after successful request (18ms)
Response Status Error Code 500 from /test
    ✅ htmx:afterRequest event contains 'successful' and 'failed' properties indicating failure after failed request (25ms)
    ✅ htmx:confirm can cancel request (29ms)
beforeSwap {"htmx-internal-data":{"initHash":3227304}} {"htmx-internal-data":{"initHash":3227305}}
afterSwap {"htmx-internal-data":{"initHash":3227305}} {"htmx-internal-data":{"initHash":3227305}}
    ✅ has updated target available when target set via htmx:beforeSwap (21ms)


  632 passing (19s)
  3 pending
```

```
After my perf Optimizations:
    ✅ htmx:sendError is called after a failed request (32ms)
    ✅ htmx:afterRequest is called when replacing outerHTML (1ms)
    ✅ htmx:afterOnLoad is called when replacing outerHTML (1ms)
    ✅ htmx:beforeProcessNode is called when replacing outerHTML (1ms)
    ✅ htmx:beforeProcessNode allows htmx attribute tweaking (1ms)
    ✅ htmx:afterProcessNode is called after replacing outerHTML (1ms)
    ✅ htmx:afterRequest is called when targeting a parent div (1ms)
    ✅ adding an error in htmx:configRequest stops the request
    ✅ preventDefault() in htmx:configRequest stops the request (1ms)
    ✅ preventDefault() in the htmx:beforeRequest event cancels the request
    ✅ preventDefault() in the htmx:beforeOnLoad event cancels the swap (1ms)
    ✅ htmx:afterRequest event contains 'successful' and 'failed' properties indicating success after successful request (1ms)
Response Status Error Code 500 from /test
    ✅ htmx:afterRequest event contains 'successful' and 'failed' properties indicating failure after failed request (1ms)
    ✅ htmx:confirm can cancel request (1ms)
beforeSwap {"htmx-internal-data":{"initHash":3227304}} {"htmx-internal-data":{"initHash":3227305}}
afterSwap {"htmx-internal-data":{"initHash":3227305}} {"htmx-internal-data":{"initHash":3227305}}
    ✅ has updated target available when target set via htmx:beforeSwap (1ms)


  632 passing (3s)
  3 pending
```
So from this quick test it shows that it saved 16 seconds compared to the 3 second normal run.  Both runs had the 1000 times extension checking loop code but with my updated version this didn't add any time to the normal 3 second test suite.  If we divide my very impresive 16 second speed gain by 1000 to find the real perf gain and we are then down to just 16ms which does not seem that impressive you can say I just saved 16 ms in a 3000ms test suite which is like 0.5% perf gain for a 0.32% bigger gz size.  

Also when I include the 3 extension tests it didn't change the 19 second old code run timing by much but my code then took 13 seconds and was much slower than before.  This is because once you get 1/4 of the way though the tests and it loads the extensions most if not all of the extension actions get activated and this disables my optimizations as it only gives perf savings when only some extension actions are enabled.

So I think it does give a good speed gain for the situations where extensions are used sparingly but it is all sub ms savings per request so may not be worth the effort.  May still be useful to generate a more realistic complex web page and do some more testing on that though.

Telroshan:
> may not be worth the effort

Yeah I would say that 0.5% of perf gain is indeed really too low, but there might be more to it!

I think some tests in the test suite do network calls to remote URLs (there's a timeout test I think), which could interfere with the performance measurements.
And as there are a lot of things happening at the same time here, it's hard to conclude anything from it tbh.
Even if you put a 1000 iterations loop (which is not that much for a computer btw!), it would be best to isolate the performance measurement as well.

A good test could be to make htmx process a huge DOM with say 10 000 elements of hx-attributes, something like that, and measure the performance impact before/after this change. Simply increase the amount of elements to process as needed to get a more precise performance measurement.
The goal here is to isolate as much as possible the extensions lookup that we're talking about. You can use `performance.now` for example and increment a global counter that you divide by the iterations count at the end, and/or store min/max excution times of each solution.

For example, see PR #1540 where I had suggested a performance improvement for trigger specs evaluation.
In this PR, the suggested solution (a simple cache object) gave a substantial performance impact (~x46 speed), but this was only noticeable on a large DOM (here, 2000 elements using `hx-trigger`).

Which is why I'm suggesting, if you feel like it, making more precise performance measurements!

MichaelWest22:
I'm going to close this PR as all my further testing was unable to show a meaningful performance gain.  I looked at a few other options to try and optimize it but the time spent checking for extensions being active or not on a element was simply not a real issue and is fast enough.  I tested with crazy large documents with 1000 nested div's with up to 100MB of total raw html content and the code to find the extension tags up this large DOM had almost no major impact on performance.  It did get slow to respond but the extension lookup was not the cause.  When i added back in my fake change to make it repeat getExtensions 1000 times it did make a huge performance change showing my performance changes were doing something but with my 1000 nested divs and the 1000 time repeat added together it was a crazy test that often timed out and broke the tests and I don't think this is relevant.

PR #2715, opened 2024-07-12
Fix example for responseHandling config via <meta> tag

ekzyis:
## Description

In the docs, the following example is given to configure response handling via the `<meta>` tag:

```html
<meta name="htmx-config" content='{code:"204", swap: false},   // 204 - No Content by default does nothing, but is not an error
                                  {code:"[23]..", swap: true}, // 200 & 300 responses are non-errors and are swapped
                                  {code:"422", swap: true}, // 422 responses are swapped
                                  {code:"[45]..", swap: false, error:true}, // 400 & 500 responses are not swapped and are errors'>
```

However, if this is used in a site which loads [htmx v0.2.0](https://unpkg.com/htmx.org@2.0.0/dist/htmx.js), the following error is thrown:

```
htmx.js:2914 SyntaxError: Expected property name or '}' in JSON at position 1 (line 1 column 2)
    at JSON.parse (<anonymous>)
    at parseJSON (htmx.js:816:19)
    at getMetaConfig (htmx.js:4902:14)
    at mergeMetaConfig (htmx.js:4909:24)
    at HTMLDocument.<anonymous> (htmx.js:4917:5)
```

Fixing all errors like unquoted keys, comments and missing `responseHandling` key lead me to this config:

```html
<meta
	name="htmx-config"
	content='{
		"responseHandling":[
			{"code":"204", "swap": false},
			{"code":"[23]..", "swap": true},
			{"code":"422", "swap": true},
			{"code":"[45]..", "swap": false, "error":true}
		]
	}'
/>
```

It's unfortunate that JSON does not support comments but I think having a working example in the documentation is more important.

Corresponding issue: n/a

## Testing

I tested this change by copy-pasting the configuration into `<head>` of my site

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/docs.md
+++ b/www/content/docs.md
@@ -972,10 +972,17 @@ Using the [meta config](#configuration-options) mechanism for configuring respon
 config:
 
 ```html
-<meta name="htmx-config" content='{code:"204", swap: false},   // 204 - No Content by default does nothing, but is not an error
-                                  {code:"[23]..", swap: true}, // 200 & 300 responses are non-errors and are swapped
-                                  {code:"422", swap: true}, // 422 responses are swapped
-                                  {code:"[45]..", swap: false, error:true}, // 400 & 500 responses are not swapped and are errors'>
+<meta
+	name="htmx-config"
+	content='{
+        "responseHandling":[
+            {"code":"204", "swap": false},
+            {"code":"[23]..", "swap": true},
+            {"code":"422", "swap": true},
+            {"code":"[45]..", "swap": false, "error":true}
+        ]
+    }'
+/>
 ```
 
 If you wanted to swap everything, regardless of HTTP response code, you could use this configuration:


gemmadlou:
Hi @ekzyis 
I had this problem too and realised I need to format it in JSON. Thanks for the fix. Hope it gets reviewed soon.

But what I failed to understand was I needed to include the `responseHandling` property on the JSON object, whereas I was just doing this:

```
<meta name="htmx-config" content='{"code":".*", "swap": true}'>
```

But this works a treat!

```
<meta name="htmx-config" content='{
        "responseHandling":[
            {"code":".*", "swap": true}
        ]
    }' />
```

alexpetros:
First of all, great fix, much appreciated.

Would you mind updating it do include the original commented info as an *HTML* comment above the config? I do think the explanation is necessary there.

geoffrey-eisenbarth:
One thing that I think is important to specify in the documentation is that if there are any status codes that you don't handle in the config, they won't be swapped. I mentioned this on discord and Carson mentioned that you should include the catch-all case `"code": "...", "swap": true`

ekzyis:
Oh, sorry for the slow response, I kept forgetting after I read the notifications to do the changes when I got home and now I read the notification that this was merged now.

> Would you mind updating it to include the original commented info as an HTML comment above the config? I do think the explanation is necessary there.

I'll add this now and the catch-all case!

alexpetros:
Carson just merged it so just open a new PR and tag me :)

PR #2721, opened 2024-07-13
fix: type definitions for HtmxExtension

cngJo:
The @see tag prevented `tsc` to properly generate types for `HtmxExtension`. Moving the `@see` above `@typedef` fixes this and adds types for `HtmxExtension` back to the package ;-)

## Description

https://github.com/bigskysoftware/htmx/pull/2336 updated type generation but as a side effect the type definition of `HtmxExtension` got dropped / changed to `any`. 

This is fixed now ;) 

Note: I did not create an issue to prevent spam, if it's important, I'm more than happy to create one after the fact 😉

## Testing

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1753,7 +1753,7 @@ var htmx = (function() {
           try {
             const newElements = ext.handleSwap(swapStyle, target, fragment, settleInfo)
             if (newElements) {
-              if (typeof newElements.length !== 'undefined') {
+              if (Array.isArray(newElements)) {
                 // if handleSwap returns an array (like) of elements, we handle them
                 for (let j = 0; j < newElements.length; j++) {
                   const child = newElements[j]
@@ -5121,12 +5121,13 @@ var htmx = (function() {
  */
 
 /**
+ * @see https://github.com/bigskysoftware/htmx-extensions/blob/main/README.md
  * @typedef {Object} HtmxExtension
- * @see https://htmx.org/extensions/#defining
  * @property {(api: any) => void} init
  * @property {(name: string, event: Event|CustomEvent) => boolean} onEvent
  * @property {(text: string, xhr: XMLHttpRequest, elt: Element) => string} transformResponse
  * @property {(swapStyle: HtmxSwapStyle) => boolean} isInlineSwap
- * @property {(swapStyle: HtmxSwapStyle, target: Element, fragment: Node, settleInfo: HtmxSettleInfo) => boolean} handleSwap
- * @property {(xhr: XMLHttpRequest, parameters: FormData, elt: Element) => *|string|null} encodeParameters
+ * @property {(swapStyle: HtmxSwapStyle, target: Node, fragment: Node, settleInfo: HtmxSettleInfo) => boolean|Node[]} handleSwap
+ * @property {(xhr: XMLHttpRequest, parameters: FormData, elt: Node) => *|string|null} encodeParameters
+ * @property {() => string[]|null} getSelectors
  */
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -15,7 +15,7 @@ declare namespace htmx {
     const toggleClass: (elt: string | Element, clazz: string) => void;
     const takeClass: (elt: string | Node, clazz: string) => void;
     const swap: (target: string | Element, content: string, swapSpec: HtmxSwapSpecification, swapOptions?: SwapOptions) => void;
-    const defineExtension: (name: string, extension: any) => void;
+    const defineExtension: (name: string, extension: HtmxExtension) => void;
     const removeExtension: (name: string) => void;
     const logAll: () => void;
     const logNone: () => void;
@@ -192,4 +192,12 @@ type HtmxSettleInfo = {
     elts: Element[];
     title?: string;
 };
-type HtmxExtension = any;
+type HtmxExtension = {
+    init: (api: any) => void;
+    onEvent: (name: string, event: Event | CustomEvent) => boolean;
+    transformResponse: (text: string, xhr: XMLHttpRequest, elt: Element) => string;
+    isInlineSwap: (swapStyle: HtmxSwapStyle) => boolean;
+    handleSwap: (swapStyle: HtmxSwapStyle, target: Node, fragment: Node, settleInfo: HtmxSettleInfo) => boolean | Node[];
+    encodeParameters: (xhr: XMLHttpRequest, parameters: FormData, elt: Node) => any | string | null;
+    getSelectors: () => string[] | null;
+};


MichaelWest22:
I found a missing new extension field

 * @property {() => string[]} getSelectors

when i was working on #2711 so this may need to be added in.  This was added a few months ago for the SSE extension but it may not have been updated fully in the extension repo as well i'm thinking

cngJo:
The two errors in the test are related to `handleSwap` on `HtmxExtension` being expected to return an array of elements instead of `boolean` what the type suggests. 

I took a look at the implementations at `htmx-extensions` for `handleSwap` and saw both usages: 

- [`multi-swap`](https://github.com/bigskysoftware/htmx-extensions/blob/719538f486614ba48a16c7b53a4ca1f069132d97/src/multi-swap/multi-swap.js#L40) returns a boolean
- [`alpine-swap`](https://github.com/bigskysoftware/htmx-extensions/blob/719538f486614ba48a16c7b53a4ca1f069132d97/src/alpine-morph/alpine-morph.js#L9-L12) returns an array of nodes.
- [`morphdom-swap`](https://github.com/bigskysoftware/htmx-extensions/blob/719538f486614ba48a16c7b53a4ca1f069132d97/src/morphdom-swap/morphdom-swap.js#L10-L13) returns an array of nodes

What would the expected behavior be? I'd offer to update `htmx-extensions` with that info as well.

cngJo:
@MichaelWest22 thats for noticing, I found it when checking the failing tests and just force-pushed a fix adding it to the `HtmxExtension`  type.

MichaelWest22:
Also the see web link is now linking to 404 after htmx 2 upgrade and the split from to a new extension.  maybe this see link needs to link to https://github.com/bigskysoftware/htmx-extensions/blob/main/README.md instead maybe since I can't see any other place this info is now.  the info here also probably needs updating...

cngJo:
There is https://htmx.org/docs/#extensions, but itself just points to the README you linked. I've updated the link accordingly and going to file a PR against https://github.com/bigskysoftware/htmx-extensions to update the type information in the README as well.

src/htmx.js
```diff
@@ -5121,12 +5121,13 @@ var htmx = (function() {
...
+ * @see https://github.com/bigskysoftware/htmx-extensions/blob/main/README.md
  * @typedef {Object} HtmxExtension
- * @see https://htmx.org/extensions/#defining
  * @property {(api: any) => void} init
  * @property {(name: string, event: Event|CustomEvent) => boolean} onEvent
  * @property {(text: string, xhr: XMLHttpRequest, elt: Element) => string} transformResponse
  * @property {(swapStyle: HtmxSwapStyle) => boolean} isInlineSwap
- * @property {(swapStyle: HtmxSwapStyle, target: Element, fragment: Node, settleInfo: HtmxSettleInfo) => boolean} handleSwap
- * @property {(xhr: XMLHttpRequest, parameters: FormData, elt: Element) => *|string|null} encodeParameters
+ * @property {(swapStyle: HtmxSwapStyle, target: Node, fragment: Node, settleInfo: HtmxSettleInfo) => boolean} handleSwap
```
Telroshan, line 5130: Hey, good catch!
You'll probably want to update the JSDoc of `handleSwap` to indicate that double return type then (only specifies a `boolean` as return type atm), as tsc currently fails the types generation step (see the errors reported by the CI above)

Telroshan (changes requested):

src/htmx.js
cngJo: Thanks! That was basically what I was asking for, and allowing both types is for sure the best option!

Unfortunately, TSC seems to not properly handle the `typeof newElements.length !== 'undefined'` as it still throws the same type error. I therefore changed the condition to a `Array.isArray` call, which is widely available in all Browsers: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

cngJo:
@Telroshan I updated the return type of `getSelectors` to also allow null, as per https://github.com/bigskysoftware/htmx-extensions/pull/49#discussion_r1676854418

Telroshan:
Perfect thanks! Last request; could you remove the changes to `dist` as per the [contributing guidelines](https://github.com/bigskysoftware/htmx/blob/master/CONTRIBUTING.md) ? We generate those files when cutting a new release

cngJo:
Oh, sorry ... but sure - I removed the changes from the `dist/` directory :-)

Telroshan (approved):

Telroshan:
Thanks!

maddalax:
@cngJo Hi, just curious, was it intentional that each field on HtmxExtension is required with the new type? I believe on HTMX v1, just the onEvent was sufficient for creating a new extension.

With this change, TS forces you to add each field to the extension obj, even if you aren't needing to override it

Telroshan:
Hey, nope that's an oversight @maddalax indeed

Though the issue isn't that PR in itself (that introduced back the types for HtmxExtension that was a `any` before it).
The issue is in the JSDoc of `HtmxExtension` that declares the properties.
The thing is, when _using_ an extension, all those properties will indeed be set and called by htmx, thus are mandatory, but when _defining_ an extension, htmx merges your passed object with the extension base object.

Maybe we should declare a separate JSDoc for the `defineExtension` where the properties are optional, but just for this function, as for any other usage of extensions, we expect those properties to be well defined.

If you'd like to investigate into it and submit a fix PR @maddalax, please do! PR welcome

See https://github.com/bigskysoftware/htmx/blob/e6a2ea15a2254986bd40160d62afbb289aa0f3dd/src/htmx.js#L4805-L4834

maddalax:
> Hey, nope that's an oversight @maddalax indeed
> 
> Though the issue isn't that PR in itself (that introduced back the types for HtmxExtension that was a `any` before it). The issue is in the JSDoc of `HtmxExtension` that declares the properties. The thing is, when _using_ an extension, all those properties will indeed be set and called by htmx, thus are mandatory, but when _defining_ an extension, htmx merges your passed object with the extension base object.
> 
> Maybe we should declare a separate JSDoc for the `defineExtension` where the properties are optional, but just for this function, as for any other usage of extensions, we expect those properties to be well defined.
> 
> If you'd like to investigate into it and submit a fix PR @maddalax, please do! PR welcome
> 
> See
> 
> https://github.com/bigskysoftware/htmx/blob/e6a2ea15a2254986bd40160d62afbb289aa0f3dd/src/htmx.js#L4805-L4834

Ah gotcha sorry I skipped over the part that it was auto generated. I may have some time to investigate this today, will see if I can come up with anything.

Thanks for letting me know :)

PR #2723, opened 2024-07-13
Allow htmx:beforeSwap to change swapOverride

JanRomero:
This half-fixes a regression, because this used to work once by updating `event.detail.etc.swapOverride`. This PR however expects the swap style in `event.detail.swapOverride` to match the other overrides.

I removed the bit about the `HX-Reswap` header because the same thing already happens further up, and the header value should be overridable by the event.

## Testing
I've added a test that will check if the swap style set in `htmx:beforeSwap` was **actually** used regardless of the `hx-swap` attribute or the `HX-Reswap` header. The header override still works if `htmx:beforeSwap` leaves it alone.

## Checklist

  source changes)
  approved via an issue

Cheers and sorry about opening this twice, I thought I could rename the branch without destroying the old PR...

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4667,7 +4667,7 @@ var htmx = (function() {
       if (historyUpdate.type) {
         saveCurrentPageToHistory()
       }
-      
+
       var swapSpec = getSwapSpecification(elt, swapOverride)
 
       if (!swapSpec.hasOwnProperty('ignoreTitle')) {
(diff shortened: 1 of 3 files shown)


Telroshan:
Hey, please run tests locally that should include the lint test which doesn't pass here.
Also, please note that the typescript definition file is generated from the JSDoc, which you will probably want to update to get that new override property in

JanRomero:
Ah, thanks, I think I got it now???

src/htmx.js
```diff
@@ -4666,9 +4668,6 @@ var htmx = (function() {
         saveCurrentPageToHistory()
       }
 
-      if (hasHeader(xhr, /HX-Reswap:/i)) {
-        swapOverride = xhr.getResponseHeader('HX-Reswap')
-      }
```
1cg, line 4671: Is the `HX-Reswap` header being removed here?

JanRomero: Not sure what you mean, but I removed that bit because it would override the decision of the event and the same thing has already happened in line 4627 anyway. So the header is still honored, just the event comes after it and can still change the swap style. That order makes sense to me.

Issue #2733, opened 2024-07-16
`'htmx.d.ts' is not a module`


Alexerson:
With htmx 2.0.1, I get the following error when trying to `import htmx from "htmx.org";`:

`File './node_modules/htmx.org/dist/htmx.d.ts' is not a module.ts(2306)`

I believe this is happening because the `htmx.d.ts` file should include `export default htmx;`.
Adding this line manually in the file fixes the issue. 

I will create a PR for this.

renannprado:
I worked around this issue by adding the following snippet to my `tsconfig.json`:

```json
{
    "typeAcquisition": {
        "include": ["node_modules/htmx.org/dist/htmx.d.ts"],
        "enable": true
    }
}
```

Alexerson:
I looked into this, and I’m not sure it’s fixing the issue though… 
Do you mind showing how you import the lib in your ts file?

renannprado:
> I looked into this, and I’m not sure it’s fixing the issue though… Do you mind showing how you import the lib in your ts file? Also, are you on 2.0.0 or 2.0.1?

@Alexerson I edited my comment. It was actually not correct as the `typeAcquisition` property didn't fix it in the end (it was a lot of trial an error and I was mislead to believe that config fixed it).

I also added the relevant typescript snippet. I hope it helps.

Alexerson:
Ah, thanks. But in that case, isn’t `htmxLib` just typed as `any` in your code?
My temporary solution is to manually copy the version of the .d.ts file I generate from the .esm file and manually add it in my project. I strongly believe that my PR is the proper fix though.

renannprado:
> Ah, thanks. But in that case, isn’t `htmxLib` just typed as `any` in your code?

No it's not, because of the window declaration.

alexpetros:
Fixed by: https://github.com/bigskysoftware/htmx/pull/2769

PR #2734, opened 2024-07-16
Generate .d.ts file from the esm module (#2733)

Alexerson:
## Description
Generating the `htmx.d.ts` file from `src/htmx.js` means the `export` statements are not included in the types file. Because typescript users will always want to import the `esm` file, it makes more sense to generate the types file from the module.

Corresponding issue: #2733 

## Testing
After generating the new type definition, I copied the content of `dist/` to my local project `node_modules` and the error disappears and the project builds properly.

## Checklist

  source changes)
  approved via an issue


alexpetros:
Cool, this works. Avoiding more config files is a priority of ours, and I appreciate your flexibility on finding a way forward.

alexpetros (approved):

package.json
jasonrm, line 1: I think [files](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#files) needs to be updated with the new name.

`dist/htmx.d.ts` -> `dist/htmx.esm.d.ts`

Alexerson: I believe you are right. I’ll push another PR.

PR #2738, opened 2024-07-17
allow attribute-only swap

stukennedy:
- hx-swap and hx-swap-oob to have an "attributes" value
- swap the attributes of the target element and not its children
- ensure hyperscript is processed after swap

## Description

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -779,6 +779,24 @@ return (function () {
             return false;
         }
 
+        function processHyperScript(elt) {
+            // Process hyperscript
+            if (typeof window['_hyperscript'] !== "undefined") {
+                window['_hyperscript'].processNode(elt);
+            }   
+        }
+
+        function swapAttributes(target, source) {
+            forEach(target.attributes, function (attr) {
+                target.removeAttribute(attr.name);
+            });
+            
+            // Add new attributes
+            forEach(source.attributes, function (attr) {
+                target.setAttribute(attr.name, attr.value);
+            });
+        }
+
         function cloneAttributes(mergeTo, mergeFrom) {
             forEach(mergeTo.attributes, function (attr) {
                 if (!mergeFrom.hasAttribute(attr.name) && shouldSettleAttribute(attr.name)) {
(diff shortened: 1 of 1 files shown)


schungx:
Actually this is very similar to the `class-tools` extension which swap CSS classes.

Maybe an `attr-tools` based on that?

MichaelWest22:
Yeah I think this would be awesome as an extension like this:
```javascript
htmx.defineExtension('attributes', {
  handleSwap: function(swapStyle, target, fragment) {
    if (swapStyle === 'attributes') {
      function forEach(arr, func) {
        if (arr) {
          for (let i = 0; i < arr.length; i++) {
            func(arr[i])
          }
        }
      }
      forEach(target.attributes, function (attr) {
        if(attr.name !== 'hx-ext') {
          target.removeAttribute(attr.name)
        }
      })
      // Add new attributes
      forEach(fragment.attributes, function (attr) {
        if(attr.name !== 'hx-oob-swap') {
          target.setAttribute(attr.name, attr.value)
        }
      })
      if (typeof window['_hyperscript'] !== "undefined") {
        window['_hyperscript'].processNode(elt)
      }
      return true
    }
  }
})
```
htmx has real simple alternative swap extension hook ins as you can see with the above code example
To use this just include this JS in your app to add the extension and then either add hx-ext="attributes" tag to your body or the item you want attributes swappable on and it should work the same as your intension.  Note that in my testing i found it added hx-swap-oob to things and removed hx-ext so added exceptions for this in my example.

Telroshan:
Yup this sounds more like what should be an extension to me. Even though it's somewhat arbitrary, the general idea is to only add changes that _feel core_ to the lib core, and add other, more specific or niche features as extensions. I would say we're in the latter case here.
Extensions should let you do almost the exact same thing given that htmx will check its extensions for the provided swap style if it's none of the default keywords, so implementing it as an extension will probably look very much like your added case in the swap style switch here.

You could publish your own extension repository and list it at https://github.com/bigskysoftware/htmx-extensions in the community extensions, see you there! 😁

stukennedy:
Awesome, thanks for the advice.
I'll create the extension instead.

Issue #2678, opened 2024-06-26
Websocket/SSE test server is not working


Bechma:
As you removed /src/ws.js and /src/sse.js , now the test server is not working. Also, the about pages are 404ing

I fix it in https://github.com/Bechma/htmx/commit/8fdcf14073a7235599c19e86ef32d48715c66d0d
If you are ok with the changes I can open a PR(or change the approach), otherwise, just letting you know :)

Telroshan:
Oh indeed, we forgot this one behind when we moved extensions to a [separate repository](https://github.com/bigskysoftware/htmx-extensions) 🤦 
Sure, bugfix PR welcome!

Telroshan:
See #3052

PR #2750, opened 2024-07-21
fix ws tests

Bechma:
## Description
Replace missing extension files to unpkg ones

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/2678

## Testing

## Checklist

  source changes)
  approved via an issue


test/ws-sse/static/site-base.html
```diff
@@ -2,8 +2,8 @@
 <link rel="stylesheet" href="/stylesheet.css">
 <title>&lt;/&gt; htmx Realtime Test Server</title>
 <script src="/htmx/htmx.js"></script>
-<script src="/htmx/ext/sse.js"></script>
-<script src="/htmx/ext/ws.js"></script>
+<script src="https://unpkg.com/htmx-ext-sse@2.0.0/sse.js"></script>
```
Telroshan, line 5: We're at version [2.2.1](https://github.com/bigskysoftware/htmx-extensions/blob/main/src/sse/package.json#L4) now

Maybe we should use `@latest` here? So we don't have to update the versions in this example as we release updates to these extensions in the future

Let me know what you think (objections welcome), otherwise at least let's use the latest version number here

Telroshan (changes requested):

alexpetros:
We ported these tests over to the extensions repo, so there's actually no need to have them here at all.

Telroshan:
@alexpetros this seems to be a complete suite to manually test WS & SSE and see how it works with htmx, which we don't have on the extensions repo that only contains the automated test suite for both WS & SSE

Maybe we should move that part to the extensions repo though?

alexpetros:
Yeah I think so—as long as people who are looking for it in core are properly signposted to the extensions repo.

PR #2752, opened 2024-07-22
Do not boost forms with method="dialog"

pokonski:
## Description

Boosting forms with method="dialog" should not trigger AJAX requests because these are only meant to close parent `<dialog>`  elements and not submit anything.

> When a `<form>` within a `<dialog>` is submitted via the dialog method, the dialog box closes, the states of the form controls are saved but not submitted

source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#usage_notes

## Testing

Can be tested with the following snippet:

```html

<body hx-boost="true">
  <dialog id="modal">
    <form method="dialog">
      <button>Close modal</button>
    </form>
  </dialog>
</body>

<script type="text/javascript">
   document.getElementById("modal").showModal()
</script>
```

Clicking the "Close modal" button should not trigger any AJAX requests

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2273,7 +2273,7 @@ var htmx = (function() {
    * @param {HtmxTriggerSpecification[]} triggerSpecs
    */
   function boostElement(elt, nodeData, triggerSpecs) {
-    if ((elt instanceof HTMLAnchorElement && isLocalLink(elt) && (elt.target === '' || elt.target === '_self')) || elt.tagName === 'FORM') {
+    if ((elt instanceof HTMLAnchorElement && isLocalLink(elt) && (elt.target === '' || elt.target === '_self')) || (elt.tagName === 'FORM' && String(getRawAttribute(elt, 'method')).toLowerCase() !== 'dialog')) {
       nodeData.boosted = true
       let verb, path
       if (elt.tagName === 'A') {
--- a/test/attributes/hx-boost.js
+++ b/test/attributes/hx-boost.js
@@ -72,7 +72,7 @@ describe('hx-boost attribute', function() {
   })
 
   it('does not boost forms with method="dialog"', function() {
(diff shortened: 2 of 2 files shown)


Telroshan:
Hey, while I perfectly understand the concern of following the specs, the issue is always about retro-compatibility ; as you've noticed yourself on PR #1867 and related issue #1866, since it's been merged, at least the OP and likely other persons have been running this setup, which would suddenly not work anymore as they would expect it to make the request with the fallback method specified.

That's exactly the kind of stuff that led us to make htmx 2 ; making breaking changes, changing some defaults to either fit the specs, or set new defaults that felt more logical (security ones for ex such as `selfRequestsOnly`).

And now that htmx 2 is officially released, it would take an htmx 3, which I'll be honest about, is very, very unlikely to happen anytime soon (if at all).
So, for this kind of change, I'm afraid you'd have to go for an opt-in approach, i.e. leave the default behavior as is and provide a configuration option to make `dialog` skip submit altogether.

I know it can be frustrating, I've [been there](https://github.com/bigskysoftware/htmx/issues/1514), I hope you'll understand!

pokonski:
Hey @Telroshan ! this fix won't break #1867 since that one uses explicit hx-verb attribute. Hx-boosting is a separate logic from my brief experience with the issue

Telroshan:
Oh ok! I misunderstood indeed, doesn't cover the same situation.

Then, if we didn't have any test for it, I'd consider this as an _undocumented_ behavior, thus sounds fine to directly fix to me, as it's no more a breaking change as it would have been otherwise.
Take that with a grain of salt though as I'm not the decision-maker here! Wouldn't want to give false hope either

pokonski:
> I misunderstood indeed, doesn't cover the same situation.

No worries, I confused myself while doing this fix 😁

Issue #2554, opened 2024-05-18
hx-boosted forms issue requests even when `method="dialog"` or `formmethod="dialog"`


PizzasBear:
As I understand it, hx-boost is meant only to replace standard browser requests with htmx ajax ones.
When using `method="dialog"` or `formmethod="dialog"` the browser doesn't issue a request, so hx-boost shouldn't to intervene.

Example:
```html
<dialog open>
    <form hx-boost="true" method="POST" action="create">
        <h1>Create a new thing</h1>
        <p>
            <label>
                Name:
                <input type="text" name="name">
            </label>
        </p>
        <p>
            <button>Create</button>
            <button formmethod="dialog" formnovalidate>Cancel</button>
        </p>
    </form>
</dialog>
```
When the cancel button is pressed, this will issue a request contrary to browser behaviour.

fiik346:
You just use `hx-boost="false"` on the dialog

Tortoaster:
There was a recent-ish related issue on the topic of `formmethod="dialog"` (#1866), not specific to boosted forms. The linked pull request changed the way HTMX handles it by having it send a POST request rather than an invalid DIALOG request. While that is an improvement of sorts, I think it would be more intuitive to follow the spec and close the dialog without sending a request.

PR #2753, opened 2024-07-22
Properly handle buttons with formmethod=dialog

geoffrey-eisenbarth:
## Description
Buttons with `formmethod` specified should only be handled by HTMX if the value is an HTTP verb. This partially addresses Issue #2554 (along with PR #2752) and (imho) fixes the behavior introduced in PR #1867. If a button with `formmethod="dialog"` is a clicked, HTMX should ignore the request and let the browser handle closing the dialog (see [this comment](https://github.com/bigskysoftware/htmx/pull/1867#issuecomment-2242733913)).

## Testing
I removed two ajax tests added by PR #2752 (that were also duplicated in the shadowdom tests) and replaced them with a test that should be checking that the server had no response. The tests passed but I'm not 100% the test were correctly written.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4077,9 +4077,11 @@ var htmx = (function() {
 
       const buttonVerb = getRawAttribute(submitter, 'formmethod')
       if (buttonVerb != null) {
-      // ignore buttons with formmethod="dialog"
-        if (buttonVerb.toLowerCase() !== 'dialog') {
+        if (VERBS.includes(buttonVerb.toLowerCase())) {
           verb = (/** @type HttpVerb */(buttonVerb))
+        } else {
+          maybeCall(resolve)
+          return promise
         }
       }
     }
--- a/test/core/ajax.js
+++ b/test/core/ajax.js
@@ -1178,32 +1178,16 @@ describe('Core htmx AJAX Tests', function() {
     values.should.deep.equal({ t1: 'textValue', b1: ['inputValue', 'buttonValue'], s1: 'selectValue' })
   })
 
-  it('handles form post with button formmethod dialog properly', function() {
-    var values
+  it('properly handles clicked buttons with a dialog formmethod', function() {
+    var request = false
     this.server.respondWith('POST', '/test', function(xhr) {
-      values = getParameters(xhr)
-      xhr.respond(200, {}, '')
+      request = true
+      xhr.respond(200, {}, '<button>Bar</button>')
     })
-
     make('<dialog><form hx-post="/test"><button id="submit" formmethod="dialog" name="foo" value="bar">submit</button></form></dialog>')
-
     byId('submit').click()
     this.server.respond()
-    values.should.deep.equal({ foo: 'bar' })
-  })
-
-  it('handles form get with button formmethod dialog properly', function() {
-    var responded = false
-    this.server.respondWith('GET', '/test', function(xhr) {
-      responded = true
-      xhr.respond(200, {}, '')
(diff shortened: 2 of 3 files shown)


Telroshan:
Hey, I'm gonna shamelessly copy-paste the message I left on the other PR, that actually doesn't apply to it but applies here!

> Hey, while I perfectly understand the concern of following the specs, the issue is always about retro-compatibility ; as you've noticed yourself on PR #1867 and related issue #1866, since it's been merged, at least the OP and likely other persons have been running this setup, which would suddenly not work anymore as they would expect it to make the request with the fallback method specified.
>
> That's exactly the kind of stuff that led us to make htmx 2 ; making breaking changes, changing some defaults to either fit the specs, or set new defaults that felt more logical (security ones for ex such as `selfRequestsOnly`).
>
> And now that htmx 2 is officially released, it would take an htmx 3, which I'll be honest about, is very, very unlikely to happen anytime soon (if at all).
So, for this kind of change, I'm afraid you'd have to go for an opt-in approach, i.e. leave the default behavior as is and provide a configuration option to make `dialog` skip submit altogether.
>
> I know it can be frustrating, I've [been there](https://github.com/bigskysoftware/htmx/issues/1514), I hope you'll understand!

geoffrey-eisenbarth:
Perhaps for the purposes of discussion, @SamDudley could chime in here and help us understand the intent of his PR? Given their example below, 

```
<dialog>
    <form hx-post="/test">
        <button formmethod="dialog" name="foo" value="bar">Submit</button>
    </form>
</dialog>
```
Is this supposed to 1) submit the form (via htmx) and then 2) close the dialog (via user agent)? Or something else?

geoffrey-eisenbarth:
I completely understand not wanting to break backwards compatibility, and would be willing to add a config value if that's deemed an acceptable path forward.

Thanks!

SamDudley:
Hey, it's been awhile since I worked on the project that prompted my PR, but I'll try to recall my intentions.

In the project I was working on, I wanted the form to be submitted (via htmx) and then for the dialog to be closed. I seem to recall that closing the dialog required some additional code, so it's possible the example would only submit the form (via htmx) and not close the dialog (testing required).

Having read this [comment](https://github.com/bigskysoftware/htmx/pull/1867#issuecomment-2242733913), and looking through the MDN docs myself, I agree that the behaviour I introduced is inconsistent with the MDN docs (and I assume the HTML spec).

It's interesting to think about what approach people would expect `htmx` to take. Should the `hx-post` (like in my example) take precedence and a form submission be made? or should the `formmethod` behaviour be retained and the request ignored?

Either way, personally I have no issues with this PR changing the behaviour to be consistent with the HTML spec, however in the interest of maintaining backwards compatibility, making the behaviour opt-in seems like the most sensible approach.

geoffrey-eisenbarth:
Just spitballing here, but it seems to me like the "correct" way to submit a form (via hx-post) and then close the dialog would be to do something like

```
<dialog>
    <form hx-post="/test" hx-on:htmx:after-swap="event.target.closest('dialog').close()">
        <button name="foo" value="bar">Submit</button>
    </form>
</dialog>
```

This way people who use HTMX as progressive enhancement only can still have nice no-js behavior:

```
<dialog hx-boost="true">
  <form method="post">
    [...]
    <button type="submit" name="foo" value="bar">Submit</button>
    <button formmethod="dialog">Close</button>
  </form>
</dialog>
```

PR #2768, opened 2024-07-29
Add ability to trigger an event on another element using `HX-Trigger` response header

bencroker:
## Description

This PR adds the ability to trigger an event on another element using the [`HX-Trigger`](https://htmx.org/headers/hx-trigger/) response header (also `HX-Trigger-After-Swap` and `HX-Trigger-After-Settle`) by allowing a `target` argument.

```
HX-Trigger: {"myEvent": {"target": "#myElement"}}
```

## Testing

Added a test to ensure the event is triggered on the `target` element.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1952,11 +1952,11 @@ var htmx = (function() {
       for (const eventName in triggers) {
         if (triggers.hasOwnProperty(eventName)) {
           let detail = triggers[eventName]
-          if (!isRawObject(detail)) {
-            detail = { value: detail }
-          } else {
+          if (isRawObject(detail)) {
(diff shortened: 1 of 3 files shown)


gijsleussink:
Beautiful improvement. 👍  In combination with takeClass it gives the opportunity to "activate" an element after a network call. In my case a list item.

What if the target element cannot be found? I have a list of 60k items (i.e. keywords in a dictionary) that are not all loaded on first request.
Shouldn't absence of the target element be handled? I think I now get a strange error.

bencroker:
> I think I now get a strange error. 

What error?

Issue #2755, opened 2024-07-23
Trigger modifier "from" prevents default behaviour of some input fields


pfiadDi:
Hi,

I came across a bug with versions 1.9.X and 2.X. When I use `hx-trigger="click from:body"` or more useful ones like `hx-trigger="click[event.target.matches('button#btnid')] from:body"`, any checkboxes can't be selected anymore, and any input field of type date doesn't show a datepicker flyout anymore.

The moment you remove the modifier, the input fields work as expected.

Here is an example code to reproduce:
EDIT: [link to working JS Fidddle](https://jsfiddle.net/et4j60bp/)
Thank you
Br

**EDIT:
Better description after investigation:**
- htmx prevents the default behaviour after a click on a form 
- if a form is triggered by a click modified by from:body (e.g. click from:body, click[event.target.matches('button#dynamicallyAddedElement')] from:body) not only the form def. behaviour is prevented, also that of the clicked element.  
- That means, datepicker flyouts for input type dates don't work, checkboxes can't be selected etc.

Telroshan:
Hey, I'm not sure I understand the issue here.

See [this JSFiddle](https://jsfiddle.net/3h4rgkq5/), I copied your example above and added for ex a `<div hx-get="/test"  hx-trigger="click[event.target.matches('button#btnid')] from:body">`

When I click the button, a request is correctly fired, and the date input + checkbox correctly work for me, did I miss something?

pfiadDi:
Hey @Telroshan , I am sorry I haven't seen your reply earlier. And my example code was trashed by github.
This was my original example:
https://jsfiddle.net/et4j60bp/

I can confirm, when a div is triggered it works but in my example I tried to trigger a form - in this case it doesn't work.

(in my app I have fields outside of a form and store selected values in a form with hidden inputs - after click on some elements I want to trigger the form) in this case the bug arises.

Telroshan:
Ooooh ok I understand now, thanks for the JSFiddle @pfiadDi.

Seems to be an unhandled specifity of the `from` modifier indeed ; I can see in htmx's code that the function determining whether an event should be default prevented checks this:
https://github.com/bigskysoftware/htmx/blob/0f70de7c0f8da72b3667fa7aef44636e51caf94e/src/htmx.js#L2307-L2315

The issue here being that the element against which the tag name is checked, is the `form` itself and not the element that triggered the event (which would here be the date or checkbox input).
[...]

PR #2771, opened 2024-07-29
Bug-Fix: Modified click trigger on form elements prevent default behaviour of clicked element

pfiadDi:
## Description
The `shouldCancle` function prevents the default behaviour for form elements when clicked. The problem is, when a dynamically added element should trigger the form and the user adds something like: `hx-trigger="click[event.target.matches('button#dynamicallyAddedElement')] from:body"` the default behaviour of other elements are prevented as well. 

A checkbox can't be checked anymore for example. 

I added a check so that we only cancel the event when the node with the trigger is a form AND the event target is also one:
Old:
` if (elt.tagName === 'FORM') {
        return true
 }`

New:
` if (elt.tagName === 'FORM' && asElement(evt.target)?.tagName === 'FORM') {
        return true
 }`

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/2755

## Testing
Manually:
- Added an input field type date and a form with hx-trigger click from:body
- Clicked on the date picker symbol and got the expected datepicker flyout

New tests:
- added a test with the same setup and checked in the event object if property "defaultPrevented" equals false (tbh wasn't sure where so I added it to re
- adapted the existing tests in internal to work with the adapted function
- added one test for a missing case in function "shouldCancel" 

## Checklist

  source changes)
  approved via an issue
* [] I ran the test suite locally (`npm run test`) and verified that it succeeded **-> YES and NO:**

I had troubles with the test suite also before working on the bug, I always got this result:
641 passing (9s)
  3 pending
  2 failing

  1) Core htmx Parameter Handling
       form includes last focused button:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

  2) Core htmx Parameter Handling
       form includes last focused submit:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

**After adding my fix and the new tests, my new tests were successful and the same two failed...**


Telroshan:
Hey,

- Could you rebase against `dev`? Seems like you are bringing some unrelated commits along, which makes the diff harder to review.
- I'm worried about this sentence
> adapted the existing tests in internal to work with the adapted function

As we don't want to introduce breaking changes to the lib, and as `shouldCancel` is exposed to the internal API (that any of the official, community of custom extensions could use), existing tests shouldn't need to be modified to use the new syntax. Ideally, the fallback should work as it did before in these tests, as if I understood correctly, this is about a bugged behavior that hadn't been identified since now, but the tested ones are fine

pfiadDi:
Hi @Telroshan thank you very much for your feedback. It's my first PR I highly appreciate it :) 
I'll work on that later today or tomorrow
Br

test/core/internals.js
```diff
@@ -94,15 +94,20 @@ describe('Core htmx internals Tests', function() {
...
+
+    htmx._('shouldCancel')({ type: 'click', target: document.createElement('form') }, form).should.equal(true)
+
+    // function works also when target isn't an Element
+    htmx._('shouldCancel')({ type: 'click', target: null }, form).should.equal(false)
 
     var form = make("<form><input id='i1' type='submit'></form>")
     var input = byId('i1')
-    htmx._('shouldCancel')({ type: 'click' }, input).should.equal(true)
+    htmx._('shouldCancel')({ type: 'click', target: document.createElement('input') }, input).should.equal(true)
```
Telroshan, line 106: Shouldn't we use the `input` var here though instead of creating a dummy element on the fly? As `input` is the clicked element, the event's target would be that very element in a real situation right?

Same goes for the `form` above, and the `button` after

src/htmx.js
```diff
@@ -2313,7 +2313,7 @@ var htmx = (function() {
       return false
     }
     if (evt.type === 'submit' || evt.type === 'click') {
-      if (elt.tagName === 'FORM') {
+      if (elt.tagName === 'FORM' && asElement(evt.target)?.tagName === 'FORM') {
```
Telroshan, line 2316: I'm wondering ; couldn't we simply check `if (asElement(evt.target) || elt).tagName === 'FORM'` in this case?

So that, if defined, `evt.target` takes precedence, otherwise we keep the existing behavior. You would probably not even need to add the `target` property to mock the events of the test suite in this case? As `elt` is the fallback

Btw, should we use the same technique for the other checks in this same function? Anchors for ex. Maybe we should simply change to`const elt = asElement(evt.target) || asElement(node)` at the start of the function and not need to modify anything else inside?
Just some thoughts, completely untested, let me know!

Telroshan (commented):
Hey @pfiadDi , seems the rebase went wrong as other unrelated commits are still here 😬 

Got it, you're right, didn't notice these tests were incorrectly mocking an event's properties.

test/core/internals.js
pfiadDi: Yes you're right

src/htmx.js
pfiadDi: @Telroshan the problem is, the event target is an [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) and an EventTarget can be a Dom element but doesn't have to be. So the property tagname can be undefined. (It was my first approach but the type checker complaint). That's why I chose this way to ensure we only check properties which exist and don't run into any errors when trying to access undefined properties

Telroshan: > an EventTarget can be a Dom element but doesn't have to be. So the property tagname can be undefined.

Hence my suggestion @pfiadDi : `const elt = asElement(evt.target) || asElement(node)`
Unless I'm mistaked (could totally be, didn't test this 😆 ), this will get the `evt.target` if it's a DOM element, but if it's not an element, will fallback to `node`.
Then we have the already present `if` statement that returns early, in case neither `evt.target` nor `node` were elements.
I believe you shouldn't get any type checker complaint with that approach as we cast to `Element` thanks to `asElement`?

MichaelWest22:
Resolved in #3336 and now in 2.0.5

PR #2779, opened 2024-07-31
HX-Redirect causing indicators to clear before redirect

webdevcody:
not sure if this would break stuff, but I was seeing issues with the indicators clearing before the redirect finished.  See video for issue.

https://github.com/user-attachments/assets/a24f750b-285e-4394-b66d-9183dc8a7e87

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4343,7 +4343,9 @@ var htmx = (function() {
         const hierarchy = hierarchyForElt(elt)
         responseInfo.pathInfo.responsePath = getPathFromResponse(xhr)
         responseHandler(elt, responseInfo)
-        removeRequestIndicators(indicators, disableElts)
+        if (!hasHeader(xhr, /HX-Redirect:/i)) {
+          removeRequestIndicators(indicators, disableElts);
+        }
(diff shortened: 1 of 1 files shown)


Telroshan:
Hey, seems you'll want to run `npm format` to properly lint your code.

Other than that, I'm wondering what this `pageshow` event handling is about, is it necessary in this case? What issue does it solve?

webdevcody:
@Telroshan Here is the issue without the pageshow


https://github.com/user-attachments/assets/a97d7af3-ade4-4ca5-9b8c-07d3da9c4912

jadamita:
> @Telroshan Here is the issue without the pageshow. Basically the browser is going to cache the state of the previous page, so since we do not ever re-enable the button or hide the indicator, it'll show when the user clicks back.
> 
>  pageshow.mp4

off topic, but what software was used for that recording? it's buttery smooth.

JonathanFajobi:
@jadamita

https://www.screen.studio/

1cg:
i think I already implemented this, based on @webdevcody's feedback, thank you!

PR #2781, opened 2024-08-01
Optimize Head regex

MichaelWest22:
## Description
Their was a change to add a regex tag utility function in #2024 and this combined three regex creations into a single function.  Since then it seems the svg and title regex were simplified out with the htmx 2.0 re-write and we are now left with a redundant function that is only used to generate a single HEAD_TAG_REGEX which is only used once in makeFragment.  So my proposed change here is to just simply remove this utility function and move the single use regex inline which should simplify readability and code minification.

Previous regex format was:

``` javascript
new RegExp(`<${tag}(\\s[^>]*>|>)([\\s\\S]*?)<\\/${tag}>`, global ? 'gim' : 'im')
```
And I'm simplified this to:
``` javascript
/<head(\s[^>]*)?>.*?<\/head>/is
```
Note [\s\S] is now .* because of the s option and m option is not needed here.  

Corresponding issue:

## Testing
Found the manual tests added with the #2024 change and was able to manually test this change worked fine with these manual tests after updating the test to link to the now external head-support extension.
Also tested with `npm run test` but found that this remove head feature in makeFragment has no coverage in these tests as breaking it caused no test failures.  Because it is only used by head-support extension and not core htmx.
Manually ran the manual-tests folder in the head-support extension with the updated htmx changes as well and found no issues.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -583,7 +583,7 @@ var htmx = (function() {
    */
   function makeFragment(response) {
     // strip head tag to determine shape of response we are dealing with
-    const responseWithNoHead = response.replace(/<head(\s[^>]*>|>)(.*?)<\/head>/ism, '')
+    const responseWithNoHead = response.replace(/<head(\s[^>]*)?>.*?<\/head>/is, '')
     const startTag = getStartTag(responseWithNoHead)
     /** @type DocumentFragmentWithTitle */
     let fragment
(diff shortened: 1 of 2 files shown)


Telroshan:
Note: I'm labeling this as `performance` as it doesn't make any functional change, but contributes in decreasing the library's size before & after minification

alexpetros:
Sure, looks good

PR #2799, opened 2024-08-06
:bug: Prevents erroring on null vals

ekwoka:
## Description

This prevents accessing the `forEach` key on `null` or `undefined` values. `null` is valid in JSON, so it should be supported by the `hx-vals` attribute. This is a regression from 1.x

Could use optional chaining, but I'd imagine that isn't an accepted feature. I also assume the fact it doesn't use `Array.isArray` is because it wants to support other containers that implement `forEach`. So just doing a truthy test is satisfactory, since there is 0% chance that someone is going to be passing `document.all` as a value.

Corresponding issue:

## Testing

Added test. Change is common sense, but the test ensures it's condition into the future.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3927,7 +3927,7 @@ var htmx = (function() {
     const formData = new FormData()
     for (const key in obj) {
       if (obj.hasOwnProperty(key)) {
-        if (typeof obj[key].forEach === 'function') {
+        if (obj[key] && typeof obj[key].forEach === 'function') {
           obj[key].forEach(function(v) { formData.append(key, v) })
         } else if (typeof obj[key] === 'object' && !(obj[key] instanceof Blob)) {
           formData.append(key, JSON.stringify(obj[key]))
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4020,7 +4020,7 @@ var htmx = (function() {
           return false
         }
         target.delete(name)
-        if (typeof value.forEach === 'function') {
+        if (value && typeof value.forEach === 'function') {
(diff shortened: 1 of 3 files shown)


Telroshan:
Hey, good catch!
You'll probably also want to add this check for formDataProxy's `set` method, where we don't check if value is truthy at all either?
https://github.com/bigskysoftware/htmx/blob/eb32358873a7b5e2b232e27232ad4ba6ae85b348/src/htmx.js#L4004-L4015

ekwoka:
Implemented that. I couldn't see easily how to do a test as sufficiently high level for how the tests are set up of that.

Telroshan:
@ekwoka the `formDataProxy` is used for example by `getInputValues`, which returns it in its `values`  property.
`getInputValues` is itself exposed in the internal API and I seem to recall we have a bunch of tests for this function specifically, so I think you could add one here by using the proxy syntax to do something like `result.values["test"] = null`?
Didn't test this though, you would want to make sure first that this indeed throws an error without your fix

ekwoka:
Yes, I will try to identify it. I just couldn't really tell where a proper test of that feature would be or what it would look like. Thanks for the feedback, I will try to identify that spot, validate the error, and add the test.

ekwoka:
Okay! Got it! I removed the fix on the formdata proxy, validated that this test fails (from the throw), and that the fix works.

ekwoka:
Any updates or concerns?

1cg:
nope just me moving slow, thank you!

PR #2802, opened 2024-08-07
Better graceful degradation of boosted form element

NDoolan360:
## Description

The expected behaviour of a form is not currently adhered to in htmx, making hx-boost degrade in an unexpected way for forms. (I have only tested this on chrome, safari and firefox)

When the action attribute is not set on a form with hx-boost and there are query params in the url, these values are included in the request as well as any values from the form. This does not match how forms normally handle an empty action attribute without hx-boost.

Another attempt at https://github.com/bigskysoftware/htmx/pull/1811

Corresponding issue: #1788

## Testing
I have tested this manually in one of my own projects. 
I'm happy to implement anything in the htmx test suite but unsure how to mock query params in url in the tests I looked into.

## Checklist

  source changes)
  approved via an issue
* [x] I ran the test suite locally (`npm run test`) and verified that it succeeded - `652 passing (5s), 3 pending`


NDoolan360:
Hmmm, thinking there may need to be another check that the method is "GET" for this instance.
The [form submission algorithm (step 26 table)](https://html.spec.whatwg.org/#form-submission-algorithm) outlines the different schemes for how to handle the url.
For http/https GET requests, this is ["Mutate action URL"](https://html.spec.whatwg.org/#submit-mutate-action), where notably:
> Let query be the result of running the application/x-www-form-urlencoded serializer with pairs and encoding.
>
>Set parsed action's query component to query.

This seems to imply the existing query component is overriden.

Telroshan:
Hey, could you add a test case to the test suite, to first confirm the issue with the lib's current state, then confirm that your PR fixes it? (You might also want to add a test case for POST)
Thanks!

NDoolan360:
Hey @Telroshan, I've added some tests.
Please let me know if there's more you need. 😄

Issue #2790, opened 2024-08-04
`hx-swap-oob` strips away the outer element when used with `beforeend`


EmrysMyrddin:
Hi, I'm facing an issue with Out Of Band swapping feature. Perhaps I'm miss understanding how it works, in such case the documentation is not very clear about it.

My use case is the following:

```html
<ul id="list-1">
  <li hx-post="/move/item-1/list-2" hx-swap="outerHTML">
    <div class="item">Item 1</div>
  </li>
</ul>
<ul id="list-2">
</ul>
```

I have 2 lists, and want to be able to move items from one list to the other.

The response of the `/move/item-1/list-2` is the following:
```html
<li hx-swap-oob="beforeend:#list-2"  hx-post="/move/item-1/list-2"  hx-swap="outerHTML">
  <div class="item">Item 1</div>
</li>
```

My expectation is to have this resulting DOM:
```html
<ul id="list-1">
</ul>
<ul id="list-2">
  <li hx-post="/move/item-1/list-2"  hx-swap="outerHTML">
    <div class="item">Item 1</div>
  </li>
</ul>
```

It kind of work, since there is not other element than the OOB one, the first `<li>` is swapped away, but the new expected `<li>` is broken:
```html
<ul id="list-1">
</ul>
<ul id="list-2">
  <div class="item">Item 1</div>
</ul>
```

For some reason, the `<li>` element which has the `hx-swap-oob` attribute is stripped away.

Is this an expected behavior ? It seems to be specific to `before*` and `after*` strategies. If it is, it should be documented.

I've worked around it for now by just wrapping the `<li>` into a `<div>`, moving the `hx-swap-oob` attribute to the `<div>`:

```html
<div hx-swap-oob="beforeend:#list-2" >
  <li hx-post="/move/item-1/list-2"  hx-swap="outerHTML">
    <div class="item">Item 1</div>
  </li>
</div>
```

This gives the expected result:
```html
<ul id="list-1">
</ul>
<ul id="list-2">
  <li hx-post="/move/item-1/list-2"  hx-swap="outerHTML">
    <div class="item">Item 1</div>
  </li>
</ul>
```

MichaelWest22:
From my quick testing it seems that for normal swaps the html fragment is processed by makeFragment() and wrapped in a body tag as it is turned into html document fragment.  This is then passed to all the different swapping functions and handled as expected.  

[...]

PR #2806, opened 2024-08-08
clear up oob swap expected behaviour

MichaelWest22:
## Description
Try to clarify the expected behavior of oob swap strategies which work a little different to normal swaps.  all the non inline swap strategies like beforeend etc seem to work differently than they do with full swaps in that it is the inner contents of the tag supplied with the hx-swap-oob attribute that is swapped in and not the tag itself.  This gives you the flexibility to swap in non tag wrapped data if needed and to not add additional hx-swap-oob tags into the DOM that have no client DOM function but it does mean you have to add an additional layer by wrapping the oob content in a dummy div tag sometimes.

Corresponding issue:
#2790 

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/attributes/hx-swap-oob.md
+++ b/www/content/attributes/hx-swap-oob.md
@@ -34,7 +34,7 @@ If another swap value is given, that swap strategy will be used instead but note
  ...
 </div>
 <div id="alerts" hx-swap-oob="beforeend">
-    <p>Saved again!</p> <!--swap inner contents to end of alters but not wraping oob div-->
+    <p>Saved again!</p> <!--swap inner contents to end of alerts but not wraping oob div-->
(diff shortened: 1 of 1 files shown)


EmrysMyrddin:
Thank you for taking care of this!

I think we need a very short example to demonstrate how it works with other strategies. I understand what you've added because I know the issue, not sure I would have understood it the first time a read the documentation :-)

MichaelWest22:
Added a couple of examples of other oob syntax

EmrysMyrddin (commented):
That's cool ! I think it will help understand the behavior better :-)

MichaelWest22:
Clossing as #2889 merged with better documentation.

Issue #2264, opened 2024-01-31
Running ready function on readyState === 'interactive'


andrew-ebsco:
Hello htmx team!

I have been working on a confusing bug that I was able to fix, but want to present it to the team before making a pull request. I have started using htmx in [Magento](https://github.com/magento/magento2) for some custom features. Magento uses RequireJS to load AMD modules and htmx runs well in the set up. However, RequireJS loads all modules with `async=""` on all script tags. The bug I ran into is that the htmx initialization function was not running sometimes. Since the document body was not being processed, it required a page refresh to get the htmx behavior to execute. After some debugging, I found the issue was in how the `ready` function checks for a complete DOM.

```js
var isReady = false
getDocument().addEventListener('DOMContentLoaded', function() {
    isReady = true
})

/**
 * Execute a function now if DOMContentLoaded has fired, otherwise listen for it.
 *
 * This function uses isReady because there is no realiable way to ask the browswer whether
 * the DOMContentLoaded event has already been fired; there's a gap between DOMContentLoaded
 * firing and readystate=complete.
 */
function ready(fn) {
    // Checking readyState here is a failsafe in case the htmx script tag entered the DOM by
    // some means other than the initial page load.
    if (isReady || getDocument().readyState === 'complete') {
        fn();
    } else {
        getDocument().addEventListener('DOMContentLoaded', fn);
    }
}
```

In the above code, there is a scenario where `fn` is never executed. Sometimes the htmx code was being executed after the `DOMContentLoaded` event had fired, but before `document.readyState === 'complete'`. After doing some research, I found that `DOMContentLoaded` can fire when `document.readyState === 'interactive'`. According to both [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) and the [HTML Spec](https://html.spec.whatwg.org/multipage/dom.html#reporting-document-loading-status), the `DOMContentLoaded` event and a readyState of interactive are equivalent.

As a result, my proposal is to change the `ready` function to check for both a readyState of interactive as well as complete.
```js
var isReady = false
getDocument().addEventListener('DOMContentLoaded', function() {
    isReady = true
})

/**
 * Execute a function now if DOMContentLoaded has fired, otherwise listen for it.
 *
 * This function uses isReady because there is no realiable way to ask the browswer whether
 * the DOMContentLoaded event has already been fired; there's a gap between DOMContentLoaded
 * firing and readystate=complete.
 */
function ready(fn) {
[...]

PR #2808, opened 2024-08-09
Checking readyState is interactive before listening for DOMContentLoaded

andrew-ebsco:
## Description
This is to fix issue #2264. Sometimes DOMContentLoaded can fire before readyState === 'complete'.

Corresponding issue:

## Testing
The unit test suite passed.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4899,7 +4899,7 @@ var htmx = (function() {
   function ready(fn) {
     // Checking readyState here is a failsafe in case the htmx script tag entered the DOM by
     // some means other than the initial page load.
-    if (isReady || getDocument().readyState === 'complete') {
+    if (isReady || getDocument().readyState === 'interactive' || getDocument().readyState === 'complete') {
       fn()
     } else {
       getDocument().addEventListener('DOMContentLoaded', fn)


src/htmx.js
```diff
@@ -4899,7 +4899,7 @@ var htmx = (function() {
   function ready(fn) {
     // Checking readyState here is a failsafe in case the htmx script tag entered the DOM by
     // some means other than the initial page load.
-    if (isReady || getDocument().readyState === 'complete') {
+    if (isReady || getDocument().readyState === 'interactive' || getDocument().readyState === 'complete') {
```
Telroshan, line 4902: Shouldn't we simply check `readyState !== 'loading'` in this case?

alexpetros: We've tried that—I know this looks a little clunky but I'd rather both checks be included.

Telroshan: > We've tried that

Sorry but this confuses me @alexpetros , are you suggesting this might result in a different behavior than `!== 'loading'`?
As it's a [3 values](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) property, this should be the exact same behavior, except that this adds more hardcoded strings not to be minified

alexpetros: Ah, I didn't read your suggestion closely enough (I missed the `!==`, that's my bad). Yes @andrew-ebsco can you make that change?

Telroshan: @alexpetros just to emphasize on this, this would take us back to august 2023, aka the first time we started dealing with these ready state / load event issues.
See #1672 which was the first one about this matter ; we precisely had a `!== 'loading'` check, exactly what we would go back to with this PR.
See my [comment in #2264](https://github.com/bigskysoftware/htmx/issues/2264#issuecomment-2282671789) with the history of PRs and issues related to this topic.

I'm not sure that's really what we should go for here, as we're getting back to a previous state that we _already know_ is going to cause issues, as those very issues have been reported in the past (again, cf the issues and PRs listed in my comment, linked above)

alexpetros: > @alexpetros just to emphasize on this, this would take us back to august 2023, aka the first time we started dealing with these ready state / load event issues.

I don't think is true, because what we have now that we didn't august 2023 is a variable that captures [whether DOMContentLoaded fired](https://github.com/bigskysoftware/htmx/blob/51807d094abd5bcf6ec0db33e08e8733b5afbb0d/src/htmx.js#L4887-L4890).

Telroshan: Hmm I'm not sure this variable is useful anymore if we roll back to a `!== 'loading'` check @alexpetros 

When you added that `ready` variable in #1688, it was precisely because we were only checking `=== 'complete'`, thus we could miss the timeframe between `interactive` and `complete` state, where we would bind to `DOMContentLoaded` while the document was already `interactive`, thus the listener would never be called.

If we roll back to a `!== 'loading'` check, I don't think there's any gap between the capture you linked above, and the first `ready` call of the lib.
Between this https://github.com/bigskysoftware/htmx/blob/51807d094abd5bcf6ec0db33e08e8733b5afbb0d/src/htmx.js#L4887-L4890

And this https://github.com/bigskysoftware/htmx/blob/51807d094abd5bcf6ec0db33e08e8733b5afbb0d/src/htmx.js#L4938-L4939

No async work is performed, so the `DOMContentLoaded` listeners would be bound one after the other, in the same frame.

So either the `DOMContentLoaded` event hasn't been fired yet, and both the capture and `ready` will bind a listener to it, making the capture useless here
Or, the event has already fired when htmx loads, so ready executes right away (as the state is `interactive` and no more `loading`), making the capture also useless here.

As per my [comment in #2264](https://github.com/bigskysoftware/htmx/issues/2264#issuecomment-2288115312) I really think we should take some time to dive in this topic again @alexpetros with @1cg as well, to see if A) we still agree with our past selves about the init function, and B) try to setup reproducible cases to test the different loading scenarios

alexpetros:
Closing for the reasons well-explained by Micheal here: https://github.com/bigskysoftware/htmx/pull/3365#issuecomment-3065080028

PR #2824, opened 2024-08-16
Minor change in the documentation

FeatureSpitter:
As a new user, I initially thought `"delay"` acted like a debouncer, waiting for the user to stop typing before sending a request. However, `throttle` is the correct approach for this scenario. Although the example doesn't focus on efficient event handling, changing it to `throttle` would prevent confusion and align with best practices, as `delay` simply postpones requests without debouncing. This update ensures clarity for new users.

Thanks!

### DIFF
--- a/www/content/docs.md
+++ b/www/content/docs.md
@@ -806,7 +806,7 @@ However, you could wrap the htmx-enhanced input in a form element:
     <input class="form-control" type="search"
         name="search" placeholder="Begin typing to search users..."
         hx-post="/search"
-        hx-trigger="keyup changed delay:500ms, search"
+        hx-trigger="keyup changed throttle:500ms, search"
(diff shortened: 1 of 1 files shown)


MichaelWest22:
see https://htmx.org/attributes/hx-trigger/ . delay does act as a debouncer as the delay resets its delay timer if the event is triggered during the delay interval.   Throttle just makes it not request too often but it will trigger after they type the first letter and then not trigger a second time as they type more characters till the throttle delay ends.  And when they start typing again it will trigger again on the first letter and then after the delay again.  So throttle would lead to twice as many search requests I think.

FeatureSpitter:
Oh I see. So delay is some kind of lazy debouncing and throttle is eager debouncing.

I'll close this PR then.

Thank you for clarifying.

Issue #2676, opened 2024-06-25
BUG: pre-submit custom form validation not applied to radiobuttons and checkboxes


ehenighan:
Found while testing a slightly niche use-case in a questionnaire form I'm building - when you apply custom validation across a form using the HTML5 hooks to enforce that the user has to fill out _at least one_ of a set of answers on a page (i.e. no particular radiogroup is required, but at least one of the visible radiogroups must have a value), the onsubmit validation isn't respected.

This is because the 'shouldInclude' function excludes unticked checkboxes/radiobuttons but the 'validateElement' check is inside the 'shouldInclude' block.

It's easy to fix and I've tested locally - just separated out 'couldInclude' and 'shouldValidate' from 'shouldInclude' in order to lift up the validation outside the 'shouldInclude' block, but wasn't sure where you'd want a PR targeting now there's version 2 and 1.9.12 on the go at the same time.

Will check back shortly to add code examples!

ehenighan:
[HTMX Bug #2676 - Replication.zip](https://github.com/user-attachments/files/15976516/HTMX.Bug.2676.-.Replication.zip)

Zip file contains a basic HTML file with two forms demonstrating the issue:
- First form has a number input and a text input and won't submit unless you set at least one of them
- Second form has a checkbox input and a radiogroup and will show the same error but (try to) submit in the background anyway

N.B. that the validation's broken after the first pass due to the fake POST that goes nowhere.

Will update with proposed fix shortly

ehenighan:
Forked and added the fix I've tested locally here, for comparison:

https://github.com/bigskysoftware/htmx/compare/dev...ehenighan:htmx:bug-2676-pre-submit-custom-validation

Haven't raised a PR yet as I'm neck-deep in a project at work and I'm mindful that I've not worked with this test suite before etc. Would you want the test going into test/core/validation.js next to the existing one about custom-validation? https://github.com/bigskysoftware/htmx/blob/6f83885de3e3881d262a6267fd55be69a18daa75/test/core/validation.js#L87

Telroshan:
Hey!
> wasn't sure where you'd want a PR targeting now there's version 2 and 1.9.12 on the go at the same time

I see you're using htmx 1.9.12 in your bug replication sample, would you mind checking if it also happens on htmx 2?

As it's a bug, and if it's occuring on both versions, I'd say htmx would benefit from the bugfix on both htmx 1 and 2.
[...]

PR #2829, opened 2024-08-19
Issue #2676 - Tests for v2 to prevent regression of issue from v1

ehenighan:
## Description
Added a basic test for radio input form submission, following the example for checkbox inputs.
Added tests for custom validation on radio and checkbox inputs when the inputs are empty, to support things like 'you must choose an option from at least one of these groups' etc.
These tests pass for v2 due to the refactoring of `processInputValue` in commit `b9322450ff42a5bdf25b194fcdb6ecc744353819` : https://github.com/bigskysoftware/htmx/commit/b9322450ff42a5bdf25b194fcdb6ecc744353819#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dR2528-R2530
 - however, it's not totally obvious reading that code why there are now two separate calls to `validateElement` under different conditions, so while I'll submit the main bugfix for https://github.com/bigskysoftware/htmx/issues/2676 under v1, I would like to make sure there's test coverage of the problem for any future refactoring applied to this function in v2.

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/2676

## Testing
As above - manually tested the fix in v1 using the replication sample attached to the ticket, wrote the tests and saw they failed when without the fix. Carried the fixes over to v2, saw that the tests passed now, checked why, and thought it best to submit the tests anyway.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/test/core/ajax.js
+++ b/test/core/ajax.js
@@ -559,6 +559,34 @@ describe('Core htmx AJAX Tests', function() {
     values.should.deep.equal({ c1: ['cb1', 'cb3'], c2: 'cb5', c3: 'cb6' })
   })
 
+  it('properly handles radio inputs', function() {
+    var values
+    this.server.respondWith('Post', '/test', function(xhr) {
+      values = getParameters(xhr)
+      xhr.respond(204, {}, '')
+    })
+
+    var form = make('<form hx-post="/test" hx-trigger="click">' +
+        '<div role="radiogroup">' +
+        '<input id="rb1" name="r1" value="rb1" type="radio">' +
+        '<input id="rb2" name="r1" value="rb2" type="radio">' +
+        '<input id="rb3" name="r1" value="rb3" type="radio">' +
+        '<input id="rb4" name="r2" value="rb4"  type="radio">' +
+        '<input id="rb5" name="r2" value="rb5"  type="radio">' +
+        '<input id="rb6" name="r3" value="rb6"  type="radio">' +
+        '</div>' +
+        '</form>')
+
+    form.click()
+    this.server.respond()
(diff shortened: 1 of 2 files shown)


test/core/ajax.js
```diff
@@ -559,6 +559,34 @@ describe('Core htmx AJAX Tests', function() {
     values.should.deep.equal({ c1: ['cb1', 'cb3'], c2: 'cb5', c3: 'cb6' })
   })
 
+  it('properly handles radio inputs', function() {
```
ehenighan, line 562: This is essentially just a cut-down clone of the function above which tests checkboxes in detail, to cover the `shouldInclude` logic which applies to both input types:

`if (elt.type === 'checkbox' || elt.type === 'radio') {
      return elt.checked
    }`

test/core/validation.js
```diff
@@ -102,6 +102,42 @@ describe('Core htmx client side validation tests', function() {
...
+    form.textContent.should.equal('No Request')
+    form.click()
+    this.server.respond()
+    form.textContent.should.equal('No Request')
+    byId('i1').setCustomValidity('')
+    form.click()
+    this.server.respond()
+    form.textContent.should.equal('Clicked!')
+  })
+
```
ehenighan, line 140: These two function are essentially clones of the previous function but applied specifically to the input types with special handling in `shouldInclude` (`if (elt.type === 'checkbox' || elt.type === 'radio') {
      return elt.checked
    }`) to ensure that custom validation isn't broken for unchecked items again in the future

Telroshan (approved):
Thanks for the details & the new test cases!

PR #2836, opened 2024-08-20
Ability to add `options` argument to event listener

bencroker:
This PR adds the ability to add an `options` (or `useCapture`) argument to the event listener when using `htmx.on()`.

As per the [`addEventListener()`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) method spec, the third argument accepts an [`options`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#options) object or a [`useCapture`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#usecapture) boolean.

Usage:

```js
// Using an object
htmx.on('htmx:afterSwap', () => console.log('Swapped'), { once: true }); 
htmx.on('#my-div', 'htmx:afterSwap', () => console.log('Swapped'), { once: true });

// Using a boolean
htmx.on('htmx:afterSwap', () => console.log('Swapped'), true); 
htmx.on('#my-div', 'htmx:afterSwap', () => console.log('Swapped'), true);
```

**Note to the reviewer:** please double check my JSDoc annotations, I think they’re accurate but I’m not 100% sure.

## Testing

I added a test that verifies that the `options` argument is respected.

https://github.com/bigskysoftware/htmx/blob/b266104d3ecaaf850450b0475c2c1f6d95c5f4ea/test/core/events.js#L80-L93

## Checklist

  source changes)
* [x] ~~This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue~~ This is a minor addition to an existing feature, not pre-approved 😬 

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1247,9 +1247,6 @@ var htmx = (function() {
    * @returns {EventArgs}
    */
   function processEventArgs(arg1, arg2, arg3, arg4) {
-    if (arg4 == null) {
-      arg4 = {}
-    }
     if (isFunction(arg2)) {
       return {
         target: getDocument().body,
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1242,8 +1242,8 @@ var htmx = (function() {
   /**
    * @param {EventTarget|AnyEventName} arg1
    * @param {AnyEventName|EventListener} arg2
-   * @param {EventListener|Object} [arg3]
-   * @param {Object} [arg4]
+   * @param {EventListener|Object|boolean} [arg3]
+   * @param {Object|boolean} [arg4]
    * @returns {EventArgs}
    */
   function processEventArgs(arg1, arg2, arg3, arg4) {
@@ -1274,8 +1274,8 @@ var htmx = (function() {
    *
    * @param {EventTarget|string} arg1 the element to add the listener to | the event name to add the listener for
    * @param {string|EventListener} arg2 the event name to add the listener for | the listener to add
-   * @param {EventListener|Object} [arg3] the listener to add | options to add
-   * @param {Object} [arg4] options to add
+   * @param {EventListener|Object|boolean} [arg3] the listener to add | options to add
+   * @param {Object|boolean} [arg4] options to add
    * @returns {EventListener}
    */
   function addEventListenerImpl(arg1, arg2, arg3, arg4) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1243,10 +1243,13 @@ var htmx = (function() {
    * @param {EventTarget|AnyEventName} arg1
    * @param {AnyEventName|EventListener} arg2
    * @param {EventListener|Object|boolean} [arg3]
-   * @param {Object|boolean} [arg4]
+   * @param {Object|boolean=} [arg4]
    * @returns {EventArgs}
    */
   function processEventArgs(arg1, arg2, arg3, arg4) {
+    if (arg4 == null) {
+      arg4 = {}
+    }
     if (isFunction(arg2)) {
       return {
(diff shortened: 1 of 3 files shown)


src/htmx.js
```diff
@@ -1236,26 +1236,33 @@ var htmx = (function() {
...
    * @param {AnyEventName|EventListener} arg2
-   * @param {EventListener} [arg3]
+   * @param {EventListener|Object|boolean} [arg3]
+   * @param {Object|boolean} [arg4]
    * @returns {EventArgs}
    */
-  function processEventArgs(arg1, arg2, arg3) {
+  function processEventArgs(arg1, arg2, arg3, arg4) {
+    if (arg4 == null) {
+      arg4 = {}
```
Telroshan, line 1251: Is this necessary? Doesn't addEventListener gracefully handle a null (as well as undefined) value?

test/core/events.js
```diff
@@ -77,6 +77,21 @@ describe('Core htmx Events', function() {
     }
   })
 
+  it('events accept an options argument', function() {
```
Telroshan, line 80: Would you mind adding a test case for the boolean `useCapture` counterpart to `options` as well? Just to make sure we support both cases and prevent any future regression on this

Telroshan (commented):
JSDoc seems perfectly fine to me!

src/htmx.js
bencroker: You’re right, it does! Removed that redundant code, thanks.

Telroshan:
Note that I'll remove the "ready for review" label for now; despite its generic name, we actually use this label just for 1cg, to tag PRs that are ready for his final review, meanwhile you can simply ping me for any update!

Hit me up when you figure out this `useCapture` test! At first glance, I'd expect the `event.target` to always equal the element that the event was triggered onto, no matter if `useCapture` is set, which would explain the test failure here?
I never use `useCapture` though so I may totally misundertand it!

bencroker:
> Note that I'll remove the "ready for review" label for now; despite its generic name, we actually use this label just for 1cg, to tag PRs that are ready for his final review

Apologies, I hadn’t realised that.

>  I never use useCapture though so I may totally misundertand it!

I’m just learning about it too, but in doing so I’m realising that there’s really little or no value in us testing that the `addEventListener()` method works as it should, and I can’t come up with a useful to test that confirms passing a third/fourth argument to it. I’m thinking that perhaps this addition doesn’t require a test 😱. What say you?

test/core/events.js
bencroker: As per https://github.com/bigskysoftware/htmx/pull/2836#issuecomment-2305544011, I don’t think there’s any value in testing each value that the `addEventListener()` method accepts. I’ll mark this as resolved and we can continue discussing below.

Telroshan:
>  I hadn’t realised that.

No worries, it isn't written anywhere and has been an implicit convention between us, you couldn't know!

> perhaps this addition doesn’t require a test

Well, I agree on this, `useCapture` is indeed, according to the docs, the old parameter type, that has now been replaced by `options` which includes a `capture` property anyway.

As you have a test for `once`, there's no reason it wouldn't work for the other properties, and I agree that the current test is doing its job; proving that we correctly pass the `options` to `addEventListener`, we don't need more than that.

Besides, `options` isn't [IE11 compatible](https://caniuse.com/?search=addeventlistener%20options), but as we dropped IE11 support for htmx 2 anyway, there's no need to go beyond reason on older browsers support either.

With that, back to ready for review!

bencroker:
Merci!

1cg:
great change ben!

Issue #2833, opened 2024-08-19
Page title not restored correctly after back button in Firefox


ainz-95:
This seems to happen in Firefox (129.0) but not in Chrome (127.0) with version 2.0.2, where the user clicks back button and the page title is not correctly restored. See reproduction link below:

https://stackblitz.com/edit/htmx-firefox-back-title-issue?file=index.html

1. Click _Open Preview in New Tab_
2. Page title is `Home Page`
3. Navigate via `Click Me!`
4. Page title is `Hello Page`
5. Navigate via browser back button
6. Page title is _still_ `Hello Page`

MichaelWest22:
reproduced this in firefox.  debugging it and I can see that in chrome the title is cached by chrome itself for back history requests and the htmx code does not set the title because it uses the fragment data to get the title but it doesn't store the title in fragment so it just leaves it up to the browser to set title.  The title is stored in the cache though so just changing one line from fragment.title to cached.title fixed it for me in firefox.

```javascript
  function restoreHistory(path) {
    saveCurrentPageToHistory()
    path = path || location.pathname + location.search
    const cached = getCachedHistory(path)
    if (cached) {
      const fragment = makeFragment(cached.content)
      const historyElement = getHistoryElement()
      const settleInfo = makeSettleInfo(historyElement)
      handleTitle(cached.title) // changed this one line from fragment.title
```

PR #2841, opened 2024-08-22
fix restoreHistory title replacment

MichaelWest22:
## Description
It seems that when the handleTitle() function was introduced the restoreHistory() title setting code was accidently changed to use fragment.title instead of the proper cached.title.  Firefox does not keep track of title during history switching so title doesn't change now because there is no title inside the fragment data.  Chrome works fine as it tracks the page title itself.

Corresponding issue:
#2833 

## Testing
I've tested this trival change in firefox and chrome to make sure it works fine and updates the title as expected

## Checklist

  source changes)
  approved via an issue


1cg:
Great fix, thank you!

Issue #2850, opened 2024-08-26
Typescript: Configuration variables shouldn't be const


ColonelThirtyTwo:
In `htmx.d.ts`, the variables in the `htmx.config` are declared using `const`, so projects using typescript will be unable to set them without typescript throwing an error.

Same with `htmx.logger`.

Telroshan:
Oof, good catch, hadn't realized this!

Ok so it's hard to get `tsc` to generate what we want, we had to do crazy tricks sometimes to get a proper result (see #2336, #2653)
If you figure out a way to have tsc (see the [`npm run type-declarations`](https://github.com/bigskysoftware/htmx/blob/master/package.json#L34) command) generate the correct thing, a bugfix PR would be very much welcome!

ColonelThirtyTwo:
@Telroshan `type-declarations` fails because it can't find `tsconfig.json` - I don't see one anywhere in the repo.

ColonelThirtyTwo:
Looks like e1143de8500dc91f54ed40aa05306a0049720804 generates the `d.ts` as part of `npm run types-generate`, and both the `type-declarations` command and the `src/html.d.ts` file are not used anymore. Should probably delete those.

ColonelThirtyTwo:
Looks like Typescript v5.1 and up generate `let` instead of `const` for all of the exported members of the htmx namespace, so the solution might just be to update typescript.

MichaelWest22:
Yeah newer typescript generates with let which hopefully solves your issue.  How do we test the updated esm.d.ts file to confirm it still works fine and resolves the problem?

v5.5 version seems to add additional changes because it has stricter rules it seems.  I don't know if the updates type inference it does is an improvement we want now or not.
example:
```javascript
// typescript v4.9.5
    const trigger: (elt: string | EventTarget, eventName: string, detail?: any) => boolean;
    const ajax: (verb: HttpVerb, path: string, context: string | Element | HtmxAjaxHelperContext) => Promise<void>;
    const find: (eltOrSelector: string | ParentNode, selector?: string) => Element;
    const findAll: (eltOrSelector: string | ParentNode, selector?: string) => NodeListOf<Element>;
    const closest: (elt: string | Element, selector: string) => Element;

// typescript v5.4.5
    let trigger: (elt: string | EventTarget, eventName: string, detail?: any) => boolean;
    let ajax: (verb: HttpVerb, path: string, context: string | Element | HtmxAjaxHelperContext) => Promise<void>;
[...]

PR #2851, opened 2024-08-27
Upgrade Typescript to move configuration from const to let

MichaelWest22:
## Description
Upgrade of Typescript version to 5.5.4 to allow creation of d.ts file with let instead of const which seems to be the case with version 5.1 and above and this hopefully will allow typescript users to modify config items that were blocked from re-assignment with const.

tested with all recent releases of typescript and 5.5.4 have new type generation rules that seem to respect the jsdoc type annotations better.  see Node Reuse in Declaration Emit in https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/

Also removed the duplicated src/htmx.d.ts file that is no longer used for the distribution? need to verify we no longer need to keep maintaining this file manually.

Resolved one type check for HttpVerb detected by updated typescript.

Corresponding issue:
#2850 

## Testing
Ran test suite and but I have not tested the generated d.ts file in a typescript project to confirm it resolves the original problem

## Checklist

  source changes)
  approved via an issue


Telroshan:
Love it! Didn't know that simply upgrading TS would give us what we want here, thought it was an issue with our JSDoc annotations 😆 

Could you rebase against `dev`? Lint issues have been fixed in 0fd8547 in the meantime

MichaelWest22:
rebase attempt has left a mess so recreating PR

PR #2866, opened 2024-09-04
Add shadowRoot host selector

MichaelWest22:
## Description
In the web component documentation there is a reference to using 'host' to target the web components shadow DOM host element but this was never implemented as a valid target in htmx.  So I've implemented this as a valid selector and added tests to confirm swapping it and elements outside the shadow DOM can be targeted for replacement.

There was a similar 'root' one implemented however from my testing this seems to not be usable for two reasons.  First is that because the shadow root has a very different type called ShadowRoot which does not support most of the options a Node or Element would support which means swaps to this target fail when trying to add classes to the root. Also target has now been wrapped in asElement() which strips root target.  So we should probably remove this non functional root selector as it is not documented anywhere either.  The 'root' target could have been used by non web component use but this would make it return Document which is also not a Element so it gets stripped as well making it unusable here as well.  So there should be no downside I can see if we remove 'root'.

Corresponding issue:
#2846

## Testing
Wrote new tests to confirm it is working as expected plus did some manual testing in a test app to confirm Host replacment was working as expected.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1161,6 +1161,8 @@ var htmx = (function() {
       return [document.body]
     } else if (selector === 'root') {
       return [getRootNode(elt, !!global)]
+    } else if (selector === 'host') {
+      return [(/** @type ShadowRoot */(elt.getRootNode())).host]
     } else if (selector.indexOf('global ') === 0) {
       return querySelectorAllExt(elt, selector.slice(7), true)
     } else {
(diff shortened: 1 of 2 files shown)


Telroshan:
Note: I'm labelling this as a bug, as even though this is technically a new feature/enhancement, it was (and still is) documented as working in [our docs](https://htmx.org/examples/web-components/)

PR #2868, opened 2024-09-05
Add detail.question and detail.issueRequest(skipConfirmation=false) documentation to htmx:confirm event

eduardolat:
## Description
The documentation for this two details is not clear enough, so this pr improves a little bit the documentation for this thing

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/events.md
+++ b/www/content/events.md
@@ -192,7 +192,7 @@ document.body.addEventListener('htmx:confirm', function(evt) {
     evt.preventDefault();
     swal({
       title: "Are you sure?",
-      text: "Are you sure you are sure?",
+      text: evt.detail.question,
       icon: "warning",
       buttons: true,
       dangerMode: true,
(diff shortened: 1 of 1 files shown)


www/content/events.md
```diff
@@ -192,7 +192,7 @@ document.body.addEventListener('htmx:confirm', function(evt) {
     evt.preventDefault();
     swal({
       title: "Are you sure?",
-      text: "Are you sure you are sure?",
+      text: evt.detail.question,
```
Telroshan, line 195: Not sure this is relevant here, as [the doc](https://htmx.org/docs/#confirming-requests-using-events) states:
> This event is fired on every trigger for a request (not just on elements that have a hx-confirm attribute)

So in the context of this doc, we can't assume that a `hx-confirm` attribute is set, and it likely is not since this example is precisely about using another confirm mechanism from a lib, instead of htmx's one

www/content/events.md
```diff
@@ -207,15 +207,16 @@ document.body.addEventListener('htmx:confirm', function(evt) {
 
 ##### Details
 
-{target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest}
+{target: target, elt: elt, path: path, verb: verb, triggeringEvent: event, etc: etc, issueRequest: issueRequest, question: question}
```
Telroshan, line 210: I think this line can probably be removed altogether, seems like the previous line shouldn't have been there at all looking at the other events' docs 🤔 maybe a copied object that was forgotten and left in here at the time this doc was written?

www/content/events.md
```diff
@@ -207,15 +207,16 @@ document.body.addEventListener('htmx:confirm', function(evt) {
...
 
 * `detail.elt` - the element in question
 * `detail.etc` - additional request information (mostly unused)
-* `detail.issueRequest` - a no argument function that can be invoked to issue the request (should be paired with `evt.preventDefault()`!)
+* `detail.issueRequest(skipConfirmation=false)` - a function that can be invoked to issue the request (should be paired with `evt.preventDefault()`!), if skipConfirmation is `true` the original `window.confirm()` is not executed
```
Telroshan, line 214: Good catch! Maybe we should also update the example above then? Shouldn't we pass an explicit `true` to `issueRequest` in the sweet alert example?

alexpetros:
This one looks like it's not quite ready. I'll re-open if you make the changes @Telroshan describes.

eduardolat:
Hi @alexpetros , first of all thank you for your amazing work on HTMX. Secondly i just opened a new PR #2926 taking into account @Telroshan suggestions and other modifications to make it more understandable

Issue #1703, opened 2023-08-18
View transitions not applied on history changes


tillschander:
View transitions are great and even work nice with `hx-boost="true"`. But only during normal navigation with links. If users jump to a previous history entry (e.g. via the browser back button) they won't get to see a view transition.

Currently `document.startViewTransition()` is only ever called in [`handleAjaxResponse()`](https://github.com/bigskysoftware/htmx/blob/193a9ea9d2a5f2a216f1c5494812b1a6ada93bb2/src/htmx.js#L3258)
https://github.com/bigskysoftware/htmx/blob/193a9ea9d2a5f2a216f1c5494812b1a6ada93bb2/src/htmx.js#L3482-L3485

It should also be called on history changes. For example in [`restoreHistory()`](https://github.com/bigskysoftware/htmx/blob/193a9ea9d2a5f2a216f1c5494812b1a6ada93bb2/src/htmx.js#restoreHistory)
https://github.com/bigskysoftware/htmx/blob/193a9ea9d2a5f2a216f1c5494812b1a6ada93bb2/src/htmx.js#L2279

There one might wrap the call to `swapInnerHTML()` in a call to `startViewTransition()`

Something like this works already (but isn't yet production ready, of course):
```diff
+ document.startViewTransition(function () {
     swapInnerHTML(historyElement, fragment, settleInfo)
     settleImmediately(settleInfo.tasks);
+ });
````

TheStanian:
I have done some testing on a local project with this. It breaks scroll position restore, but this can be solved by increasing the timeout of the scroll. (e.g. the default settle delay). I currently have both the view transition and the increased scroll delay locked behind the `globalViewTransitions` config value, and this seems to work nicely.

MichaelWest22:
Just put up a possible PR that can hopefully fix this gap

1cg:
Fixed in 2.x by #3328. History restore now goes through swap(), which honors htmx.config.globalViewTransitions.

PR #2873, opened 2024-09-05
Global view transitions for history restore

davidwilde:
## Description
I added ViewTransitions to the restoreHistory function. In this pull request I only made this change when the user has enabled htmx.config.globalViewTransitions. This is as a PR for opening a discussion on this as anything else as I have some questions to the best approach for this work. 

I understand that because this not a either a bugfix, a documentation update, or a new feature that has been explicitly approved via an issue. I apologise as I only saw those particular guideline

### Questions

1. Should this have its own config flag to enable this behavior? Potentially this is breaking expectations by adding this to navigating history. Is adding this under globalViewTransitions sensible here? Maybe a config flag like `htmx.config.historyViewTransitions`? 
2. Should a htmx:beforeTransition event be thrown?

### Idea

During development of this I had an idea but I thought it would be too big a PR and I wanted to start this initial discussion first, but I considered that it could be possible to add a transition flag into the history. This would be caught from the hx-swap of each request and mean that 

Corresponding issue:

https://github.com/bigskysoftware/htmx/issues/1703

## Testing
I tested this manually, switching on and off the `htmx.config.globalViewTransitions` flag. I ran the test suite locally and the linting rules

## Checklist

  source changes)
  approved via an issue


davidwilde:
I'm really sorry as to why I seem incapable of opening this PR properly - I fully believed I'd opened it from `dev` when I made it

pokonski:
This would be great to have as currently navigating back is jarring compared to navigating forward

src/htmx.js
```diff
@@ -3205,19 +3205,33 @@ var htmx = (function() {
...
+      if (shouldTransition &&
+        // @ts-ignore experimental feature atm
+        document.startViewTransition) {
+        const innerRestoreHistoryFromCache = restoreHistoryFromCache
+        // wrap the original doRestoreHistory() in a call to startViewTransition()
+        // @ts-ignore experimental feature atm
+        restoreHistoryFromCache = document.startViewTransition(
+          () => innerRestoreHistoryFromCache()
+        )
+      }
```
1cg, line 3233: Since there is no potential delay here I think this be an if/else instead of wrapping the restoreHistoryFromCache function: pass it directly to startViewTransition else invoke it

1cg (changes requested):

MichaelWest22:
This change is now merged in as part of #3328 which is now part of 2.0.5 release so you can test it works as expected in this release now

Issue #2869, opened 2024-09-05
htmx.ajax swaps body's innerHTML when the target id doesn't exist


ohmyohmyohmy:
I have an app where tabs hide certain div-ids when not loaded. When the div with the id visible, htmx.ajax() POST works as expected swapping the innerHTML as directed. But when the div with the id is no longer visible, the function replaces the entire innerHTML of the body with the response. Is this what's supposed to happen?

MichaelWest22:
This is happening because the htmx.ajax() ajax helper function is calling the the issueAjaxRequest function with the default body as the element and passing in a targetOverride set to the resolved target you passed in.  If the targetOveride passed in is null because the resolved target is not found then it falls back to checking the body element for its target.  And if you happen to have boosted=true on the body then it makes the body target the body causing your issue.

I think this is an edge case bug and not the intended behaviour.

to resolve it in your situation i would do:
```javascript
if(htmx.find(target)) {
  htmx.ajax('GET', url, target)
}
```

ohmyohmyohmy:
I didn't think of that. Thank you for that suggestion.

MichaelWest22:
Actually looking at it the boosted=true is not required as if target is invalid it falls back to checking body target and unless you set a hx-target on the body it will always return itself as the target boost or no boost.  I don't know if this is really intended and may just be what it falls back to.  Can fix the issue with https://github.com/bigskysoftware/htmx/compare/dev...MichaelWest22:htmx:ajax-no-target which gets it to htmx:targetError on bad requests but this could be a breaking change if someone was depending on this broken fallback behaviour?

Telroshan:
> this could be a breaking change if someone was depending on this broken fallback behaviour

It could but honestly it looks really convoluted to me, why would you want the server's reponse that you intended to swap at a very specific place (since you passed in a selector/target element), to instead replace the whole body's content?
Imho it should reflect the behavior of `hx-target`, aka that if no target is found, the request doesn't happen and a `htmx:targetError` is fired. And as you did in your commit, only do that if a target override was _explicitly_ set.

I like your suggested change @MichaelWest22 , feel free to open a bugfix PR.
Just a few questions/suggestions though
- Couldn't this be greatly simplified by removing `asElement(getTarget(resolveTarget(context.source)))) || DUMMY_ELT` altogether ? I would expect htmx to precisely fallback to the `source` element's `hx-target` value, I don't think this is needed, but let me know!
[...]

PR #2878, opened 2024-09-06
fix htmx.ajax defaulting to swap body when target not found

MichaelWest22:
## Description
The htmx.ajax helper function can cause issues when the target is not set correctly or the target selector happens to be missing in the DOM as it passes in null target override to the ajax request which defaults to targeting the body.  Changing it to detect invalid targets and pass in a DUMMY_ELT which triggers the htmx:targetError solves this.  Also had to handle situations where a valid source is supplied as its valid to fall back to source targeting an explicit target is not defined but if we still want it to fail if target was defined.

Corresponding issue:
#2869 

## Testing
Wrote some new tests to cover most invalid targeting situations and also tested calling htmx.ajax manually to confirm it rejects invalid targets but passes though valid ones fine.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3891,7 +3891,7 @@ var htmx = (function() {
     if (context) {
       if (context instanceof Element || typeof context === 'string') {
         return issueAjaxRequest(verb, path, null, null, {
-          targetOverride: resolveTarget(context),
+          targetOverride: resolveTarget(context) || DUMMY_ELT,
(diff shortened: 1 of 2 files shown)


MichaelWest22:
rewrote the logic into an expanded if statement so its easier to understand and debug and added inline comment

Telroshan (approved):

PR #2882, opened 2024-09-08
Documentation of svg oob swaps

MichaelWest22:
## Description
SVGs have issues when you try and modify their contents because they have a different namespace that is only setup when you wrap them in a svg tag.  So to allow internal SVG replacment I've found you can use wrapping in both template and svg tags which then works well with OOB swaps.  It is also possible to do similar full swaps but you have to use hx-select and its not as easy.

Corresponding issue:

## Testing
Performed manual testing of different oob internal svg swaps to check it works as expected

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/attributes/hx-swap-oob.md
+++ b/www/content/attributes/hx-swap-oob.md
@@ -53,7 +53,7 @@ Note that these template tags will be removed from the final content of the page
 
 ### Slipery SVGs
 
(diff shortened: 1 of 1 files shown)


Telroshan:
(Note: took the liberty to fix a typo in the title and added some punctuation.)
Thanks !

Telroshan (approved):

PR #2893, opened 2024-09-11
Triggerspec delay unique

jsonn:
## Description
If a node has multiple trigger specifications, they should have independent delays. If a single delayed trigger specification is targets multiple nodes, they should share the delay.

Corresponding issue: #2892

## Testing
Both variants are explicitly tested. The non-sharing of multiple trigger specifications failed before.

## Checklist

* [X] I have read the contribution guidelines
* [X] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [ ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [X] I ran the test suite locally (`npm run test`) and verified that it succeeded

Note: the change is on top of #2891

PR #2896, opened 2024-09-11
Fix checkbox trigger specifier when used with checkboxes

jsonn:
## Description
Unlike other form elements, checkbox elements will return `on` whether or not they are currently ticked. Explicitly handle this by checking for the tag name and type and then use the checked property.

## Testing
A new test case is included, failing before.

## Checklist

  source changes)
  approved via an issue

Note: this PR is on top of #2893

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -332,6 +332,7 @@ var htmx = (function() {
     getAttributeValue,
     getClosestAttributeValue,
     getClosestMatch,
+    getElementValue,
     getExpressionVars,
     getHeaders,
     getInputValues,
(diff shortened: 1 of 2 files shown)


Telroshan:
Note: I'm labelling this as "under discussion" until #2891 is processed, as the latter includes some changes that this PR depends on.
If and when #2891 is merged in, this PR's diff will be smaller thus easier to review.

jsonn:
Rebased on current dev, should be easy to merge now.

1cg:
thank you for this contribution, sorry it sat so long, looks like this affects 4.x too so I'm talking w/ @MichaelWest22 about best fixes

MichaelWest22:
Fixed in my 4 PR now.

Issue #2660, opened 2024-06-22
`hx-disabled-elt` does not work with multiple parameterized selectors


mtlynch:
Sidenote: Thanks for this library and for *Hypermedia Systems*! I'm reading the book and experimenting with a port of [one of my apps](https://github.com/mtlynch/screenjournal) to htmx.

From my experimentation with htmx 2.0.0, it seems like the `hx-disabled-elt` attribute works fine with multiple selectors when they're simple CSS selectors like `hx-disabled-elt="#b1, #b2"`, but when I specify multiple selectors using `next` or `find`, then only the first element in the list is processed.

Here's an example using `next`, but I've also seen this with `find`:

```javascript
  it('multiple elts can be disabled with next', function() {
    this.server.respondWith('GET', '/test', 'Clicked!')
    var b1 = make('<button hx-get="/test" hx-disabled-elt="next #b2, next #b3">Click Me!</button>')
    var b2 = make('<button id="b2">Click Me!</button>')
    var b3 = make('<button id="b3">Demo</button>')

    b2.hasAttribute('disabled').should.equal(false)
    b3.hasAttribute('disabled').should.equal(false)

    b1.click()
    b2.hasAttribute('disabled').should.equal(true)
    b3.hasAttribute('disabled').should.equal(true)

    this.server.respond()

    b2.hasAttribute('disabled').should.equal(false)
    b3.hasAttribute('disabled').should.equal(false)
  })
```

```bash
$ npm run test
...
  643 passing (5s)
  3 pending
  1 failing

  1) hx-disabled-elt attribute
       multiple elts can be disabled with next:

      AssertionError: expected false to equal true
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (attributes/hx-disabled-elt.js:94:40)
```

Telroshan:
Hey, you're totally right, multiple extended selectors aren't supported yet.
See #2645 and #2610 for similar issues & discussions.

A current workaround for this is to rely on multiple standard CSS selectors, using for example a parent ID and a [child selector](https://developer.mozilla.org/fr/docs/Web/CSS/Child_combinator), or a [sibling selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Next-sibling_combinator)

tbuckley:
I can use multiple standard CSS selectors in my use case, but just wanted to note the docs show multiple extended selectors being supported:

```
<form hx-post="/example" hx-disabled-elt="find input[type='text'], find button">
```

[...]

Issue #2645, opened 2024-06-18
hx-include - allow multiple extended selectors


rdehuyss:
It would be really handy if `hx-include` could use multiple selectors, just like [`hx-trigger`](https://htmx.org/attributes/hx-trigger/#multiple-triggers).

### Reasoning:
- htmx by default submits everything in the closest `<form>` element. This confused me as the html spec says that forms may not be nested.
- I tried using nested `<form>` elements (to only submit a part for a partial update) but it is not allowed according to the spec and chrome does not support it
- so, in the end, I had to remove all forms and I use `hx-include` to only post the input fields nested inside a div. But in that case, I cannot include another input field that is not part of that div but global.

rdehuyss:
Relates to https://github.com/bigskysoftware/htmx/issues/1754

Telroshan:
> htmx by default submits everything in the closest form element. This confused me as the html spec says that forms may not be nested.

Please note this is the case for non-GET requests but most importantly for "isolated inputs", as a form's `closest` form is that very form itself. As per the [`closest`' doc](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest):
> Returns the closest ancestor [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) or itself, which matches the selectors.

https://github.com/bigskysoftware/htmx/blob/f9b3f883576c979b533ed657b8a7556c3b1982e4/src/htmx.js#L3416-L3419

The idea here is that if you make a non-GET `hx-` request on an input that is inside a form, it's very likely that you'll want the full context along, i.e. the parent form's data. If the request comes from a form already, this won't do anything as the closest form is that same form.
You should indeed not use nested form elements and we don't encourage this at all.

As for `hx-include`, you should already be able to use multiple selectors.
As per [the doc](https://htmx.org/attributes/hx-include/):
> A CSS query selector of the elements to include.
> A standard CSS selector resolves to [document.querySelectorAll](https://developer.mozilla.org/docs/Web/API/Document/querySelectorAll) and will include multiple elements, while the extended selectors such as find or next only return a single element at most to include

Which means any standard CSS selector matching multiple elements, will return those multiple elements. The exception here are the extended selectors that htmx provides (`closest`, `find`, `next`, `previous`) which for now are limited to a single selector. But using IDs you may totally reach your goal using `hx-include="#container1 .something, #container2"` for example.

[...]

Issue #2610, opened 2024-06-12
`previous input` not work as expected.


zw963:
Plese check following example, there are two input element which has same name.

```html
<input type="hidden" value="WAnUI0ZhE8Gdab2pVgFPPDG-6Av7NuIdmXDlY_z08xQ" name="_csrf">
....
<label>
  <input type="hidden" name="university:marked" value="false" id="marked_unmark">
  <input type="checkbox" name="university:marked" value="true" id="marked" hx-put="/universities/812" hx-swap="none" hx-indicator="#spinner" hx-include="[name='_csrf'],previous input" class="">
  <span>标记</span>
</label>
```

The `previous input` not work in above include attribute, it not includes the hidden input elements which above it, 
and cause `university:marked set to false` is not available when user uncheck the checkbox.

It not work if i move hidden input after the checkbox input, and use `next input` too.

But, if use `#id` selector instead,  it work as expected as following code:

```html
<input type="hidden" value="WAnUI0ZhE8Gdab2pVgFPPDG-6Av7NuIdmXDlY_z08xQ" name="_csrf">
....
<label>
  <input type="hidden" name="university:marked" value="false" id="marked_unmark">
  <input type="checkbox" name="university:marked" value="true" id="marked" hx-put="/universities/812" hx-swap="none" hx-indicator="#spinner" hx-include="[name='_csrf'],input#marked_unmark" class="">
  <span>标记</span>
</label>
```

** Version

htmx 1.9.12
hyperscript 0.9.12

Thanks

Telroshan:
Hey, unfortunately it's indeed a weakness of our extended selectors syntax at the moment. It's only handled if it's the only selector, i.e:
- `previous input` will work
- But `div, previous input` won't, and will simply result in a `document.querySelectorAll("div, previous input")`, which of course won't retrieve the element you're looking for

That's a weakness we'll have to fix in the future, and for now you may indeed use this ID workaround that you already found out!

PR #2902, opened 2024-09-12
Support multiple extended selectors for hx-include, hx-trigger from, and hx-disabled-elt

Telroshan:
## Description
`hx-include` will include all elements matching the given selector, but currently only supports 1 extended selector at a time, meaning that
```html
<div hx-get="/" hx-include="input, #test">
```
will work, and include values from **all** elements that match any of these 2 selectors.

Using 1 extended selector works well, like
```html
<div hx-get="/" hx-include="find input">
```
will retrieve the first child input's value, and works as expected.

However, the following won't work:
```html
<div hx-get="/" hx-include="find input, previous input">
```

Htmx will currently simply parse the first `find` keyword and use `input, previous input` as the selector to test against previous elements. So, the second part  `previous input` won't work at all, as its `previous` keyword won't even be interpreted by htmx.
_It will instead be processed as a standard selector, meaning searching for an `input` under a `previous` tag (which ofc, doesn't exist)_

## Explanations
This PR does the following:
- Break the selector string by comma, to retrieve individual selectors
- Check for special extended keywords, at the start of each of those selectors, and combine the results
- Standard selectors are joined back at the end (effectively resulting in the initial selector with any extended selector removed from it) and passed to `querySelectorAll` as it used to.

Corresponding issues: #2645 #2660 #2610

## Testing
Added 2 test cases

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1145,7 +1145,7 @@ var htmx = (function() {
         }
         if (char === '<') {
           chevronsCount++
-        } else if (selector.substring(i, i + 2) === '/>') {
+        } else if (char === '/' && i < selector.length - 1 && selector[i + 1] === '>') {
           chevronsCount--
         }
       }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1127,33 +1127,57 @@ var htmx = (function() {
    */
   function querySelectorAllExt(elt, selector, global) {
     elt = resolveTarget(elt)
-    if (selector.indexOf('closest ') === 0) {
-      return [closest(asElement(elt), normalizeSelector(selector.slice(8)))]
-    } else if (selector.indexOf('find ') === 0) {
-      return [find(asParentNode(elt), normalizeSelector(selector.slice(5)))]
-    } else if (selector === 'next') {
-      return [asElement(elt).nextElementSibling]
-    } else if (selector.indexOf('next ') === 0) {
-      return [scanForwardQuery(elt, normalizeSelector(selector.slice(5)), !!global)]
-    } else if (selector === 'previous') {
-      return [asElement(elt).previousElementSibling]
-    } else if (selector.indexOf('previous ') === 0) {
-      return [scanBackwardsQuery(elt, normalizeSelector(selector.slice(9)), !!global)]
-    } else if (selector === 'document') {
-      return [document]
-    } else if (selector === 'window') {
-      return [window]
-    } else if (selector === 'body') {
-      return [document.body]
-    } else if (selector === 'root') {
-      return [getRootNode(elt, !!global)]
-    } else if (selector === 'host') {
-      return [(/** @type ShadowRoot */(elt.getRootNode())).host]
-    } else if (selector.indexOf('global ') === 0) {
-      return querySelectorAllExt(elt, selector.slice(7), true)
-    } else {
-      return toArray(asParentNode(getRootNode(elt, !!global)).querySelectorAll(normalizeSelector(selector)))
+
+    const parts = selector.split(',')
+    const result = []
+    const unprocessedParts = []
+    while (parts.length > 0) {
+      const selector = normalizeSelector(parts.shift())
+      let item
+      if (selector.indexOf('closest ') === 0) {
+        item = closest(asElement(elt), normalizeSelector(selector.substr(8)))
+      } else if (selector.indexOf('find ') === 0) {
+        item = find(asParentNode(elt), normalizeSelector(selector.substr(5)))
+      } else if (selector === 'next' || selector === 'nextElementSibling') {
+        item = asElement(elt).nextElementSibling
+      } else if (selector.indexOf('next ') === 0) {
+        item = scanForwardQuery(elt, normalizeSelector(selector.substr(5)), !!global)
+      } else if (selector === 'previous' || selector === 'previousElementSibling') {
+        item = asElement(elt).previousElementSibling
+      } else if (selector.indexOf('previous ') === 0) {
+        item = scanBackwardsQuery(elt, normalizeSelector(selector.substr(9)), !!global)
+      } else if (selector === 'document') {
+        item = document
+      } else if (selector === 'window') {
+        item = window
+      } else if (selector === 'body') {
+        item = document.body
+      } else if (selector === 'root') {
+        item = getRootNode(elt, !!global)
+      } else if (selector === 'host') {
+        item = (/** @type ShadowRoot */(elt.getRootNode())).host
+      } else if (selector.indexOf('global ') === 0) {
+        // Previous implementation of `global` only supported it at the first position and applied it to the entire selector string.
+        // For backward compatibility and to maintain logical consistency, we make it apply to everything that follows.
+        parts.unshift(selector.slice(7))
+        result.push(...querySelectorAllExt(elt, parts.join(','), true))
+        break
+      } else {
+        unprocessedParts.push(selector)
+      }
+
+      if (item) {
+        result.push(item)
+      }
     }
+
+    if (unprocessedParts.length > 0) {
+      const standardSelector = unprocessedParts.join(',')
+      const rootNode = asParentNode(getRootNode(elt, !!global))
+      result.push(...toArray(rootNode.querySelectorAll(standardSelector)))
+    }
+
+    return result
   }
 
   /**
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2370,7 +2370,7 @@ var htmx = (function() {
           path = getDocument().location.href
         }
         if (verb === 'get' && path.includes('?')) {
-          path = path.replace(/\?[^#]+/, '');
+          path = path.replace(/\?[^#]+/, '')
         }
       }
       triggerSpecs.forEach(function(triggerSpec) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1126,6 +1126,10 @@ var htmx = (function() {
    * @returns {(Node|Window)[]}
    */
   function querySelectorAllExt(elt, selector, global) {
+    if (selector.indexOf('global ') === 0) {
+      return querySelectorAllExt(elt, selector.slice(7), true)
+    }
+
     elt = resolveTarget(elt)
 
     const parts = selector.split(',')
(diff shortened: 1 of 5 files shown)


src/htmx.js
```diff
@@ -1141,31 +1141,51 @@ var htmx = (function() {
...
-    } else if (selector.indexOf('global ') === 0) {
-      return querySelectorAllExt(elt, selector.slice(7), true)
-    } else {
-      return toArray(asParentNode(getRootNode(elt, !!global)).querySelectorAll(normalizeSelector(selector)))
+
+    const parts = selector.split(',')
+    const result = []
+    const unprocessedParts = []
+    for (let i = 0; i < parts.length; i++) {
+      let selector = parts[i].trimStart()
```
MichaelWest22, line 1149: trim() is pobably better here as trailing spaces could cause issues as well for 'closest ', 'global ', 'find ' etc.  trailing spaces should never really be passed in but why not deal with them anyway just in case anyway.

src/htmx.js
```diff
@@ -1141,31 +1141,51 @@ var htmx = (function() {
...
+
+      if (selector.indexOf('global ') === 0) {
+        global = true
```
MichaelWest22, line 1152: overriding global here will impact all later iterations making them global as well.  not likely in practice but could be fixed just in case anyway.

src/htmx.js
```diff
@@ -1141,31 +1141,51 @@ var htmx = (function() {
...
+        result.push(document)
+      } else if (selector === 'window') {
+        result.push(window)
+      } else if (selector === 'body') {
+        result.push(document.body)
+      } else if (selector === 'root') {
+        result.push(getRootNode(elt, !!global))
+      } else {
+        // Push back the unaltered string part, so that the standard selector fallback uses the exact initial string
+        unprocessedParts.push(parts[i])
```
MichaelWest22, line 1179: push selector here and you won't have to reassign parts[i] = selector if global

MichaelWest22: not sure if trimming whitespace would cause issues though??  my testing in chrome shows whitespace ignored but is this the same for all browsers?

src/htmx.js
MichaelWest22: Actually maybe don't need to fix this as supporting full global and non global flexibility is probably not a real useful feature and it will take a lot of work to solve this properly. Or you would have to remove your unprocessedParts method and rename and then wrap the original function unchanged with an outer for loop.  

It would still support having global at the start like it works now so won't break any existing use cases even if we don't handle every edge case.

src/htmx.js
```diff
@@ -1141,31 +1141,51 @@ var htmx = (function() {
    */
   function querySelectorAllExt(elt, selector, global) {
```
MichaelWest22, line 1142: global optional field can be removed with your change as it is never called except by internal recursion which you have removed.  This function is never used elsewhere with this param and it is not published for internalAPI etc.

src/htmx.js
MichaelWest22: Actually we could use `normalizeSelector(parts[i])`  This does a trim and some additional handling and will allow us to handle `<input/>, <button/>` which would not be handled well otherwise as it will strip to `input/>, <button`.  we could then remove normalizeSelector from the last unprocessed parts query i think but may still need it for the middle 4 cases for someone doing `find <input/>`

src/htmx.js
MichaelWest22: also once you define `let global = false` at the top you no longer need to !! the global's to make the boolean.

Sorry for too many comments :/

Telroshan: You're right! I went into "keep it backwards compatible" by habit without even checking usages

No worries for the comments count, especially when it helps improving the code 😛

src/htmx.js
Telroshan: Indeed, forgot about `normalizeSelector` which is literally above this function 🤦‍♂️  thanks!

src/htmx.js
Telroshan: > overriding global here will impact all later iterations

Yeah I wanted to preserve the old behavior where `global` at the start of a selector would apply to all following selectors.

I wondered a lot about this, and I'm still unsure about what we should strive for functionnally:
- do we keep `global` applying to all selectors if put at the start, but only to the selector it's put in front of if it's not in first position ? 
- do we keep `global` applying to all selectors if put at the start, and to all subsequent selectors no matter its position ?

It's a bummer that we have to maintain this "`global` first = apply to all" for backwards compatibility but we can't help that. Still unsure on the way to go about this. When I wrote this code I felt that the second option was the most logical (or the less nonsensical), now I'm not sure anymore 😆

src/htmx.js
Telroshan: Just noticed that the `querySelectorAll` call below is using `normalizeSelector` anyway so... you're right, this can be simplified, whitespaces were always being removed it seems!

src/htmx.js
MichaelWest22: All your changes are looking good to me.  I think you can leave the global handling as you have it now.  It should be backwards compatible. If someone does need to do local web component scoped selector plus global scoped at the same time you can still do `find input, global button` and it will select a local web component scoped input plus all document scoped buttons for example.  the only issue is if someone does `input, global button` it will not work as they may expect as input moves to global scope because non extended selectors will go in unprocessedParts and now be global.

If we wanted to support global not applying to earlier non extended selectors like this you could remove the unprocessedParts fallback like this:
```javascript
        item = getRootNode(elt, global)
      } else {
        const rootNode = asParentNode(getRootNode(elt, global))
        result.push(...toArray(rootNode.querySelectorAll(selector)))
      }

      if (item) {
        result.push(item)
      }
    }

    return result
```
Or make it more complex with unprocessedGlobalParts[] as well.

Telroshan: > it will not work as they may expect as input moves to global scope because non extended selectors will go in unprocessedParts and now be global.

Ooooh right, totally overlooked that. Indeed, that's not ideal, I'll fix this to at least work in a predictable manner for users

Telroshan: Alright should be fine now @MichaelWest22, I also added a test to ensure that `global` doesn't apply to previous selectors.
I didn't go for the "push at every iteration" approach, as it would break nested standard selectors (see the added `'hx-include processes nested standard selectors correctly'` test).
As the initial selectors string is split on commas, this would break the following standard selector in 2 unusable halves if not joined back in a single string to pass to `querySelectorAll` : `:is([name='i1'], [name='i2'])`

MichaelWest22: Yeah that looks to work and cover all the cases well now!  You had to add some duplication and complexity which is why you need such big inline comments.  I Just tried rewriting it by turning the for parts[i] loop using a while loop with shift/unshift and i got this: https://github.com/MichaelWest22/htmx/commit/1b4a7949f9b8584ab08695b22366f066d2cc9c21

It follows the original global recursive pattern unchanged but seems to handle everything well.  The only functional change is that it adds global results first before standard selector results but i don't think result return order is a breaking change.  Anyway you can use this and add any required inline comments or just go with what you have already that works fine.

Telroshan: Oh, very smart @MichaelWest22 indeed.

> The only functional change is that it adds global results first before standard selector results but i don't think result return order is a breaking change

I don't it is either, the only case I can imagine is if you want to preserve values order for multiple inputs that share the same `name`, but for it to be an issue, that would mean you have some of them in the shadow dom and some other outside... Sounds like a bit too-convoluted use-case.
Anyway, if that's an issue for someone, it can always be fixed later on (after all, we're not breaking anything, given that `global` not in first position wasn't working at all formerly!)

The other advantage of your suggestion here is that it keeps the diff minimal compared to the original code, which will greatly improve the chances of this PR to get reviewed & merged😁 

Thanks a lot for your help on this!

src/htmx.js
```diff
@@ -1141,31 +1141,53 @@ var htmx = (function() {
...
+      } else if (selector === 'window') {
+        item = window
+      } else if (selector === 'body') {
+        item = document.body
+      } else if (selector === 'root') {
+        item = getRootNode(elt, !!global)
+      } else if (selector.indexOf('global ') === 0) {
+        parts.unshift(selector.slice(7))
+        result.push(...querySelectorAllExt(elt, parts.join(','), true))
+        parts.length = 0
```
MichaelWest22, line 1174: Sorry I this line was not ideal after i posted my commit.  this should be a break instead so its easier to read the intent.

Telroshan:
After discussion with @1cg, here's the direction this PR should take:
- add a syntax support for enclosing comma-selectors with extended ones, in the format `closest <input, p/>` (cf `normalizeSelector`)
- keep `global` as a "apply to all" first-position-only keyword, and don't handle it in the loop at all (so, preserve current behavior, but don't introduce any new, we'll see if/when people using web components request such a feature, how we should implement it)

I'll work on these changes when I have the chance

MichaelWest22:
You might be able to add support for the `< />` enclosed extended selectors with a change like this:

```javascript
    let selectors = selector
    forEach(selectors.match(/<.*?\/>/g), function(enclosed) {
      selectors = selectors.replace(enclosed, enclosed.replaceAll(',', '%2C'))
    })
    const parts = selectors.split(',')
```
Trying to split on the commas but ignore the commas in the enclosed tags is very complex to do with just a regex so here I've just used the most basic regex to find all enclosed portions of the full selectors input string and then for each one replace all the commas with an encoded comma and update this in the main selectors string.  Finally you can just add the reverse of this comma encoding in the normalizeSelector() function

https://github.com/bigskysoftware/htmx/commit/cf10b71bb0e0bd4cc47a4ef9ca23d501087b0fce example commit

One thing to note is the next/previous/closest/find extended selectors only return a single node by design so supporting multiple css selectors via the list `,` combinator will only return the first item found that matches one of the selectors that comes first.  But they could now also do `next input, next div` instead of `next <input, div/>` if they wanted the next input AND div instead of next input OR div.

1cg:
I'm fine w/ a loop-based mini-lexer, the logic isn't too brutal (basically count "<" and "/>") and ignore commas when count > 0) but I'm happy with whatever solution y'all come up w/

Telroshan:
Damn, that's an interesting benchmark, thanks for looking into it @MichaelWest22 ! I wouldn't have expected it to be such a difference vs using substring
Just added an extra length check to avoid overflowing out of the array

MichaelWest22:
fun fact.  unlike most languages there is no such thing as out of bounds array access error on strings/arrays in javascript.  strings/arrays get auto extended with undefined if you read out of bounds so such checks are never needed if you already deal with the undefined case as `=== '>'` already does for us here.

you could also do
```javascript
        } else if (char === '>' && selector[i-1] === '/') {

or if you have to keep the out of bounds check for some reason:

        } else if (char === '>' && i > 0 && selector[i-1] === '/') {
```

1cg:
@Telroshan can you update this PR against `dev` and resolve the (hopefully small) conflict?

Telroshan:
Sure, done @1cg 
Pushed a `npm run format` fix along, as otherwise the CI fails on the linting step

Issue #2912, opened 2024-09-16
HTMX does not respond to events from elements appended to a form by HTMX


brackendawson:
I found this while making a form which has an array of options, I wanted to add more selectors to the form as the user wants to specify the option more times. I've found that events only work for the original form elements.

## Steps to recreate
This unit test:
```javascript
  it('handles events from appended form elements', function () {
    this.server.respondWith('GET', '/button', '<button id="new-button" class="pressy">Click me!</button>')
    this.server.respondWith('GET', '/test', 'Clicked!')

    var frm = make(`<form hx-get="/test" hx-trigger="click from:.pressy">
      <button id="get-button" hx-get="/button" hx-swap="afterend">Get the button I should click!</button>
    </form>`)

    var btn = frm.querySelector('#get-button')
    btn.click()
    this.server.respond()

    btn = frm.querySelector('#new-button')
    btn.click()
    this.server.respond()

    frm.innerHTML.should.equal('Clicked!')
  })
```

## Expected outcome
The form innerHTML will ultimately be replaced with "Clicked!"

## Actual outcome
Clicking the "Click me!" button has no effect.

## Debug
A version where the element was in the form to begin with works as you would expect:
```javascript
  it('handles events form original form elements', function () {
    this.server.respondWith('GET', '/test', 'Clicked!')

    var frm = make(`<form hx-get="/test" hx-trigger="click from:.pressy">
      <button id="get-button" hx-get="/button" hx-swap="afterend">Get the button I should click!</button>
      <button id="new-button" class="pressy">Click me!</button>
    </form>`)

    var btn = frm.querySelector('#new-button')
    btn.click()
    this.server.respond()

    frm.innerHTML.should.equal('Clicked!')
  })
```

seanmorton:
I ran into something similar like this recently and https://htmx.org/api/#process may be your friend here, though I think you should also try adding `type=submit` to the new button to leverage htmx's default form triggers first.

brackendawson:
I have tried using htmx.process on the bottom and the form, it didn't help.

I'm my real form it's actually a <select> and not a <button>, so it can't have a type=submit attribute. But I do want to submit the form every time the select is changed.

Telroshan:
Hey, indeed hx-trigger is evaluated on load to bind event listeners but doesn't update automatically
As the [hx-trigger doc](https://htmx.org/attributes/hx-trigger/) states:
[...]

PR #2914, opened 2024-09-17
Clarify event filter uses in hx-trigger

brackendawson:
## Description
* Clarify the example use of an event filter so that it is clear that you still use the `hx-trigger` attribute and that you use a CSS selector.

Corresponding issue: #2912 

## Testing
*Viewed the document as rendered markdown.

## Checklist

  source changes)
  approved via an issue
  Two unit tests time out on my machine.

### DIFF
--- a/www/content/attributes/hx-trigger.md
+++ b/www/content/attributes/hx-trigger.md
@@ -56,7 +56,7 @@ is seen again it will reset the delay.
 is seen again before the delay completes, it is ignored, the element will trigger at the end of the delay.
 * `from:<Extended CSS selector>` - allows the event that triggers a request to come from another element in the document (e.g. listening to a key event on the body, to support hot keys)
   * A standard CSS selector resolves to all elements matching that selector. Thus, `from:input` would listen on every input on the page.
-  * The CSS selector is only evaluated once and is not re-evaluated when the page changes. If you need to detect dynamically added elements use an event filter, for example `hx-trigger="click[event.target.matches('button')]"`. Unlike a standard CSS selector, an event filter will only catch events from the element the `hx-trigger` attribute is defined on or its descendants, it can be combined with `from:body` to catch events from every button on the page. 
+  * The CSS selector is only evaluated once and is not re-evaluated when the page changes. If you need to detect dynamically added elements use an event filter, for example `hx-trigger="click[event.target.matches('button')]"`. Unlike a standard CSS selector, an event filter will only catch events from the element the `hx-trigger` attribute is defined on or its descendants, it can be combined with `from:body` to catch events from every button on the page. Event filters [require eval](@/docs.md#configuration-options).
   * The extended CSS selector here allows for the following non-standard CSS values:
     * `document` - listen for events on the document
     * `window` - listen for events on the window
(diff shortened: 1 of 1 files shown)


Telroshan:
Hey, tbh I'm not so sure about the `<CSS selector>` here, I'm thinking that it's not a bad thing that an _example_ uses any given tag, to precisely provide a "real" example.

However, this section could probably be improved to make things clearer.
For example, it could be useful to make it clear that `hx-trigger="click[event.target.matches('input')]"` will indeed work for any `input` element, but within the element declaring the hx-trigger itself only! As the `click` event listener will catch `click` events fired from the root element or any descendant, but not other elements on the page (maybe add a reference to [`event bubbling`](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Event_bubbling) in here too?).

Maybe it could be useful to add this precision, as for ex a `hx-trigger="click from:input"` will bind to _all_ `inputs` on the page, while `click[event.target.matches('input')]"` will only work on child inputs, in which case you would more want to do something like `click[event.target.matches('input')]" from:body`

Just some ideas out loud, let me know what you think!

Telroshan (commented):
Ok so very sorry about it, but I just realized I forgot to take the whole context into account while reviewing this.

As we're in the `hx-trigger` doc, though more specifically in the `from` details here, I feel like it might be too much information for this place.
Also, it feels weird to me that the docs goes from `from` to event filtering to then referencing combining with `from` once again while we're still in the `from` section's itself

What would you think about:
- Moving the `eval` reference to the "standard event filters" above that describes how filtering works ?
- Add a link to the event filtering section in the `from` details using a link anchor _(i.e `#standard-event-filters` to match the id of that part and auto-scroll to it on click)_
- Maybe simply illustrate the initial sentence here by 2 examples, something like
> for example `hx-trigger="click[event.target.matches('button')]"` would intercept any dynamically swapped in button descendant as the listener is registered on the root element, and `hx-trigger="click[event.target.matches('button')]" from:body` would intercept any dynamically swapped in button across the whole page we'd be listening on the body
Rough idea, likely too verbose, we'd probably want a more concise sentence than that but you get the idea!
- Or, we could as well put more examples in the event filters section, precisely one with a `event.target.matches` example, then just refer to it in the `from` section maybe?

Let me know your thoughts!

brackendawson:
No worries at all on the back and forth, lets do this well. I like the above, though I'm thinking for the example in "from" we only use the second, one which is most equivalent.

www/content/attributes/hx-trigger.md
```diff
@@ -56,7 +56,7 @@ is seen again it will reset the delay.
 is seen again before the delay completes, it is ignored, the element will trigger at the end of the delay.
 * `from:<Extended CSS selector>` - allows the event that triggers a request to come from another element in the document (e.g. listening to a key event on the body, to support hot keys)
   * A standard CSS selector resolves to all elements matching that selector. Thus, `from:input` would listen on every input on the page.
-  * The CSS selector is only evaluated once and is not re-evaluated when the page changes. If you need to detect dynamically added elements use an event filter, for example `click[event.target.matches('input')]`
+  * The CSS selector is only evaluated once and is not re-evaluated when the page changes. If you need to detect dynamically added elements use a [standard event filter](#standard-event-filters), for example `hx-trigger="click[event.target.matches('button')] from:body"` which would listen on every button on the page.
```
Telroshan, line 59: If you allow me a last nitpicking, technically this listens on the body and not the buttons _(the `addEventListener` is on the body)_, but it will... catch? intercept? (those are the 2 words that come to mind but there may be a better one to describe it) the button `click` events.

I don't think we need to explain what event bubbling here is (which could make the sentence too verbose), but maybe we can just insert a link on the verb you will choose, that points to the [MDN event bubbling doc](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Event_bubbling) ?

Telroshan (approved):

Telroshan:
Thanks for your patience!

Issue #2913, opened 2024-09-16
hx-disabled-elt causes unhandled exception when targets not found


dwasyl:
Hi there,

Following along with the example of settings for `hx-disabled-elt`, and taking advantage of the inheritance, when the elements specified in the `hx-disabled-elt` are not found htmx fails with  `Uncaught TypeError: Cannot read properties of null (reading 'htmx-internal-data')`.

After a fair bit of debugging, the issue comes from a setup like this:
```
<div class="parent-container" hx-disabled-elt="find input[type='text'], find button">
  <form hx-post="<SUBMIT URL>">
     <input type="text" name="name1">
     <button type="submit">Submit</button>
    </form>
    <a hx-get="<URL>" href="#" hx-target="#some-div">Do something</a>
</div>
```

In the above example, the elements in the form are disabled and work great, when when the second link it pressed, because there are no child elements of `input` or `button` for find to work on, the uncaught exception is generated.

When the elements specified in a `hx-disabled-elt` are not found (or not children exist) the failure should be more graceful, especially since as an inheritable property it may be used in a number of different circumstances.

This is occurring with htmx 2.0.2.

MichaelWest22:
yeah interesting bug.  I've tracked it down to the querySelectorAllExt() function which for the `closest` `find` `next` and `previous` css selector lookups it does not handle miss lookups.  What is happening is these do a querySelector() function that returns one element or null if not found but this is then converted to an array by wrapping in [] because the querySelectorAllExt() function can also be used for multiple item lookups and [null] as the result causes the problems.

also note that `hx-disabled-elt="find input[type='text'], find button"` is not working properly as this attribute does not support multiple lookups so everything after the comma is ignored I think.  But it still breaks with that error with just the one find.

You can work around this issue in your case by not using find and instead finding the components in another way.  for example you could add a custom class to all elements you want to disable like "hx-disable" and then use ".hx-disable" as the selector which allows it to find multiple avoiding the , issue and the exception issue.

dwasyl:
> yeah interesting bug.  I've tracked it down to the querySelectorAllExt() function which for the `closest` `find` `next` and `previous` css selector lookups it does not handle miss lookups.  What is happening is these do a querySelector() function that returns one element or null if not found but this is then converted to an array by wrapping in [] because the querySelectorAllExt() function can also be used for multiple item lookups and [null] as the result causes the problems.

[...]

PR #2915, opened 2024-09-17
[bug] Handle missing selectors for closest/find/next/previous

MichaelWest22:
## Description
when using some of the custom selector modes like find etc it now wraps the response in an array.  But when there happens to be no elements found it then wraps the null which will throw exceptions up the chain so we need to get it to return empty array instead.

Corresponding issue:
#2913 

## Testing
Added test and did manual testing to confirm find etc does not throw errors anymore when no elements are found

## Checklist

  source changes)
  approved via an issue


Telroshan (approved):
Nice, simple & effective fix!

MichaelWest22:
If we merge in #2902 first then this PR would need to be updated to use `result.push(...toNodeArray(xxxx))`

MichaelWest22:
actually to merge this fix into #2902 you can simplify it down to this:
```javascript
let item
      if (selector.indexOf('closest ') === 0) {
        item = closest(asElement(elt), normalizeSelector(selector.substr(8)))
      } else if (selector.indexOf('find ') === 0) {
        item = find(asParentNode(elt), normalizeSelector(selector.substr(5)))
      } else if (selector === 'next' || selector === 'nextElementSibling') {
        item = asElement(elt).nextElementSibling
      } else if (selector.indexOf('next ') === 0) {
        item = scanForwardQuery(elt, normalizeSelector(selector.substr(5)), !!global)
      } else if (selector === 'previous' || selector === 'previousElementSibling') {
        item = asElement(elt).previousElementSibling
      } else if (selector.indexOf('previous ') === 0) {
        item = scanBackwardsQuery(elt, normalizeSelector(selector.substr(9)), !!global)
      } else if (selector === 'document') {
        item = document
      } else if (selector === 'window') {
        item = window
      } else if (selector === 'body') {
        item = document.body
        item = getRootNode(elt, !!global)
      } else {
        // Push back the unaltered string part, so that the standard selector fallback uses the exact initial string
        unprocessedParts.push(selector)
      }
      item && result.push(item)
```
Where you change it from pushing to the array each time to assign to a let variable and then only push at the end if item is not null or undefined which handles the null case well.

Telroshan:
Yep makes sense @MichaelWest22 
Given that #2902 already generates a diff on all lines of that function, I think I'll directly embed your fix in here as I don't expect it to make reviewing any harder

Telroshan:
Merged your changes in #2902, closing this one!

Issue #2767, opened 2024-07-29
hx-disabled-elt problem with hx-trigger="load"


bearpreis:
As of htmx version 2.0.0 the behaviour of `hx-disabled-elt` in combination with `hx-trigger=“load”` is broken. The disabled attribute is not removed after the request.

Expected behaviour would be that the element is only disabled while the htmx request is running.

Example (will work, if loading version 1.9.9):
```
<html>
    <head>
        <script src="https://unpkg.com/htmx.org@2.0.1"></script>
    </head>
    <body>
        <span hx-post="/foo" hx-disabled-elt="#submit" hx-trigger="load">test</span>
        <button type="submit" id="submit">Save</button>
    </body>
</html>
```

Our current workaround is using `intersect` instead of load in these cases.

Telroshan:
Hey, just fiddled with your example above and it looks like this bug was introduced in htmx 1.9.10, as it works with 1.9.9 but with 1.9.10, has the same issue than with htmx 2.
We would have to dive in the commits to find where/when this was introduced, and why.
If you feel like investigating, a bugfix PR is welcome! Otherwise, I'll try to find time in the upcoming days to look into this

PR #2925, opened 2024-09-23
[bug] load trigger stops hx-disabled-elt getting re-enabled

MichaelWest22:
## Description
When we init nodes and find a "load" trigger it does the request straight away and this request may disable some elements with hx-disabled-elt.  However if these disabled elements have not yet had their nodeInit() run yet because they are placed lower in the page then when they later init they first de-init their internalData which removes the requestCount counter used to re-enable them again.  So we need it to handle situations where requestCount is not set and currently it defaults this to -1 and no 0 because of the decrement. So to solve this I've just changed it to fall back to 1 which means that if either requestCount is undefined or if it is equal to 0 it will replace it with 1 and then after the decrement will be 0 and trigger the proper code to enable the element again.

Corresponding issue:
#2767 
## Testing
Wrote a test to reproduce the issue with disabled-elt and load together and did a quick test in my test project to prove the disabled button re-enables as expected.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2549,7 +2549,11 @@ var htmx = (function() {
         handler(elt)
       }
     }
-    getWindow().setTimeout(load, delay || 0)
+    if (delay > 0) {
+      getWindow().setTimeout(load, delay)
+    } else {
+      load()
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Telroshan (approved):
This counter decrement was added in https://github.com/bigskysoftware/htmx/commit/c1bf3fa686468239b53da81dbb1a599e5d2fd9eb
I won't speak for @1cg but I would guess that this was probably an overlook/copy-paste at the time ? The `|| 0` works well for the addition as we take it from 0 to 1, which makes sense, just serves as a "if undefined then 0"
But for the decrement, I don't see any reason why we'd want to reach a negative value for the counter, we'd instead want to reach 0 just like you did in this PR.

So, with the little I know _(I wasn't there when it was first written)_, I would totally approve this change

1cg:
Yep, just me being stupid

PR #2926, opened 2024-09-23
Improved documentation for htmx:confirm event and examples for implementing sweetalert

eduardolat:
## Description

This PR is an update of the #2868 PR, sorry for the delay but I was very busy with my work.

I've read @Telroshan comments and have clarified it further so that everyone understands how it works.

I've also modified the file www/content/examples/confirm.md since the example didn't work correctly and I've added some comments clarifying how it works.

I hope this time everything is much clearer, I look forward to your comments, thank you very much!

## Checklist

  source changes)
  approved via an issue


alexpetros:
@Telroshan this looks reasonable to me, let me know what you think

www/content/events.md
```diff
@@ -179,43 +179,69 @@ than a single value.
...
+  //    attribute on it, if it doesn't we can return early and let the default behavior happen
+  if (!evt.target.matches("[confirm-with-sweet-alert='true']")) return
+
+  // 3. Prevent the default behavior (this will prevent the request from being issued)
+  evt.preventDefault();
+
+  // 4. Show the sweet alert
+  swal({
+    title: "Are you sure?",
+    text: "Are you sure you are sure?", // or evt.detail.question (only available if hx-confirm attribute is present)
```
Telroshan, line 222: I'm going to nitpick a bit here, but since in that very example you check above if the element has a `hx-confirm` attribute, and early return if it doesn't, I would say to either A) remove the early return and let this one as is, or B) keep the early return and then simply directly use `evt.detail.question` in here.
My gut guess would be that A) would make for the easiest-to-follow example but I'll let you be the judge!

It's just that, looking at this example, I wonder what's the point of `hx-confirm` at all? Since we are using a custom attribute `confirm-with-sweet-alert` to determine whether we should confirm with sweet alert or not, and that we're overriding the `hx-confirm` behavior anyway, it seems to me that it relegates `hx-confirm` to be a simple value-holder attribute for the message.
In that same example, why would one not want to simply put their confirm message as the value of `confirm-with-sweet-alert` and ignore `hx-confirm` altogether (as the former example was doing, except the message was even hardcoded in the JS) ?

Telroshan (commented):
Good inline comments to clarify, and it's cool that you added a basic example to instruct the general process!
Just allow me one question and I'll stop bothering you 😆

www/content/events.md
eduardolat: You are completly right on this, i propose to add a third example, so we have this (in order):

### 1. Basic usage
Just the basic example I added with the comments for people to use at their discretion

### 2. Sweetalert with hx-confirm
I think it would be very useful to keep an example where people can continue using hx-confirm without having to add additional attributes which they have to remember, as is the case with `confirm-with-sweet-alert`, this way we show the usefulness of `evt.detail.question`.

- **Pro:** Users would simply use `hx-confirm` as usual, even in teams with multiple developers there will be less friction because they don't have to know that a new attribute exists and would simply use `hx-confirm` as usual.
- **Con:** Not really a con, but the default behavior of `hx-confirm` would be overwritten

### 3. Sweetalert with custom attribute `confirm-with-sweet-alert`
This example would serve to show people how to add their custom confirms without having to remove the default behavior of `hx-confirm`, the only disadvantage is that they have to memorize a new attribute

What do you think about this @Telroshan @alexpetros ? I think that with this we would not leave anything unclear.

eduardolat: I also propose that we change the attribute `confirm-with-sweet-alert` to `hx-confirm-swal`, I think it's a bit clearer, shorter and can save a lot of headaches when trying to memorize it

alexpetros: We shouldn't use an attribute name that starts with `hx-`, that seems confusing. I think the current name on that one is fine.

eduardolat: Okay, discarded, is everything else okay?

Telroshan: > can save a lot of headaches when trying to memorize it

Memorizing this attribute won't be a problem because the user can name it however they want in their app, they just have to memorize it throughout the example in the docs, which will be short enough that it won't be an issue.
I agree with alex that prefixing this with `hx-` could lead people to believe this is a standard attribute of the lib, which would only add more confusion.

I'm never against more examples, sounds good to me to add a third one like you suggested @eduardolat !

alexpetros: @Telroshan I actually am against more examples in this case—I think there's a lot of value to keeping the documentation a little bit tighter and we're already going out a bit on a limb with the sweet alert custom stuff. What if we just merge what's here and take that win for now?

Telroshan: I don't have any strong opinion tbh @alexpetros, it's true that this is the `events.md` doc and not the `confirm` one, so it could be a bit too heavy for this context.
And after giving it a second thought, that third example is about "using custom attributes to do whatever you want" compared to the second, which isn't htmx-specific after all.

I would still insist on my first remark in this thread before merging this PR though @alexpetros !
> I would say to either A) remove the early return and let this one as is, or B) keep the early return and then simply directly use evt.detail.question in here.

eduardolat: @Telroshan @alexpetros I have modified it according to the requested modifications, trying to make it as clear as possible. I look forward to your comments.

Telroshan (approved):

www/content/events.md
Telroshan: Looks good, thx!

PR #2928, opened 2024-09-23
fix moveBefore pantry losing content with hx-swap none or hx-select

MichaelWest22:
## Description
The new experimental moveBefore to pantry code for the hx-preserve attribute has an issue where if you use hx-swap="none" or the hx-select feature the content moved out to the htmx-preserve-pantry will not be swapped into the main content as expected and therefore it will have no anchor element in the body left to restore the preserved element.  So to fix this we need to leave a dummy element in the preserved elements original location that will be replaced by the preserved element at the end of the swap.

Just need to create the following simple element next to the element before we move it out:
`<div id="${id}"></div>`

Corresponding issue:

## Testing
Performed manual testing to confirm it works and adjusted the move-before manual test to test a none swap that reproduced the issue where before the fix the preserved video would be removed.  Also did some performance testing in chrome to test what element type worked best as the placeholder element as I considered using `<template>` because it does not render anything visible but div seemed a little faster. 

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1493,8 +1493,8 @@ var htmx = (function() {
             getDocument().body.insertAdjacentHTML('afterend', "<div id='--htmx-preserve-pantry--'></div>")
             pantry = find('#--htmx-preserve-pantry--')
           }
-          // @ts-ignore - use proposed moveBefore feature
           existingElement.insertAdjacentHTML('beforebegin', `<div id="${id}"></div>`)
+          // @ts-ignore - use proposed moveBefore feature
           pantry.moveBefore(existingElement, null)
         } else {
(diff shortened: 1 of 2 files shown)


MichaelWest22:
More investigation and I think I'm wrong about this being anything to do with then new moveBefore as the problem with hx-swap none an hx-select is in fact the same as the existing hx-preserve feature on normal browsers.  So i don't think we should fix this like i have suggested.  Some people could have been relying on hx-preserve to move the preserved content to a different location with a partial response and this fix actually blocks his use case.

MichaelWest22:
Wrote a better fix that handles more edge cases like if the item is relocated during the replacement.  But I can't see any way to address this issue with non moveBefore supporting browsers so I don't think this fix is needed right now. closing for now

PR #2933, opened 2024-09-25
fix #2932: check parent is null for swap delete

### LINKED ISSUE #2932: Using hx-swap="delete swap:1s", element is possibly deleted
When using a `hx-swap="delete swap:1s"`, the target element could have been deleted already.

To reproduce simply do:

```js
<div id="add-users-modal" hx-replace>
  <button
		  hx-post="/add"
		  hx-target="#add-users-modal"
		  hx-swap="delete swap:1s"
		  @htmx:after-request="show=false"
		  type="button"
  >add<button>
</div>
<script>
document.addEventListener("htmx:beforeSwap", (event) => {
  const content = event.detail.serverResponse;
  if (content.includes("hx-replace")) {
    const parser = new DOMParser();
    const doc = parser.parseFromString(content, "text/html");
    const id = doc.querySelector("[hx-replace]").id;
    const el = event.detail.target.querySelector(`#${id}`);
    if (el) {
      el.remove();
    }
  }
});
<script>
```
Here `hx-replace` is an attribute I defined myself to replace if present else append.

In such use cases, I'll get `htmx.esm.js:1699 Uncaught TypeError: Cannot read properties of null (reading 'removeChild')`

image

jackielii:
## Description

Corresponding issue: fix #2932

## Testing

## Checklist

  source changes)
  approved via an issue


Telroshan:
Hey, could you add a test case, just to ensure that the issue happens without your fix when running the test suite, then that your PR fixes it ? (also to prevent any future regression on this), thanks !

jackielii:
Hi @Telroshan ,

I added test, and found out about the outerHTML swap issue. So fixed that and added a test for it as well.

Let me know if you need anything else.

Telroshan:
Thanks, though if I run your tests locally without your PR's fix (thus with the current library's state), the test seem to already pass 😬 
image

The tests should be setup so that they fail with the current state of the lib, and pass with your fix (to prove the resolution, and ensure we don't introduce regressions in the future).
So in their current state, the 2 new tests appear to be "useless" in the way that they don't fail with the buggy lib behavior in the first place.

It's weird because I don't seem to even get an exception in the console. Didn't spend time investigating this so I can't be of much more help right now unfortunately!

jackielii:
Hmm, that's weird, I clearly run the tests without the fix and they fail. Maybe it's my last minute clean up. Will check and modify

jackielii:
hi @Telroshan , I realised I removed the `div.remove()` because I thought remove parent would remove the div. Turns out not to be the case. so the `div.remove()` is needed in the test. Should be all good now

Telroshan (approved):
Yep all good, tests correctly fail now with the current lib's state, then pass with your fix, thanks!

Issue #2922, opened 2024-09-20
hx-preserve does not work in oob-swapped content


irfinnew:
I'm having an issue where an element with `hx-preserve` set is not actually preserved if the element is updated through an OOB swap.

Proof of concept code on the main page:
```html
<div id="foo" hx-get="partial.html" hx-trigger="load"></div>

<p><button onclick="document.getElementById('bar').innerHTML = 'goodbye';">change</button></p>
<p><button hx-get="partial.html" hx-swap="none">update</button></p>

partial.html:
<div id="foo" hx-swap-oob="true">
	<p><span id="bar" hx-preserve="true">hello</span></p>
</div>
```

Pressing the `change` button changes `span#bar`'s content (expected), but `update` reverts this change (unexpected).

The same PoC _does_ work with a regular targeted non-OOB swap:

```html
<div id="foo" hx-get="partial.html" hx-trigger="load"></div>

<p><button onclick="document.getElementById('bar').innerHTML = 'goodbye';">change</button></p>
<p><button hx-get="partial.html" hx-target="#foo">update</button></p>

partial.html:
<div id="foo">
	<p><span id="bar" hx-preserve="true">hello</span></p>
</div>
```

I tested with htmx 1.8.6, 1.9.12, and 2.0.2, on both Chrome and Firefox. The OOB version didn't seem to work on any combination of them.

btw, thank you for making web development fun again :)

Telroshan:
Hey, looking at the code, it seems that what's happening is:
- oob swapping is processed first
https://github.com/bigskysoftware/htmx/blob/b71af75f1af7d5d3ae562686625fe131f6923e5a/src/htmx.js#L1856
- it removes the element from the response fragment after swapping it
https://github.com/bigskysoftware/htmx/blob/b71af75f1af7d5d3ae562686625fe131f6923e5a/src/htmx.js#L1455
- only then, elements to preserve are processed (which happens _before_ the standard swap)
https://github.com/bigskysoftware/htmx/blob/b71af75f1af7d5d3ae562686625fe131f6923e5a/src/htmx.js#L1872-L1873

While it looks like we never supported `hx-preserve` along oob swap, I don't see why we wouldn't want to, and I'd consider this as a bug since the documentation doesn't mention any exception about oob swapping so one would expect it to work just like standard swapping.
So, if you'd like to investigate and submit a PR, please feel free to do so!

MichaelWest22:
[...]

PR #2934, opened 2024-09-27
enable hx-preserve handing for oob swaps

MichaelWest22:
## TL;DR

As found in issue #2922, htmx does not process hx-preserve attributes properly when supplied inside a oob-swap response.  Instead the preserved content is just overwritten with the new content from the oob response.
So I've found that simply wrapping the oobSwap() function with handle and restore preserved elements we can indeed support this with just these two lines.  

## Description

I also tried with simply moving the handlePreservedElements() funciton up before the oob swap handling code and this in fact does mostly work but it has some bad side effects as when it preserves the element into the fragment it removes the hx-swap-oob tags if done wrong and it can bring oob-swap tags from the page into the fragment which can then act as oob swaps in error.  So to avoid this issue I moved the change to just wrap inside the oobSwap() function and it behaves much better.  

What i've found is that with this change their is no way to support hx-preserve on the same element as the hx-swap-oob like:

```html
<div id="foo" hx-swap-oob="true" hx-preserve>dummy data</div>
``` 
This just can't work for several reasons.  I have failed to find a valid reason to even do this as you are asking it to swap foo back into foo with no changes so it is just a null update at best. Some of the other swap styles with additional targeting might do something meaningful but it not only breaks my brain trying to work out how it would work but it also breaks the code if you try and make it work.  But the good news is that the code by default uses a findAll() to run a querySelectorAll() to find all children that match the hx-preserve attribute and this ignores the attribute on the root element of the oob swap. So this is a reasonable safe default to have where it prevents you doing the wrong kind of use case and just treats it as a hx-swap-oob with no preserve if you try this.  Could add code to report an error and ignore the swap if needed but I think this is better fixed with correct documentation.

I also Found some interesting emergent behavior with how hx-preserve works. It base use case is to keep some important client side content from changing while the whole page gets updated and re-arranged around it and this does allow when doing full page reloads like boost's to re-parent the preserved elements to a new page location.  And with standard htmx partial page updates if you include a hx-preserve item in your partial response it can in fact move the preserved item from another location on the page into the new target location allowing you to for example drag a video iframe to a different page location etc.  And with oob swaps you also have this same new re-parenting super power:

```html
Main page content<div id="topleft" hx-swap-oob="beforeend"><div id="video" hx-preserve></div></div>
```
With this simple response we can relocate video element to the end of the topleft div from anywhere on the page without it changing.

```html
<p>Item 6 moved above item 3</p><div id="list-item-3" hx-swap-oob="beforebegin"><div id="list-item-6" hx-preserve></div></div>
```
And here you can move a preserved item up to above item 3 in some list without having to know update or change its contents.

https://michaelwest22.github.io/htmx/index.html here is a quick example of using this oob swaps in practice.  it works best with experimental moveBefore enabled.

I think I just need to add some documentation to the hx-preserve.md page to highlight the basic behaviour when using hx-swap-oob.

Corresponding issue:
#2922

## Testing
Added tests for the basic use cases and performed testing on both canary chrome with moveBefore and normal chrome to confirm oob swaps work as expected.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1420,6 +1420,7 @@ var htmx = (function() {
    * @returns
    */
   function oobSwap(oobValue, oobElement, settleInfo) {
+    handlePreservedElements(oobElement)
     let selector = '#' + getRawAttribute(oobElement, 'id')
     /** @type HtmxSwapStyle */
     let swapStyle = 'outerHTML'
@@ -1462,6 +1463,7 @@ var htmx = (function() {
       oobElement.parentNode.removeChild(oobElement)
       triggerErrorEvent(getDocument().body, 'htmx:oobErrorNoTarget', { content: oobElement })
     }
+    restorePreservedElements()
     return oobValue
   }
 
@@ -1479,7 +1481,7 @@ var htmx = (function() {
   }
 
   /**
-   * @param {DocumentFragment} fragment
+   * @param {DocumentFragment|Element} fragment
    */
   function handlePreservedElements(fragment) {
     forEach(findAll(fragment, '[hx-preserve], [data-hx-preserve]'), function(preservedElt) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1420,7 +1420,6 @@ var htmx = (function() {
    * @returns
    */
   function oobSwap(oobValue, oobElement, settleInfo) {
-    handlePreservedElements(oobElement)
     let selector = '#' + getRawAttribute(oobElement, 'id')
     /** @type HtmxSwapStyle */
     let swapStyle = 'outerHTML'
@@ -1451,7 +1450,9 @@ var htmx = (function() {
 
           target = beforeSwapDetails.target // allow re-targeting
           if (beforeSwapDetails.shouldSwap) {
+            handlePreservedElements(fragment)
             swapWithStyle(swapStyle, target, target, fragment, settleInfo)
+            restorePreservedElements()
           }
           forEach(settleInfo.elts, function(elt) {
             triggerEvent(elt, 'htmx:oobAfterSwap', beforeSwapDetails)
@@ -1463,7 +1464,6 @@ var htmx = (function() {
       oobElement.parentNode.removeChild(oobElement)
       triggerErrorEvent(getDocument().body, 'htmx:oobErrorNoTarget', { content: oobElement })
     }
-    restorePreservedElements()
     return oobValue
   }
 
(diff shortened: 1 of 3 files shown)


irfinnew:
Thanks! I was intending to look at this myself at some point, but I'm not much of a Javascript dev, so I'm grateful someone else fixed it.

I tested master+PR in my project and I can confirm it solves my issue.

Telroshan (approved):
Simple & effective fix, good catch!

_(Note: I took the liberty of isolating a `TL;DR` part at the top of your PR description, just to maximize its chances of being read through upon review, hope you don't mind!)_

gilesbradshaw:
I wonder what I am missing as i cant get this to work.  I am using https://michaelwest22.github.io/htmx/htmx.js which is code from demo.

I swap: 
```
<div id="will-be-oobed">
    <div id="will-be-oob-preserved" hx-preserve="true">
      ah
      <b>hoping</b> all this will remain..
    </div>
  </div>


```

with

```
<div id="will-be-oobed" hx-swap-oob="innerHTML">
  <div>more stuff</div>
</div>

```

but it doesn't preserve..

irfinnew:
You need to have `div#will-be-oob-preserved` in your fragment response as well, see the [hx-preserve docs](https://htmx.org/attributes/hx-preserve/):

> You must set an unchanging id on elements for hx-preserve to work. The response requires an element with the same id, but its type and other attributes are ignored.

Something like this should work. Original HTML:

```
<div id="will-be-oobed">
    This will change.
    <div id="will-be-oob-preserved" hx-preserve="true">
        This will stay untouched.
    </div>
</div>
```

OOB replacement:
```
<div id="will-be-oobed" hx-swap-oob="innerHTML">
    Changed!
    <div id="will-be-oob-preserved" hx-preserve="true">
        You should not see this!
    </div>
</div>
```

MichaelWest22:
@Telroshan if I wanted to add a few lines to hx-preserve.md documentation to make it clear how use oob swap with hx-preserve and its limitations should this be added to this PR or another separate PR?  And if it is another one should it target dev or master as this shouldn't merge until this PR possibly gets merged to master with a full release.

Telroshan:
@MichaelWest22 Since the documentation you're mentioning would depend on this PR to be merged first, and as it'd make even more clear what this PR is adding, I'd say it's totally fine & desired to add it to this very PR!

gilesbradshaw:
> (this is a reply to a now deleted comment by someone who couldn't get oob-preserve-swapping to work)


Thanks for your reply.  I deleted my question in the hope no one had seen me make a fool of myself :P

1cg:
@MichaelWest22 continues to crush

PR #2939, opened 2024-10-03
Add an example of evaluated values passed to hx-headers.

maciejtrybilo:
Add an example of evaluated values passed to hx-headers.

## Description
I haven't been clear from the documentation how to exactly have the javascript in the attribute evaluated as I'm not such a web dev expert. Perhaps it could help someone else if it's more explicit.

The example is more or less taken from my use case. Please feel free to suggest something else!

HTMX works great with Swift, BTW.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/attributes/hx-headers.md
+++ b/www/content/attributes/hx-headers.md
@@ -7,10 +7,14 @@ The `hx-headers` attribute allows you to add to the headers that will be submitt
 By default, the value of this attribute is a list of name-expression values in [JSON (JavaScript Object Notation)](https://www.json.org/json-en.html) 
 format.
 
+```html
+  <div hx-get="/example" hx-headers='{"myHeader": "My Value"}'>Get Some HTML, Including A Custom Header in the Request</div>
+```
+
 If you wish for `hx-headers` to *evaluate* the values given, you can prefix the values with `javascript:` or `js:`.
 
 ```html
-  <div hx-get="/example" hx-headers='{"myHeader": "My Value"}'>Get Some HTML, Including A Custom Header in the Request</div>
+  <body hx-headers='js:{ "Locale" : new Intl.DateTimeFormat().resolvedOptions().locale }'>...</body>
 ```
 
 ## Security Considerations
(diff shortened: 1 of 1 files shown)


www/content/attributes/hx-headers.md
```diff
@@ -7,10 +7,14 @@ The `hx-headers` attribute allows you to add to the headers that will be submitt
...
 
+```html
+  <div hx-get="/example" hx-headers='{"myHeader": "My Value"}'>Get Some HTML, Including A Custom Header in the Request</div>
+```
+
 If you wish for `hx-headers` to *evaluate* the values given, you can prefix the values with `javascript:` or `js:`.
 
 ```html
-  <div hx-get="/example" hx-headers='{"myHeader": "My Value"}'>Get Some HTML, Including A Custom Header in the Request</div>
+  <body hx-headers='js:{ "Locale" : new Intl.DateTimeFormat().resolvedOptions().locale }'>...</body>
```
Telroshan, line 17: I think to keep the example consistent, we could take inspiration from the [`hx-vals`](https://htmx.org/attributes/hx-vals/) documentation, which has the following example:

```html
<div hx-get="/example" hx-vals='{"myVal": "My Value"}'>Get Some HTML, Including A Value in the Request</div>

<div hx-get="/example" hx-vals='js:{myVal: calculateValue()}'>Get Some HTML, Including a Dynamic Value from Javascript in the Request</div>
```

I think it's a concise & effective approach, using the same key for the same context element, just changing _how_ the value is retrieved and illustrating how `js:` works.

Let me know your thoughts on this! Your opinion will have more value than mine on this as I'm likely biased by my htmx experience (as you said, we want the docs to be clear enough for newcomers!)

maciejtrybilo:
@Telroshan Makes sense, I wasn't aware that the hx-vals page does have an example. And sorry for taking so long!

Telroshan (approved):

Telroshan:
Thanks!

Issue #100, opened 2020-06-15
Events page missing on main menu.


zodman:
Not easy to find.

chg20:
Added a contents header:

https://htmx.org/reference/

PR #2951, opened 2024-10-07
Replace deprecated String.prototype.substr usage

scrhartley:
## Description
Inspired by the already merged in extensions [PR #100](https://github.com/bigskysoftware/htmx-extensions/pull/100).

On String, the substr method is deprecated and is not part of the main ECMAScript specification
(see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr).
Replace with the alternatives slice and substring, which are both already used in the codebase.

## Testing
I made sure that giving bad values caused the existing test cases to fail.
I temporarily added additionally logging to verify that it was doing what I expected.
I checked that most of the changes are trivial so that this could be considered as a single change.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1130,17 +1130,17 @@ var htmx = (function() {
   function querySelectorAllExt(elt, selector, global) {
     elt = resolveTarget(elt)
     if (selector.indexOf('closest ') === 0) {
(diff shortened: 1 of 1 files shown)


alexpetros:
Yup, seems like a good idea.

1cg:
image

Issue #2973, opened 2024-10-18
Ajax request sometimes triggered twice with hx-trigger="load" in web components


basvk:
I'm working with web components (with lit.dev) and using htmx to populate or replace them in the page. This saves me from writing a lot of business logic and networking code in the frontend. I'm not using shadow dom, and in general it works pretty well. I'm calling `htmx.process` inside the components on update.

Sometimes I'm seeing duplicate ajax requests occurring on (nested) elements with `hx-trigger="load"`, mainly on Safari 17.6.
In my case the response is swapped in with `hx-swap="afterbegin"`, causing duplicated children in the web component.
The element triggering the load is in a nested hierarchy. Top nodes do not show this behaviour.

Here-in lies the issue: in a regular web page `htmx.process` (or rather `processNode`) after initial page load is only called on the body, running initNode on all child elements with htmx attributes once.
With web components however, htmx.process is called for every web component, and every web component nested within it. This means children are visited multiple times, once for every parent component that calls htmx.process plus the child itself.
Luckily, this by itself is not a problem, because initNode checks if the node already has a corresponding initHash value and will not init again in that case.
However, the initHash is calculated from the node's attributes. Between the different calls to htmx.process from each webcomponent, the attributes sometimes do change, causing initHash to differ and to call initNode again on a node. 

In my scenario between calls to htmx.process, the element (when visited during `processNode` on a parent element) with hx-trigger="load" starts an ajax request and gets class="htmx-request" set. This alters the attributes, meaning that the element's initHash is not correct anymore for the attributes. This results in the node being initialized again during the next call to htmx.process (on another parent or itself) and starting a second ajax request...

Now I've solved this particular case by adding a check to not initialize if the node has `htmx-request` in its classList (see this commit: https://github.com/bigskysoftware/htmx/compare/master...basvk:htmx:fix-double-load - let me know if you'd like a pull request), but I'm not sure if there are other cases where multiple calls to htmx.process might cause these kind of issues.

Maybe initHash should be based on a subset of attributes, or calculated in a different way when dealing with web components? What is your opinion?

Telroshan:
Interesting issue, in this case I think you might want to define a delay to postpone the load trigger, with for example `hx-trigger="load delay:1ms"`, 1ms is already probably enough to get the load event out of the main queue and get all the children nodes processed before this event fires?

Though I agree that ideally you shouldn't need to do that, htmx should handle it itself by default.
[...]

PR #2976, opened 2024-10-20
Do not execute hx-trigger="load" on re-initialization of an existing node

basvk:
## Description
Sometimes, requests that should be only triggered once at load (with `hx-trigger="load"`) are executed twice. This happens for example when calling htmx.process on an existing node (or a parent node) after changing an attribute on the element with `hx-trigger="load"`.
This PR adds a `firstInitCompleted` flag to an element's internalData and adds a check for that flag when handling the `load` triggerEvent.

Corresponding issue:
#2973 

## Testing
Added a test scenario that changes `hx-swap="afterbegin"` to `hx-swap="beforeend"` on an element with `hx-trigger="load"`, followed by a call to `htmx.process` and checks whether the innerHTML of the element  stayed the same.
This test fails on the dev branch, and passes with this fix.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -693,6 +693,7 @@ var htmx = (function() {
    * @property {XMLHttpRequest} [xhr]
    * @property {(() => void)[]} [queuedRequests]
    * @property {boolean} [abortable]
+   * @property {boolean} [firstInit]
    *
    * Event data
    * @property {HtmxTriggerSpecification} [triggerSpec]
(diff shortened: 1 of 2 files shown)


Telroshan (commented):
Just an idea out loud, not a change request, what would you think of "reversing" the check to simplify the code, i.e. only process the `load` trigger if the `nodeData.something` property is falsy, and only set that property to true after initialization ?
This would remove the need for that first `true` assignment in `initNode` as well as the `undefined` comparison, as a falsy check would cover it (so only assign it to `true` after initialization, and check that it's falsy in `addTriggerHandler`)

basvk:
Thanks, good idea! I've changed `firstInit` into `firstInitCompleted` and removed that unnecessary check in initNode.

Telroshan (approved):

PR #2985, opened 2024-10-27
feat: trigger oobErrorNoTarget for missing oob swap targets

stenehall:
## Description
Currently the parent of an `oob-swap` will not get an `oobErrorNoTarget ` event when added to the `hxOn` event trigger.

Given the following code:

```html
<ul hx-ext="ws" ws-connect="/ws" id="hosts" hx-on:htmx:oob-error-no-target="console.log('error triggered')></ul>
```

And the following message from the server 

```html
<tr id="some-id" hx-swap-oob="outerhtml:#none-existing-id">
```

Nothing will happen. But with this change the `console.log` will run.

Corresponding issue:

## Testing
This was tested manually using a very similar code to the above. I was trying to figure out where a test would fit in `hx-swap-oob.js` but it wasn't immediately obvious to me. If you think the change is correct I will spend more time on figuring out a test for it.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1427,7 +1427,7 @@ var htmx = (function() {
     oobElement.removeAttribute('data-hx-swap-oob')
 
     const targets = querySelectorAllExt(rootNode, selector, false)
-    if (targets) {
+    if (targets && targets.length > 0) {
       forEach(
         targets,
         function(target) {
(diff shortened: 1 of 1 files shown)


scrhartley:
For existing projects that were already using this on body as a general error mechanism,
it seems that it will now trigger multiple times due to event bubbling.
Additionally, if the target is `body`, will it trigger twice, once for the general event and once for the target event?

stenehall:
> For existing projects that were already using this on body as a general error mechanism, it seems that it will now trigger multiple times due to event bubbling. Additionally, if the target is `body`, will it trigger twice, once for the general event and once for the settle target event?

Good questions. I'm not entirely sure how it will behave in relation to event bubbling nor body. But, I can spend a bit of time creating tests for this. If the general idea seems correct here.

To me the current behavior seems wrong, but the change might be a breaking change.

1cg:
half fixed in dev, don't want to trigger the event twice, thank you!

Issue #2995, opened 2024-11-03
Fix illegal invocation using FormData proxy


scrhartley:
Currently the `htmx:configRequest` event exposes `detail.parameters` as a proxied FormData.
When trying to use it for debugging with:
```js
element.addEventListener('htmx:configRequest', function(event) {
    console.log(new URLSearchParams(event.detail.parameters).toString());
});
```
I get:
> Uncaught TypeError: Illegal invocation

To work around this, it currently needs to be converted to entries:
```js
element.addEventListener('htmx:configRequest', function(event) {
    console.log(new URLSearchParams(event.detail.parameters.entries()).toString());
});
```

With a normal FormData object, it can be passed directly:
```js
let data = new FormData();
data.set('foo', 'bar');
console.log(new URLSearchParams(data).toString()); // Outputs foo=bar
```

I have a fix and will create a PR if that's OK: https://github.com/scrhartley/htmx/commit/bdd17de17a851052a9ba54eefb0aa991686196f0

Telroshan:
Hey, bugfix PRs welcome @scrhartley! Didn't think of this usecase when writing that proxy, good fix!

scrhartley:
Fixed by #2997

PR #2997, opened 2024-11-04
Fix illegal invocation for FormData proxy

scrhartley:
## Description
Currently the form data proxy has special handing for if the return value is a function,
but not in the case where the key is a symbol. This omission can cause an error to be thrown.

Fixes constructor for URLSearchParams:
```js
element.addEventListener('htmx:configRequest', function(event) {
    let params = new URLSearchParams(event.detail.parameters);
    console.log(params.toString());
});
```

If I was implementing this from fresh then I would use the more concise:
```js
if (typeof result === 'function') {
    return result.bind(formData)
}
```
but instead I've replicated the existing approach:
```js
if (typeof result === 'function') {
  return function() {
    return result.apply(formData, arguments)
  }
}
```

Corresponding issue: #2995

## Testing
Added a new test to the existing ones for parameters.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4039,7 +4039,15 @@ var htmx = (function() {
       get: function(target, name) {
         if (typeof name === 'symbol') {
           // Forward symbol calls to the FormData itself directly
-          return Reflect.get(target, name)
+          const result = Reflect.get(target, name)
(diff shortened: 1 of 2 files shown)


Telroshan (approved):

1cg:
thank you!

PR #3006, opened 2024-11-08
fix vertical alignment of github stars button in all states

untitaker:
## Description

Without uBlock, the button's alt-text would be 1px lower than the rest
of the nav items, and inconsistently capitalized.

With uBlock, the button would be slightly too high up. Screenshot of problem:

Screenshot from 2024-11-08 01-52-34
Screenshot from 2024-11-08 01-52-00

The solution is to apply styles on the `span` that GitHub's JS inserts.
However, using margin will make the entire nav jump during pageload, so
position: relative seems more robust there.

## Testing

Tested in Firefox and Chrome latest

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/themes/htmx-theme/static/css/site.css
+++ b/www/themes/htmx-theme/static/css/site.css
@@ -419,8 +419,9 @@ transition: visibility 0s 1s, opacity 1s linear;
   padding:1em 0;
 }
 
-.github-stars {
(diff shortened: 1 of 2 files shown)


alexpetros (approved):
okay, thanks!

Issue #3000, opened 2024-11-05
Wrong detail.elt for htmx:beforeSwap


benvcutilli:
Hi, I am trying to write an extension using the `htmx:beforeSwap` event, and it appears that line 4714:

`if (!triggerEvent(target, 'htmx:beforeSwap', beforeSwapDetails)) return`

ends up resulting in `detail.elt` being set to `target` (as `target` is passed into the first parameter of `triggerEvent`, which sets `beforeSwapDetails.elt` to it). However, other calls to `triggerEvent` pass in `elt` instead. Also, in the documentation at [https://htmx.org/events/#htmx:beforeSwap](https://htmx.org/events/#htmx:beforeSwap), it says that `detail.elt` is supposed to be "the element that dispatched the request". This makes more sense to me because it makes beforeSwap more useful (you get to additionally access the request element), but also because `detail.target` already exists and can be used.

Is this intended behavior? As stated earlier, it can be really difficult to take action on this event if you don't know which element sent the request.

Telroshan:
Hey, indeed it's likely a bug, especially as the documentation says the opposite of the current behavior.
As you said, we have `detail.target` available so it doesn't make much sense to double down with an extra reference to it in `detail.elt`.
Feel free to open a bugfix PR!

MichaelWest22:
From investigating why htmx is doing this i found this commit from 2020 https://github.com/bigskysoftware/htmx/commit/6f14cba2e65356844ae1bcb7ac5b2feb01dcffd1
It changes it from elt to target with the comment

Trigger events on new content, rather than triggering element, since triggering element may have been replaced

see #79 

It seems that often the element has already been replaced at this stage so elt is no longer a valid and so the event has to fire on the target that is being swapped instead of the old possibly removed element. which can cause events to get lost it seems.

So i think the documentation instead needs updating to make this 4 year old behavior clearer.

I also think you can probably reliably find the requesting element in the additional detail provided in the event.

try detail.requestConfig.elt

benvcutilli:
That commit isn’t for the same lines, I think? Besides, I don’t think this
is intuitive behavior; changing the semantics of the same name
(`detail.elt`) across events can be pretty unwieldy, especially when
`detail.target` already exists, which would suggest that `detail.elt`
should be something different. I feel as though it should be set to
something like `null` and therefore the burden should be on the front end
to check for this.

benvcutilli:
[...]

PR #3007, opened 2024-11-08
Improve detail.elt event documentation

MichaelWest22:
## Description
Event documentation for detail.elt is incorrect and confusing as the current documentation has this field listed as `the element that dispatched the request` when in fact a different element is being passed into the event. The detail.elt is always the element that the event is triggered on in htmx and this has to often be the element in question and not the one that started the request.

Corresponding issue:
#3000 

## Testing
No testing just careful reading of source to check how the events are actually triggered

## Checklist

  source changes)
  approved via an issue


alexpetros:
Hey thanks for taking the time to do this, it's a really huge improvement!

alexpetros (approved):

Issue #3016, opened 2024-11-12
Unable to disable inherit for hx-ext with hx-disinherit or htmx.config.disableInheritance


xhaggi:
I wonder why the inheritance of `hx-ext` is not done in the same way as the inheritance of all other attributes in htmx? If you disable the inheritance globally, this has no effect on `hx-ext`. Instead, you have to write the following to "ignore" the extension in a subtree, which is a bit annoying.

```html
<form action="/user.html" method="post" hx-ext="my-form-extension">
  <div hx-ext="ignore:my-form-extension">
    <a hx-get="/details.html">Details</a>
  </div>
</form>
```

IMO it would be better to do the following:

```html
<meta name="htmx-config" content='{"disableInheritance":"true"}'>

<form action="/user.html" method="post" hx-ext="my-form-extension">
    <a hx-get="/details.html">Details</a>
</form>
```

or

```html
<form action="/user.html" method="post" hx-ext="my-form-extension" hx-disinherit="hx-ext">
    <a hx-get="/details.html">Details</a>
</form>

MichaelWest22:
Yeah inheritance in htmx can be a bit confusing and could do with improvements in documentation.  There are three unique kinds of inheritance that I can find inside htmx.  

There is attribute inheritance which applies to a subset of the common htmx attributes where inheritance would be useful like hx-target, hx-swap, hx-select etc.  Attribute inheritance is the feature that is currently turned off with disableInheritance config.

There is also a special merge-key-inheritance used for attributes hx-vars, hx-vals, hx-headers and hx-request.  These attributes do not work like the normal attributes as they apply a dynamic list of keys/values to all their children elements where it allows you to set multiple keys in multiple parent elements and the values are all merged together with most specific key overriding.  This type does not respect the attribute inheritance disable config setting or the hx-inherit/hx-disinherit attributes as this does not fit the merge-keys system well and instead you have to use the not well documented hx-vals="unset" option on a child element to disable inheritance. see #2847 

And finally we have hx-ext inheritance which is currently implemented with its own style with "ignore:XXX" as the only way to remove inheritance currently.  I don't think we can reuse the attribute inheritance based disableInheritance config as this would be a breaking change that would impact any deployed sites that use this config to disable attribute inheritance but also want to use extensions.  Many of the extensions work best by placing the hx-ext on the body element as they make very global changes to how htmx works and moving these to apply only to the one element would be a big change in expected behavior.  

[...]

PR #3020, opened 2024-11-14
Allow removing extension inheritance with this-only: option

MichaelWest22:
## Description
Allow hx-ext extension inheritance to be disabled on an element with a new `this-only:` prefix instead of having to use `ignore:` on all children.  

It works by setting isParentScan to track when scanning parent nodes recursively and then ignoring recursive calls that have ignore: prefix.  Otherwise it strips the ignore: prefix.

Also improved readability/performance/minification by:
  removing un-needed explicit undefined checks
  moving replace => trim
  moving slice => indexOf

Corresponding issue: #3016 

## Testing
Added test to confirm extension with this-only: prefix does not get inherited as expected.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4928,9 +4928,9 @@ var htmx = (function() {
           extensionsToIgnore.push(extensionName.slice(7))
           return
         }
-        if (extensionName.indexOf('local:') === 0) {
+        if (extensionName.indexOf('this-only:') === 0) {
           if (isParentScan) return
-          extensionName = extensionName.slice(6)
+          extensionName = extensionName.slice(10)
         }
         if (extensionsToIgnore.indexOf(extensionName) < 0) {
           const extension = extensions[extensionName]
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4910,22 +4910,27 @@ var htmx = (function() {
    * @returns {HtmxExtension[]}
    */
   function getExtensions(elt, extensionsToReturn, extensionsToIgnore) {
-    if (extensionsToReturn == undefined) {
+    const isParentScan = !!extensionsToReturn
+    if (!extensionsToReturn) {
       extensionsToReturn = []
     }
-    if (elt == undefined) {
+    if (!elt) {
       return extensionsToReturn
     }
-    if (extensionsToIgnore == undefined) {
+    if (!extensionsToIgnore) {
       extensionsToIgnore = []
     }
     const extensionsForElement = getAttributeValue(elt, 'hx-ext')
     if (extensionsForElement) {
       forEach(extensionsForElement.split(','), function(extensionName) {
-        extensionName = extensionName.replace(/ /g, '')
-        if (extensionName.slice(0, 7) == 'ignore:') {
+        extensionName = extensionName.trim()
+        if (extensionName.indexOf('ignore:') === 0) {
           extensionsToIgnore.push(extensionName.slice(7))
           return
+        }
+        if (extensionName.indexOf('local:') === 0) {
+          if (isParentScan) return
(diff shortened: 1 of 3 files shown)


Telroshan (approved):

Telroshan:
Not a change request at all, just throwing out alternative naming ideas for the prefix itself in case you think they sound good ; `self:`, `isolate:`, or `isolated:` came to mind while answering on the issue

MichaelWest22:
Yeah `local:` is not the perfect word for this but I've struggled to find a better word that fits as well.  local brings to mind things like local vs global scope which is well understood but in a way the current extension default behavior is already locally scoped to the form element!  it just also applies to the form elements children because they are in the same scope normally.  But in this use case we want to just break inheritance to the children which makes `self:` make more sense.  `isolate:` and `isolated:` have the issue that is not obvious what you are isolating without reading the documentation as i think these are less self explaining than 'local:'.

What i like about `local:` is that how it reads in your head when looking at hx-ext="local:my-exetnsion" as it makes you think "my-extension is a local extension that only applies locally".  But something like `self:` while possibly more accurate you have to think hx-ext="self:my-extension" means "my-extension is an extension that only applies to itself where itself refers to the local element it is tagged on".  

There are other more complex options that are even more accurate and descriptive like `disinherit:` (mirror hx-disinherit attribute) or `noninheritable:' or `nonheritable:` .  But the disinherit one feels more like the `ignore:` case and not for setting the extension up situation.  The other two *heritable ones seem to be too long and complicated to use to me.

MichaelWest22:
@Telroshan I just tried updating this change to use `this-only:` instead of `local:` to see if this improves the readability and understandability of this feature proposal.  asked for some advice on discord and had suggestions for using `self-only` or `this` as good options.  I like `this` because it has existing meaning and use in htmx like in hx-target to refer to the element itself.  `self` is also good but its not as obvious if the subject of the self is the element or the extension so I think `this` is probably more accurate.  Adding the `-only` makes it much more obvious and human readable I think as it now reads like `apply this extension to this element only`

Telroshan:
Sounds like a good name to me!

src/htmx.js
```diff
@@ -4910,23 +4910,28 @@ var htmx = (function() {
...
       forEach(extensionsForElement.split(','), function(extensionName) {
-        extensionName = extensionName.replace(/ /g, '')
-        if (extensionName.slice(0, 7) == 'ignore:') {
+        extensionName = extensionName.trim()
+        if (extensionName.indexOf('ignore:') === 0) {
           extensionsToIgnore.push(extensionName.slice(7))
           return
         }
+        if (extensionName.indexOf('this-only:') === 0) {
+          if (isParentScan) return
```
1cg, line 4932: small thing but I don't like if statements w/o a {} (I'm hypocritical about this, but hey it's a code review)

1cg (changes requested):
one small change thank you!

xhaggi:
As I said before, I'm not a big fan of these prefixes. Introducing attributes such as `hx-ext-ignore` and `hx-ext-disinherit` are the better choice in my opinion. We still have a choice now, once this change is in, it will be hard to change it.

```html
<form action="/action" hx-ext="ext1,ext2" hx-ext-disinherit="ext1"> <!-- disable inheritance for ext1 -->
  <a href="#" hx-delete="/delete"></a>
</form>
```

```html
<form action="/action" hx-ext="ext1,ext2">
  <a href="#" hx-get="/get" hx-ext-ignore="ext1,ext2"></a> <!-- ignore ext1 only on this element -->
  <a href="#" hx-delete="/delete"></a>
</form>
```

1cg:
closing as too niche

PR #3047, opened 2024-11-29
Docs anchors

yawaramin:
Fix the 'Configuring Response Handling Examples' header link and move anchor links to the right of headers to make sure they don't get cut off on certain display sizes. Also use link emoji as anchor link text rather than hashtag.


yawaramin:
Screenshot:

Screenshot 2024-11-29 at 00 34 05

yawaramin:
I can go either way. I think though instead of trying to guess sizes, it's better to just keep the anchor link flush with the margin:

Screenshot 2024-11-29 at 19 28 50

yawaramin:
Btw...too cute? 🤔

Screenshot 2024-11-29 at 19 40 51

Telroshan:
>  I think though instead of trying to guess sizes, it's better to just keep the anchor link flush with the margin

Personal opinion again, but I see value in the anchor being offset to the left ; it stands out more, which helps identifying titles better, while keeping titles' text aligned with text body below them. Very subjective I know, but it looks more readable to me to have the texts aligned when possible.
(So to me, the best looking is anchors aligned + text aligned, then anchors aligned when you can't have both)

670px looks to me to be the very minimum acceptable size for such a display change, and even there, it doesn't look very ideal to me to target these anchors, that feel a bit too close to the window's edge.
700px is then just a rounded value suggestion, to add some margin and let those breathe a bit

As for why that size in particular;
- An existing media query (under 768px of screen width) applies a max width of 640 px to the page content
- This page content itselfs adds margin on its sides to be centered
- The anchor itself is roughly 14px wide (14.2, says my dev tools)
So, mathematically, when reaching `640 + 14*2` (668 px), you start seeing the anchor going offscreen.

Just to emphasize that it's not some random size guessing, but simply the media queries & the styling that's already being applied to the website by other CSS rules, thus sounds perfectly fine to me to add another media query

yawaramin:
OK how about this: the entire header is a clickable anchor link, and shows the link emoji on hover:

Screenshot 2024-12-01 at 15 52 59

Telroshan:
Sure why not, the link color makes it clear it's clickable and makes it stand out from the standard text, sounds good to me.
However, you'll want to fix the dark theme display as the titles are rendered in white on my side (thus don't look like links nor clickable, but like regular text instead)

yawaramin:
Fixed
Screenshot 2024-12-02 at 21 48 50

1cg:
can you change the link image to just a hash character?  i'm all about old-school but that image is a bit much (i though I did that at one point...)

yawaramin:
Sure. I'm not terribly attached to the emoji.

Issue #3068, opened 2024-12-11
bug: body hx-boost = true + form tag = Uncaught Type Error: path is null


wbcat:
using htmx 2.0.3 the following html document

```html
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="utf-8">
	<title>Title</title>
	<script src="htmx.js"></script>
</head>
<body hx-boost="true">
	<form>
	</form>
</body>
</html>
```

leads to 
`Uncaught TypeError: path is null`
in
`htmx.js:2326:13`

htmx 2.0.2 is fine

kkpo4553:
Hello, we tried to solve the issue.

This is what we did:

Added a null check for the path variable before accessing its properties to prevent the 'Uncaught TypeError: path is null' error.

You can review changes in this commit: https://github.com/kkpo4553/bigskysoftware-htmx-3068/commit/84532718238f96a9d1aac9a796225c6fe9019bd1.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

MichaelWest22:
This issue has already been fixed in the latest dev build here #2967 so fix will be in the next release

wbcat:
Thank you!

PR #2967, opened 2024-10-15
Fix TypeError on null path variable

youssame:
## Description
I check if the `path` is null or empty when the user hasn't set the action property or has left it empty.

Corresponding issue: #2963

## Testing
I added two tests that check if everything is OK in two cases (action is unset or is empty)

## Checklist

* [x] I have read the contribution guidelines
* [x] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [x] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [x] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #3069, opened 2024-12-11
Fix: body hx-boost = true + form tag = Uncaught Type Error: path is null

youssame:
## Description
I add a check if `path` is `null`, if it's true I set the path with an empty value. It's the only way to avoid future regressions.

Corresponding issue: #3068

## Testing
I tested this fix by running all the unit tests, and locally with same case shared by the person who created the issue.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ npm test
 
 ## Pull Requests
 ### Technical Requirements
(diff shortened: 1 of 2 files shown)


MichaelWest22:
looks like you already fixed this issue in the this PR #2967 which is now in dev branch already.

1cg:
i added a similar fix in another change, thank you!

PR #3073, opened 2024-12-12
Fix ajax api body test

MichaelWest22:
## Description
ajax api test for body was causing problems possibly as it was swapping and then reverting body.  Change this test so it just checks if the body is the target but uses a swap of none so it won't possibly cause issues to the body state.

Corresponding issue:

## Testing
Ran tests locally to make sure they still pass.

## Checklist

  source changes)
  approved via an issue


alexpetros:
Thanks @MichaelWest22 !

Issue #3062, opened 2024-12-08
Hx-trigger doesn't trigger on first reset DOM event, unless a delay is set


zachczx:
Hi, I'm using HTMX with hx-trigger="change, reset" with hx-post to post filter selections to my Go backend.

- Change for when checkboxes' value change
- Reset for when a reset button is clicked.

```
<!-- Sample --> 
<form id="filter-form" hx-target="#posts" hx-swap="outerHTML" hx-trigger="change, reset" hx-post="/filter">
	<input id="1" type="checkbox" name="number" value="1" />
	<label id="1-label" for="1">Label 1</label>
	<button id="reset-filter" type="reset">Clear</button>
</form>
```

Checkbox value change works as intended - triggers htmx swap. 

## Problem
Reset event triggers without hx-post of empty checkboxes value on first click of reset button. This first click comes with the typical htmx events, e.g. htmx:beforeSwap. It only works when reset button is clicked the second time. 

## Workaround
Behavior works as intended if I set a delay (any delay, 0.01s works). Seems like the reset event interferes with the change trigger. 

Is this intended behavior, i.e. if I want to listen to reset events, I need to add a delay? 

Thanks!

Telroshan:
Hey, yep I would say it works as intended ; event listeners bound to standard events like this, run before the browser's logic does. This is to let you the opportunity to call `event.preventDefault()` on them.
If you were to call `event.preventDefault()` in a `reset` event listener, you'd see the form not resetting at all. As htmx binds standard event listeners, it indeed runs and builds its values set before the reset actually occurs.

To fix the issue, you'd have to use some delay/timeout anyway to first let the browser reset the form, then trigger the htmx request, whether it's hidden in an internal check in htmx, or explicity defined in your `hx-trigger` attribute.
So, your workaround is the way to go imho! A `hx-trigger="change, reset delay:1"` will fix this

zachczx:
I see, thanks for the explanation! This would be a great thing to mention briefly (or link to your explanation in this PR), would definitely save some folks a bit of time figuring out why the reset event didn't work. Cheers!! :)

Telroshan:
If you're interested, feel free to open a PR to clarify that @zachczx ! Documentation improvements are always welcome 😁

zachczx:
> If you're interested, feel free to open a PR to clarify that @zachczx ! Documentation improvements are always welcome 😁 

Yup will do during the weekend!

PR #3076, opened 2024-12-14
Update hx-trigger.md

zachczx:
## Description

Creating this PR as a follow-up action to my original question: https://github.com/bigskysoftware/htmx/issues/3062

- Added a point of clarification in the notes to suggest adding a delay when adding reset to hx-trigger.

-  I also propose to clarify the part on standard events, by adding a reference to MDN's list of web API events. As someone new to HTMX in the past I found it weird that the docs didn't clarify what standard events were for hx-trigger and had to Google for a bit.

## Checklist

- [x] I have read the contribution guidelines
- [ ] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
- [ ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
- [ ] I ran the test suite locally (`npm run test`) and verified that it succeeded

### DIFF
--- a/www/content/attributes/hx-trigger.md
+++ b/www/content/attributes/hx-trigger.md
@@ -11,6 +11,8 @@ value can be one of the following:
 
 ### Standard Events
 
+Standard events refer to [web API events](https://developer.mozilla.org/en-US/docs/Web/API/Element#events) (e.g. click, keydown, mouseup, load).
+
 A standard event, such as `click` can be specified as the trigger like so:
 
 ```html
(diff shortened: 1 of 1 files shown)


www/content/attributes/hx-trigger.md
```diff
@@ -155,3 +157,4 @@ The AJAX request can be triggered via JavaScript [`htmx.trigger()`](@/api.md#tri
 * `hx-trigger` is not inherited
 * `hx-trigger` can be used without an AJAX request, in which case it will only fire the `htmx:trigger` event
 * In order to pass a CSS selector that contains whitespace (e.g. `form input`) to the `from`- or `target`-modifier, surround the selector in parentheses or curly brackets (e.g. `from:(form input)` or `from:closest (form input)`)
+* Reset triggers only on the second successive event (e.g. 2 clicks on reset button) when hx-trigger contains other triggers (e.g. hx-trigger="change, reset"). Add any delay to reset as a workaround (e.g. hx-trigger="change, reset delay:0.01s").
```
Telroshan, line 160: Unless I misunderstood, I don't think it is what is happening (or maybe is it simply a wording issue).
Take a look at [this JSFiddle](https://jsfiddle.net/pz8w31mL/):
- `reset` does trigger on the first click, you don't have to click 2 times to trigger it. The issue is that htmx builds its values set & fires its request _before_ the browser actually resets the form. So 1 click on the reset button already triggers htmx, and clicking twice would trigger 2 requests (one with the form not reset, then the form resets, then the second click sends a request with the form's reset values)
- This is due to the event listener's logic being called _before_ the browser's logic, as you may want to call `event.preventDefault` here to prevent the reset from happening at all
- Adding any delay as a workaround works, because then you let htmx fire its request & build its values set _after_ the browser has reset the form

zachczx: My bad, you're absolutely right. I've rewritten the proposed changes with your explanation in mind.

Telroshan (approved):

Telroshan:
Thanks!

PR #3109, opened 2025-01-06
Adding example to hx-ext documentation

zachczx:
Hi, 

I read the [hx-ext docs](https://github.com/bigskysoftware/htmx/blob/master/www/content/attributes/hx-ext.md) while doing stuff, and I thought it can be made clearer with an example for those who have multiple extensions.

### DIFF
--- a/www/content/attributes/hx-ext.md
+++ b/www/content/attributes/hx-ext.md
@@ -4,7 +4,7 @@ title = "hx-ext"
 
 The `hx-ext` attribute enables an htmx [extension](https://htmx.org/extensions) for an element and all its children.
 
-The value can be a single extension name or a comma separated list of extensions to apply.
+The value can be a single extension name or a list of extensions (separated by commas).
 
 The `hx-ext` tag may be placed on parent elements if you want a plugin to apply to an entire swath of the DOM,
 and on the `body` tag for it to apply to all htmx requests.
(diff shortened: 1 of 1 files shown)


www/content/attributes/hx-ext.md
```diff
@@ -4,7 +4,7 @@ title = "hx-ext"
 
 The `hx-ext` attribute enables an htmx [extension](https://htmx.org/extensions) for an element and all its children.
 
-The value can be a single extension name or a comma separated list of extensions to apply.
+The value can be a single extension name or a list of extension names (separated by commas).
```
Telroshan, line 7: Not sure this is a relevant wording change, as "comma-separated" is a relatively common term now, for ex in the renowned [CSV](https://en.wikipedia.org/wiki/Comma-separated_values) format (we may be missing a hyphen in between those 2 words here though 🧐 )

www/content/attributes/hx-ext.md
```diff
@@ -25,4 +25,8 @@ hierarchy and it will apply to all child elements.
   </div>
 </div>
 ```
-
+```html
+<body hx-ext="preload,morph">
+...
```
Telroshan, line 30: Maybe we should add a comment here similar to the one above ?
```suggestion
  "preload" and "morph" extensions are used in this part of the tree
```

www/content/attributes/hx-ext.md
zachczx: Alright, I can see where comma-separated might resonate, I stand corrected.

Telroshan (approved):
Thanks!

PR #3112, opened 2025-01-06
Render toast as aria-live region on page load

yawaramin:
This tells the screen reader to announce updates that happen inside the `span#toast` element. Without this, the screen reader does not announce the notification message.


alexpetros (approved):
Thanks Yawar!

PR #3113, opened 2025-01-08
Fix relative URLs to be considered as sameHost

FeliciousX:
## Description

when the URL function is not used for the URL validation and the path is a relative URL
`startsWith(path, document.location.origin)` would return `false`.

Since we can guarantee that relative URLs must be on the same host, I wanted to move the check earlier up the code and return early but unsure if it is a desired pattern.

Corresponding issue:

## Testing

We had a site that was working normally but somehow a GoogleTagManager JS that we have no control over was removing the native `URL` function from the browser (which is wild)

Was implementing a workaround on our end but I thought this might break on IE11 right now if URL is really not available.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4007,7 +4007,7 @@ var htmx = (function() {
     } else {
     // IE11 doesn't support URL
       url = path
-      sameHost = startsWith(path, document.location.origin)
+      sameHost = startsWith(path, "/") || startsWith(path, document.location.origin)
     }
 
     if (htmx.config.selfRequestsOnly) {


MichaelWest22:
Hi FeliciousX,

yeah its crazy that they would break a core JS feature like URL().  This is likely to be just a bug in GTM or a mis configuration somehow. They may be trying to replace the URL() function with their own wrapper so they can track what URL's are being processed inside the page so they can be logged and reported on.  It is likely that google tag manager script you have added to your page is actually not working properly anymore after some htmx action.  HTMX with features like hx-boost and many of the ways it can replace the body during full page replacement and history navigation can cause problems with some browser features and JS.  You can check the GTM script is stable in your head tag as its possible this script is replacing the URL and then getting removed from Head somehow.  The htmx head-support extension might help your issue and I would also try setting htmx.config.historyCacheSize=0 and htmx.config.refreshOnHistoryMiss=true so it performs full page reloads each back navigation so it will reload the URL() and GTM scripts each time. I would also recommend loading up browser dev tools and adding a watch in the sources tab for 'URL' and then keep an eye on this as you perform various page navigation actions till you track down the action that breaks the function.

The issue with this fix is that htmx 2.x has fully dropped support for IE 11 so this URL fallback is no longer likely to be maintained and may be removed soon.  It could make sense to fix this in the htmx v1 branch instead of dev maybe where IE11 is still supported? would be great to have a test reproduce the issue.  It may be that because v1 does not block remote requests by default like v2 the IE bug you found was not picked up?

FeliciousX:
feel free to close this PR if it doesnt make sense here anymore (because of IE11 deprecation)

That being said, I could rebase to point this fix into v1, just let me know which branch I should pick as the base branch.

MichaelWest22:
2.0.5 just shipped and removed the IE11 fallback now.  Closing this as I don't think we need to support all edge cases like this in v1 for now

Issue #601, opened 2021-09-23
Prompt default value


r3sist:
Dear 1cg, dear developers,

As a feature request, is it possible in the future to add to hx-prompt an optional default value? It would be useful :)

+ https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt 

Thank you

vaughany:
Hey everyone. I realise it's getting on for three years since this suggestion was made, but I believe this suggestion has legs: it would be incredibly useful to be able to pre-populate a prompt when it is shown, rather than just having an empty text field.  

I know there's other options, such as SweetAlert2, or putting a form inside a Bootstrap modal, but I am trying to keep complexity to a minimum in a project I am working on.

maekoos:
I wanted this too, so I added an attribute `hx-prompt-default` (literally a two line change lol). Hope it gets approved or maybe just helped someone!

1cg:
fixed

PR #3117, opened 2025-01-10
Add `hx-prompt-default` to set the default value of a `hx-prompt` (#601)

maekoos:
## Description
Added `hx-prompt-default` attribute to (optionally) set the default value of a `hx-prompt`.
Also updated the documentation to include this new attribute.

Corresponding issue: #601 

## Testing
I added a manual test to `confirm-and-prompt.html`.

## Checklist

  source changes)
  approved via an issue **- not explicitly approved**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Using `<button hx-verb="/endpoint" type="reset">` will now reset the associated form (after submitting to `/endpoint`)
 * Using `<button formmethod="dialog">` will no longer submit its associated form
+* Added `hx-prompt-default` to set the default value of a `hx-prompt`
 
 ## [2.0.4] - 2024-12-13
 
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4325,7 +4325,8 @@ var htmx = (function() {
     }
     const promptQuestion = getClosestAttributeValue(elt, 'hx-prompt')
     if (promptQuestion) {
-      var promptResponse = prompt(promptQuestion)
+      const defaultValue = getClosestAttributeValue(elt, 'hx-prompt-default') || undefined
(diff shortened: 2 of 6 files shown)


MichaelWest22:
May not be best to add too many additional hx- attributes to htmx where possible.  Here we could keep the feature much more contained by just adding an appropriate separator that is unlikely for an existing user to have added to a existing hx-prompt question like | and make it take `hx-prompt="do you like cats?|yes"` . then it is just a .split() and pass in the two array values to prompt I think.  Not sure if there is a better separator character to choose.

maekoos:
Sure, but that would also require more sanitisation on the server side - most template engines already sanitize attributes, so it's nice to not have to think about it....

Would it be any better to rename it to something else? `hx-prompt::default`?

MichaelWest22:
templating engines may sanitize attribute names a bit but we are just talking about a attribute string value which should only have basic sanitizing to prevent escaping out of the "".  This is why it is important to pick a good sensible text character to split on.  needs to be a character that won't upset engines and is unlikely to be used by existing hx-prompt user questions.  most htmx attributes use colon ":" as the seperator but it is likely someone could have used this inside a prompt already.  It is also common in htmx to use a word as the seperator and this could work as well.  "do you like cats?default:yes" for example with split("default:") in the code as an example.

maekoos:
Most template engines don't sanitize attribute names at all - what I meant was that I can't do
```
hx-prompt="How old is {{ userName }}?|{{ defaultAge }}"
```
Since I'd have to look for `|` in the userName variable in case someone intentionally tries to break stuff...

Also, unless a really weird character is chosen we'd also have to allow escaping (like `\|`), and so a regex or similar is probably necessary.

maekoos:
Plus, reusing the hx-prompt attribute is not backwards compatible enough imo.

Issue #2475, opened 2024-04-09
Extensions documentation could make need for hx-ext clearer


ChrisJefferson:
Via google I landed on the manual page for 'response-targets', which did exactly what I want.

For a while I couldn't get it to work, because I'd never used an extension before, and didn't carefully every bit, so I didn't understand how important `hx-ext` was.

How about, under "install" in each extension, after the "Install" line, adding a line that says something like:

All extensions must be \[enabled\](@/extensions/_index.md) using \[hx-ext\](@/attributes/hx-ext.md) before that can be used.

Just to point people towards this important thing they must read (I'm linking to both extensions, and ht-ext, as those are I think the two most useful pages?)

I'm happy to make a PR, just wanted to check what people thought before I bothered going through all the extension pages.

Telroshan:
Forgot to answer back then @ChrisJefferson , but sure feel free to open a documentation improvement PR!

marisst:
@ChrisJefferson and @glossd, I have created a PR significantly elaborating the instructions of setting up HTMX extensions. Based on your inputs, I have made the importance of enabling the extensions with the `hx-ext` tag more visible and repeated it multiple times.

The PR: https://github.com/bigskysoftware/htmx/pull/3078

<img width="1039" alt="Skjermbilde 2024-12-19 kl  22 02 53" src="https://github.com/user-attachments/assets/d67b2da1-a58d-4954-b09e-de441fda749e" />

<img width="1043" alt="image" src="https://github.com/user-attachments/assets/b7cf515d-e99d-45e9-a591-3f341525b304" />

Similar communication is used in the dedicate pages of the core extensions.

I hope you like it and we can close this issue after the PR is merged.

PR #3127, opened 2025-01-12
Extension docs: npm, bundler, min/unmin and SRI hash instructions

marisst:
Pull request should be viewed together with the corresponding PR in htmx-extensions repository https://github.com/bigskysoftware/htmx-extensions/pull/123.

Documentation is updated to elaborate on extension installation and integration. Until now, extension installation instructions were limited to a CDN script tag. This change adds:

- guidance on installation with npm
- guidance on integration with bundlers
- instructions to access minified and unminified extension versions
- SRI hash in CDN script tag installation example
- links to newest extension versions which were released with the `htmx.org@2.0.4` release
- makes the need of `hx-ext` clearer #2475


Telroshan (approved):
Thanks again for your work, and your patience!

Issue #3106, opened 2025-01-05
Bug? hx-on:* not fired when using hx-trigger="load"


michal-laskowski:
This will not work
```
  it('before-request - without delay', function() {
    var requests = 0
    window.foo = 0
    this.server.respondWith('GET', '/test', function(xhr) {
      requests++
      xhr.respond(200, {}, 'Requests: ' + requests)
    })
    var div = make('<div hx-trigger="load, click" hx-get="/test" hx-on:htmx:before-request="window.foo += 1">Requests: 0</div>')
    div.innerHTML.should.equal('Requests: 0')
    this.server.respond()
    div.innerHTML.should.equal('Requests: 1')
    div.click()
    this.server.respond()
    div.innerHTML.should.equal('Requests: 2')
    delete window.foo
    window.foo.should.equal(2)
  })
```

When usind delay:1 it will work. 
```
  it('before-request - with delay', function(done) {
    var requests = 0
    window.foo = 0
    this.server.respondWith('GET', '/test', function(xhr) {
      requests++
      xhr.respond(200, {}, 'Requests: ' + requests)
    })
    var div = make('<div hx-trigger="load delay:1ms" hx-get="/test" hx-on:htmx:before-request="window.foo += 1">Requests: 0</div>')
    div.innerHTML.should.equal('Requests: 0')
    this.server.autoRespond = true
    this.server.autoRespondAfter = 5

    setTimeout(function() {
      window.foo.should.equal(1)
      delete window.foo
      div.innerHTML.should.equal('Requests: 1')
      done()
    }, 50)
  })
````

**Comment by rkilpadi** [comment]
I found similar issues with `reveal` and `intersect`. Looks like this bug happens when triggers fire before the node is fully processed (see `addTriggerHandler()`). `hx-on` attributes are processed after the rest of the node (see `processNode()`) and therefore don't have a chance to attach before these triggers fire.
I also found issues with inheritance for all these triggers when invoking `hx-on` attributes.

The following works as intended and increments `foo` __only__ when the div is not immediately revealed:
```html
<div hx-trigger="revealed" hx-get="/test" hx-on::before-request="foo++"></div>
```

However, the following doesn't increment `foo` in either case:
```html
<div hx-trigger="revealed">
    <div hx-get="/test" hx-on::before-request="foo++"></div>
</div>
```
Replacing `revealed` with `load delay:1ms` has similar results.

[...]

PR #3131, opened 2025-01-14
Attach hx-on handlers before processing node

rkilpadi:
## Description
Currently, `hx-on` attributes are processed after triggers, causing issues where some triggers (i.e. `load` and sometimes `revealed`) are fired before `hx-on` handlers attach.  
This change attaches `hx-on` attributes before triggers are processed.
[Discussion](https://discord.com/channels/725789699527933952/1130605350941368350/1328540569743392810)

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/3106

## Testing
- Added tests to ensure `hx-on` attributes are triggered `revealed` and `load`
- Updated the polling test to `every 20ms` from `10ms`. From what I can tell, this is because of quirks in the testing framework (ideally should be an async test?) rather than performance issues caused by my changes. Some evidence:
    - Without my changes, setting the poll frequency to `9ms` sometimes failed
    - Running only that test against my changes  using `it.only` worked even when I set the frequency to `1ms`
    - As an extra sanity check, I used the following example to test manually. Assuming perfect performance, the log should print 10000. However, both with and without my changes, I got something in the 2000-2020 range.
```html
<body>
  <div hx-get="/test" hx-trigger="every 1ms" hx-on::before-request="foo++"></div>
  <script>
    foo = 0
    setTimeout(() => console.log(foo), 10000)
  </script>
</body>
```

## Checklist

  source changes)
  approved via an issue


src/htmx.js
```diff
@@ -2955,9 +2943,29 @@ var htmx = (function() {
...
+        nodeData.initHash = hash
+        elementsToInit.push(e)
+      }
+    }
+
+    maybeDeInit(elt)
+    forEach(findElementsToProcess(elt), maybeDeInit)
+
     forEach(findHxOnWildcardElements(elt), processHxOnWildcard)
+    initNode(elt)
```
scrhartley, line 2967: This is being done unconditionally, where previously it was unconditional.
Plus `elt` would already be in `elementsToInit` if needed, so this line should be removed entirely.

rkilpadi: Good catch, fixed!

Telroshan (approved):

gumbarros:
LGTM

1cg:
Is there any inadvertent semantic changes with this?

src/htmx.js
```diff
@@ -2955,9 +2943,28 @@ var htmx = (function() {
...
+      }
+      const nodeData = getInternalData(e)
+      const hash = attributeHash(e)
+      if (nodeData.initHash !== hash) {
+        deInitNode(e)
+        nodeData.initHash = hash
+        elementsToInit.push(e)
+      }
+    }
+
```
1cg, line 2962: I would prefer functions to not be defined in other functions

1cg (changes requested):
Can this cause any inadvertent semantic changes?  And see my one comment on the inner function.

test/attributes/hx-on-wildcard.js
```diff
@@ -133,6 +133,20 @@ describe('hx-on:* attribute', function() {
...
+  it('should fire when triggered by load', function() {
+    this.server.respondWith('POST', '/test', 'test')
+    make("<div hx-trigger='load' hx-post='/test' hx-on:htmx:config-request='foo = true'></div>")
+    window.foo.should.equal(true)
+    delete window.foo
+  })
+
+  it('should fire when triggered by revealed', function() {
+    this.server.respondWith('POST', '/test', 'test')
+    make("<div hx-trigger='revealed' hx-post='/test' hx-on:htmx:config-request='foo = true'></div>")
```
MichaelWest22, line 145: the div would also need 
```
style='position: fixed; top: 1px; left: 1px; border: 3px solid red'
```
so that it is forced to be visible on the screen or this test is flaky for me during testing when test browser is not scrolled into view at the time.  Other revealed tests have this style set to solve this

rkilpadi:
Updated to fix merge conflicts, remove inner function, and remove the redundant check on the main element. Thanks @MichaelWest22. The refactor to `eltIsDisabled` is a lot cleaner but I figured that might be for another PR.

MichaelWest22:
Yeah That looks good to me.  Not sure if they would want this eltIsDisabled fixed or not as its just a code readability and minification improvement only.

rkilpadi:
> Not sure if they would want this eltIsDisabled fixed or not as its just a code readability and minification improvement only.

I was a little hesitant just because at first glance it looks like the change cleans up an element that wasn't affected before 
https://github.com/bigskysoftware/htmx/compare/dev...MichaelWest22:htmx:hxon#diff-f14ff7f8c91a6f76e7688c45ca01d600e8928769b268caa71cf09b8f12661f3dL2828

1cg:
@MichaelWest22 this is a little scary to me wrt unintended timing consequences for existing users.  I'll stare at the code a bit more.

MichaelWest22:
@1cg yeah just re-read the changes again and the only change is that nodes get hashed first then hx-on listeners get added and then the final init happens posting before and after events.  I was originally worried there could be issues with this change but after testing it I can't find any change that would alter expected behavior for users apps other than those using hx-on to listen to before/after process node events.  And it just goes from not supporting this use case to making it work now so seems fine to me.

PR #3139, opened 2025-01-20
Fix 1103 by evaluating hx-vals with this referring to the element

### LINKED ISSUE #1103: hx-vals errs on `event.target`
Here's an example of the issue https://codepen.io/reedspool-gh/pen/XWYbRXo 

For some reason, if clicking multiple times quickly, `hx-vals="js:{  test: event.target.value }">` causes an error.

If you click the button slowly, the count will appear and then increase. But if you click it twice quickly, an error appears and the HTMX no longer triggers upon clicking this button ever again.

I think this is two problems:

* If there is an htmx error, I think that should never stop the htmx event from triggering on the same element again
* The specific error that appears must be a bug, right?

I narrowed the bug down to the second click having an `event` value of `undefined` instead of an actual event as expected.

NOTE: I found this bug on a personal app in which the error actually occurs on the first click, every click. I don't know why it was different when I went to codepen, but the error was exactly the same.

### LINKED ISSUE #1658: `event` is undefined on `hx-vals` with `delay` modifier on `hx-trigger`
Adding a `delay` of seemingly any duration, causes the `event` in `hx-vals` to be undefined. Take the following example:

```html
<input
  type="text"
  hx-get="/fancy-endpoint"
  hx-trigger="keyup delay:1s"
  hx-vals="js:{ theBug: console.log(event) }">
```
`event` is defined if I remove the `delay:1s`.

maekoos:
## Description
Fixes #1103 by evaluating `hx-vals` attributes with `this` referring to the current element.

Example:
```html
<div hx-post="/vars" hx-vals="javascript:{ ...this.dataset, }" data-name="a name" data-hello="world"></div>
```
evaluates to `{ "name": "a name", "hello": "world" }` since `this` refers to the div with hx-vals attribute.

*Corresponding issue: #1103*

## Testing
Added a test in test/attributes/hx-vals.js. This tests that `this` acually refers to the current element.

## Checklist

  source changes)
  approved via an issue

Edit: May also fix #1658

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3855,7 +3855,7 @@ var htmx = (function() {
       }
       let varsValues
       if (evaluateValue) {
-        varsValues = maybeEval(elt, function() { return Function('return (' + str + ')')() }, {})
+        varsValues = maybeEval(elt, function() { return Function('return (' + str + ')').call(elt) }, {})
       } else {
         varsValues = parseJSON(str)
       }
--- a/test/attributes/hx-vals.js
+++ b/test/attributes/hx-vals.js
@@ -325,4 +325,16 @@ describe('hx-vals attribute', function() {
     this.server.respond()
     div.innerHTML.should.equal('Clicked!')
   })
+
+  it('js: this refers to the element with the hx-vals attribute', function() {
+    this.server.respondWith('POST', '/vars', function(xhr) {
+      var params = getParameters(xhr)
+      params.i1.should.equal('test')
+      xhr.respond(200, {}, 'Clicked!')
+    })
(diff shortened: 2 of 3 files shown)


Telroshan (approved):

scrhartley:
Merge conflicts need resolving following #3196.

I'm guessing it should be:
```js
varsValues = maybeEval(elt, function() {
  if (event) {
    return Function('event', 'return (' + str + ')').call(elt, event)
  } else { // allow window.event to be accessible
    return Function('return (' + str + ')').call(elt)
  }
}, {})
```

1cg:
@maekoos please resolve and I'll merge, it's a good fix

1cg:
closing due to https://github.com/bigskysoftware/htmx/pull/3332

PR #3147, opened 2025-01-25
Extension Proposal: `hx-drag` for `dragstart` & `drop` triggered htmx requests.

AjaniBilby:
Currently in htmx there are methods for enabling drag and drop behaviour for sorting elements using client side arrangement ([sorting example](https://htmx.org/examples/sortable/)). However this relies on client side execution, plus does not allow for more rich interactions like dragging an item from one group to another triggering a server side update.

This extension uses two top-level `hx-` attributes, `hx-drag` and `hx-drop` to allow different information to be added to the request if the element was the one dragged or the one dropped. These two values are parsed identically to [`hx-vals`](https://htmx.org/attributes/hx-vals/) (without support for `js:` dynamic calculation), and are included with the requests issued by the drag and dropped elements.

The requests created by the `hx-[drag|drop]` are only sent once both the `dragstart`, and `drop` events have been completed.
The `hx-drag` requests will then be sent based on the `hx-[drag|drop]-action` on each element if present via a `PUT` method by default (this can be overridden with `hx-[drag|drop]-method`). By default these two requests are sent in parallel, but they can be made parallel via the use of `hx-[drag|drop]-sync=true`, and by default the `drag` request will be sent first, however this can be reversed with `hx-[drag|drop]precedence="drag"`.

These two requests will include a merger of the `hx-drag` + `hx-drop` + `hx-vals` values. In the event of a conflict the current element's `hx-[drag|drop]` will override the secondary (so the `hx-drag` will override `hx-drop` values on the `hx-drag-action`).

`hx-drag` intentionally does not rely on the `hx-[get|post|put|delete|patch]` attributes, to allow distinguishing between regular `hx-trigger` events, and the more detailed `hx-drag` requests.

---

### Data-Transfer

This example simply has three different divs all of which have `hx-drag` and `hx-drop` bindings to show how the value collisions hare handled.

data-transfer example
[source](https://github.com/AjaniBilby/htmx-drag-examples/blob/main/app/routes/data-transfer.tsx)

### Organise Buckets

There is a list of immutable animals at the top of the page, and user addable buckets below which animals can be dragged into, and out of, and deleted by dragging into the trash. In this example there is no server state, and all state is managed by the DOM.

data-transfer example
[source](https://github.com/AjaniBilby/htmx-drag-examples/blob/main/app/routes/bucket.tsx)

### Organise List

This is the most extravagant example with server and client state, with sync detection. When you drag any item onto another, the dragged item will moved to before the dropped item. This example utilizes a `hx-include` and `hx-swap-oob` to use a hidden `<input name="hash">` value to detect when the client has gotten out of sync, in which case the server response with a `hx-location: ` header, forcing the client to reload the current page using `hx-boost` allowing it to resync with the server state.

data-transfer example
[source](https://github.com/AjaniBilby/htmx-drag-examples/blob/main/app/routes/list.tsx)

---

The current implementation of the extension used in the examples can be found [here](https://github.com/AjaniBilby/htmx-drag-examples/blob/main/public/script/hx-drag.js).

This implementation is not 100% complete, as it doesn't currently encapsulate the functions can global variables out of global scope, plus it does not respect `hx-ext` to only apply to approved regions of the dom. It also does not currently allow `data-` as a prefix on all of the attributes used.


AjaniBilby:
## Bonus Example

I am currently using this for an internal tool at work (which I cannot provide source for), but in this case the implementation is much simpler, because the `hx-drag` just has the data for the thing being dragged with no action bound, then the `hx-drop` has the shipment the repair is being moved to. Once the server has updated the database it just responds with a `hx-location: ` header to trigger a `hx-boost` reload since there are many different elements updating at once, and the page only takes 32ms to load anyway.

img

Telroshan (approved):
Very cool extension!

PR #3157, opened 2025-01-29
Add form-json extension

xehrad:
This pull request introduces the [`form-json`](https://github.com/xehrad/form-json/blob/main/README.md) extension, addressing [htmx-extensions issue 131](https://github.com/bigskysoftware/htmx-extensions/issues/131).  

### **Overview**  
The [`form-json`](https://github.com/xehrad/form-json/blob/main/README.md) extension enables the automatic serialization of form data into a structured JSON object, following the [HTML JSON form](https://www.w3.org/TR/html-json-forms/) specification. It integrates seamlessly with HTMX, enhancing form submission workflows.  

### **Features**  
- **Automatic JSON Serialization** – Converts form inputs into structured JSON.  
- **Nested Structures** – Supports dot (`.`) and bracket (`[]`) notation.  
- **Type Preservation** – Maintains correct typing for numbers, booleans, and file inputs.  
- **File Handling** – Encodes file inputs as Base64 for seamless transmission.  
- **HTMX Compatibility** – Works with `hx-ext`, and `hx-vals` attributes.  

### **Installation & Usage**  
To enable the extension, include the script and add `hx-ext="form-json"` to forms:  
```html
<script src="https://unpkg.com/htmx-ext-form-json@1.0.1"></script>

<form hx-ext="form-json" hx-post="/test">
  <input name="username" value="JohnDoe">
</form>
```

Looking forward to feedback! 🚀

## Checklist

* [*] I have read the contribution guidelines
* [*] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [*] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [*] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #3160, opened 2025-01-30
 Fix some external links (#3157)

xehrad:
I agree. I had tried to arrange the list in alphabetical order.  

`json-enc` converts an HTML form into a simple `{input.name: input.value}` map, whereas `form-json` allows `input.name` values containing `.` to be interpreted as paths for nested structures. 
 It also supports `<input type="file">`, converting file inputs to Base64-encoded data.

## json-enc
```html
<form hx-ext="json-enc">
  <input name="user.age" value="18" type=number" />
  <input type="file" name="document">
</form>

<!-- Submission:
{
  "user.age":  "18"
}
-->
```

## form-json
```html
<form hx-ext="form-json">
  <input name="user.age" value="18" type=number" />
  <input type="file" name="document">
</form>

<!-- Submission:
{
  "user":  {
    "age":  18
  },
  "document": {
    "type": "application/pdf",
    "name": "file.pdf",
    "body": "SSBtdXN0IG5vdCBmZWFyLlxuRmVhciBpcyB0aGUgbWluZC1raWxsZXIuCg=="
  }
-->
```

## Checklist

* [*] I have read the contribution guidelines
* [*] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [*] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [*] I ran the test suite locally (`npm run test`) and verified that it succeeded


Telroshan (approved):

MichaelWest22:
Just noticed this extension we now link to happens to have an incompatible GPLv3 license which could cause legal issues for any user linking to it inside their project. @xehrad is there a reason this library has to be GPL?  The vast majority of web development is not compatible with GPL licensed libraries.  There is the LGPL which is better but it still has some unknowns and compliance issues.  I would recommend re-licensing under a similar license to htmx like https://github.com/bigskysoftware/htmx/blob/master/LICENSE BSD-0 as this is much more friendly for web applications to consume.

xehrad:
> Just noticed this extension we now link to happens to have an incompatible GPLv3 license which could cause legal issues for any user linking to it inside their project. @xehrad is there a reason this library has to be GPL? The vast majority of web development is not compatible with GPL licensed libraries. There is the LGPL which is better but it still has some unknowns and compliance issues. I would recommend re-licensing under a similar license to htmx like https://github.com/bigskysoftware/htmx/blob/master/LICENSE BSD-0 as this is much more friendly for web applications to consume.


Thanks for catching this! You're right that GPLv3 could introduce compatibility issues for many users. I’ve gone ahead and changed the license to Zero-Clause BSD ([0BSD](https://github.com/xehrad/form-json#GPL-3.0-1-ov-file)), which aligns with htmx’s license and ensures maximum flexibility for web development.

Let me know if you have any other concerns!

MichaelWest22:
@xehrad great thanks!  I think your readme.md may need updating as well as you state the license in here as well

Issue #3165, opened 2025-02-02
hx-history not working on partial content


mrinc:
When using `hx-push-url` with partial content, pressing the back button does not actually work.  

From looking at the code, it seems that it's related to the context being on the body.  
In the case where we do not actually control the body, the history aspect fails completely.  
I have been digging and it seems there is a case where HTMX is loaded dynamically on page load where the DOMContentLoaded is never triggered.

If I instead push the events for the ready function into an array, and then manually call those functions, everything works perfectly.  
https://github.com/bigskysoftware/htmx/blob/10e8656af500d8dc3bba54459b8f29d53f326471/src/htmx.js#L5006

So, we need to find a way to reliably trigger that event... I'll keep looking and update this accordingly.

mrinc:
I have also noticed a condition that could be problematic.  

localStorage vs sessionStorage.  
Any reason why we don't use sessionStorage for history state? (didn't find any discussions about it from a quick gander)

MichaelWest22:
Yeah there is a good argument to be made that htmx should use session storage instead of local.

Are you linking to htmx in a head tag properly when you have this issue?  I've seen similar issues in #2865 where the htmx script tag is not loaded correctly in the head tag.  Htmx has to load and stay loaded all the time for history navigation to work as expected.  Pushing back if historyCacheSize is not zero should just swap the body contents but retain the head scripts like htmx in place. I only got it to work properly when htmx was loaded in the body by testing modifying it to store it's saved history url value into sessionStorage from memory.

mrinc:
No no, it's async.  
As in ...  
```sudocode
const elem = document.createElement('script');
document.head.appendChild(elem);
```
async (basically script load with a script over in head/body - ie ... we do not have control of the page)

It is loaded during page load basically.

The problem is (see my bugfix) is the ready function never gets called when doing this.
The navigation fails completely even with disabling history etc - doesn't matter.
If you try press back, it just does nothing - history fails completely.

When looking into the code, the history/forward/backward functions are setup during the DOMContentLoaded - but since that event never fires in this case, it never sets the above up - thus a failure mode.
So, I added window.onload as a additional check - so either DOMContentLoaded or window.onload (whichever is first) allows the ready function to work and setup htmx properly.

mrinc:
For the referenced issue - it is possible that it's having a similar effect since it is at the end of the body instead.  
That should work so possible related fix.

PR #2040, opened 2023-11-24
Revert readystate PR

alexpetros:
## Description
After discussion with @Telroshan, we're going to provide a new method that lets people initialize htmx at their own spot in the lifecycle if they load it in some nonstandard ways, rather than try to cover every possible gap in the loading cycle with closures. Discussion here: https://github.com/bigskysoftware/htmx/pull/2037#issuecomment-1825892415

Another PR for `initialize` to come. 

## Testing
Reverted to old tests.

## Checklist

* [x] I have read the contribution guidelines
* [x] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [x] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [x] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #3166, opened 2025-02-02
Fixing #3165

mrinc:
## Description
Fixes case where htmx is loaded async and DOMContentLoaded is never fired.   
By using window.onload as well, we cover the case where it is not fired.  
No change to current behavior but fixes the case of dynamically loading htmx.  

Corresponding issue: #3165

## Testing
Lets discuss how we should test this one and then can just update the test suite accordingly.  
It's specific case where the htmx lib is loaded async.  
htmx itself operates the same, the big issue occurs when you try go back in the browser (I could be missing something...).  
However this does bring up other notes where the ready function isn't run - so we should be testing for that currently.  
So it's not just, lets add a single test case, it's we should add a single test case for the ready function in the 2 cases, but we should also be running all the tests again, in an async loaded state to make sure htmx behaves the same in both load-state cases.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -5012,9 +5012,20 @@ var htmx = (function() {
   // Initialization
   //= ===================================================================
   var isReady = false
-  getDocument().addEventListener('DOMContentLoaded', function() {
+  var readyFunctions = []
+  /*
+   * When the HTMX lib is loaded async, the DOMContentLoaded event never fires, so the
+   * ready functions never trigger. (https://github.com/bigskysoftware/htmx/issues/3165)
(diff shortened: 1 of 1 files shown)


alexpetros:
Sorry, we've been [through the ringer](https://github.com/bigskysoftware/htmx/pull/2040) on this one and simply don't support loading htmx as a module this way. We make a good-faith effort but if what we have right now doesn't work you will need to do your own workaround.

I think we'd still be okay with exposing an `initialize` method if you want to load htmx async and then initialize it yourself.

Issue #14, opened 2020-05-16
Minor issue, scary console error on chromium


cscortes:
Get these errors, but googling the error they seems to be a non issue.  These guys says there is a  fix.  https://trackjs.com/blog/refused-unsafe-header/

Despite the RED LETTER ERRORS on the console, program never stops. 

```
kutty.js:1169 Refused to get unsafe header "X-KT-Trigger"
xhr.onload @ kutty.js:1169
load (async)
issueAjaxRequest @ kutty.js:1165
issueRequest @ kutty.js:609
eventListener @ kutty.js:614
kutty.js:1170 Refused to get unsafe header "X-KT-Push"
xhr.onload @ kutty.js:1170
load (async)
issueAjaxRequest @ kutty.js:1165
issueRequest @ kutty.js:609
eventListener @ kutty.js:614
```

chg20:
OK, should probably fix it.

chg20:
Want to take a crack?

cscortes:
okay, got this fixed. Will do a pull request in a bit.

chg20:
Was this fixed by https://github.com/bigskysoftware/htmx/commit/3eb4a56e285ec6e0586c10bdaa2a42661f0f3af1 ?

chg20:
Apparently yes

PR #3173, opened 2025-02-07
Write title as innerText instead of innerHTML

emilhem:
## Description
Changing the title update to use `innerText` instead of `innerHTML` calms some security checks such as GitHub's own code checking.

`tsc` is grumpy now though with the error `Property 'innerText' does not exist on type 'Element'.`

Not sure if i agree with `tsc` fully.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4686,6 +4686,7 @@ var htmx = (function() {
     if (title) {
       const titleElt = find('title')
       if (titleElt) {
+        /** @type HTMLElement */
         titleElt.innerText = title
       } else {
         window.document.title = title
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4686,8 +4686,7 @@ var htmx = (function() {
     if (title) {
       const titleElt = find('title')
       if (titleElt) {
-        /** @type HTMLElement */
-        titleElt.innerText = title
+        titleElt.textContent = title
       } else {
         window.document.title = title
       }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4686,7 +4686,7 @@ var htmx = (function() {
     if (title) {
       const titleElt = find('title')
       if (titleElt) {
-        titleElt.innerHTML = title
+        titleElt.innerText = title
       } else {
         window.document.title = title
       }


src/htmx.js
```diff
@@ -4686,7 +4686,7 @@ var htmx = (function() {
     if (title) {
       const titleElt = find('title')
       if (titleElt) {
-        titleElt.innerHTML = title
+        titleElt.innerText = title
```
geoffrey-eisenbarth, line 4689: The `find` function returns type `Element|null`, but `<title>` is  an `HTMLElement`, so probably just add `/** @type HTMLElement */` between lines 4688 and 4689.

emilhem: Thank you! Fixed! :)

geoffrey-eisenbarth:
Are you able to run the test suite? Let me know if you need any help!

geoffrey-eisenbarth (commented):
lgtm! 

fyi I can't approve the PR, just trying to help other contributors :)

Telroshan:
I must admit I'm a bit confused here ; `innerHTML` is a property of the `Element` interface (see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML)), so we shouldn't even need a type cast to `HTMLElement`, unless I'm missing something?

Moreover, the PR mentions replacing `innerHTML` by `innerText` but it looks like the current PR simply adds a type cast comment without changing the accessed property, is that intended?

Could you also share the ouput of the failing security checks you mentioned in the PR description?

geoffrey-eisenbarth:
@Telroshan The first commit changed `.innerHTML` to `.innerText`, but looks like when they added the type cast they reverted it back to `.innerHTML`.

You're right that `.innerHTML` is a property of `Element`, but as far as I can tell `.innerText` is not, which is why I suggested the type cast, since `<title>` will always be an `HTMLElement`.

Agreed it would be nice to have the output of the failing security test.

emilhem:
Here's the output from GitHub's CodeQL

Screenshot 2025-02-10 at 09-28-03 Client-side cross-site scripting · Code scanning alert #14 · volvo-cars_e2e-qa-blink

Unsure how I messed up the commit, but I'll commit now.

emilhem:
Well, I borked the commit history obviously... I'll see what I can do to fix it..

Telroshan:
Ok I see, thanks for the screenshot

Tbh, this kind of warning is kinda a lost cause regarding htmx as we support `eval` which is reported as dangerous by many scanners
Though, as the title tag [can only contain text](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title#technical_summary) as per the docs, I see no reason to oppose to use `ìnnerText` here

As for the property itself, I would maybe recommend `textContent` instead of `innerText`, as
- It's [a property of `Node`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent), which shouldn't require a cast at all since `Element` [inherits it](https://developer.mozilla.org/en-US/docs/Web/API/Element)
- It [doesn't trigger a reflow](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext) as opposed to `innerText` which must know what text is actually rendered

Telroshan:
As for the branch, you would indeed want to rebase to `dev` and drop all the unnecessary commits

emilhem:
Done! Thanks @Telroshan ! :)

lukewarlow:
One benefit to this change is it's one less usage of a Trusted Types sink that will need handling if HTMX ever wants to support running in Trusted Types enforced environments.

LukasPukenis:
Hello, I'm also experiencing this issue, and the PR looks good to me in terms of the fix. Is there any plan to merge it? Thanks!

Telroshan:
Hey @LukasPukenis , the current plan is likely to aim for quarterly releases (see [this 1cg's Discord message](https://discord.com/channels/725789699527933952/1046573806547910677/1323411980236619857)), as this PR is labelled "ready for review", it should be in the next update if these changes are approved!
If you need it asap, I would encourage you to fork & include this PR's change for example, then get back to the "official" htmx once the next version is out, so you're not stuck until then!

LukasPukenis:
@Telroshan thank you for the information provided <3

PR #3187, opened 2025-02-14
Dynamic csrf prevention

TracyGJG:
## Description
Additional content for the section on CSRF prevention using the `hx-headers` attribute.

Corresponding issue: n/a

## Testing

n/a but needs reviewing.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/attributes/hx-headers.md
+++ b/www/content/attributes/hx-headers.md
@@ -24,7 +24,10 @@ If you wish for `hx-headers` to *evaluate* the values given, you can prefix the 
   security considerations, especially when dealing with user input such as query strings or user-generated content,
   which could introduce a [Cross-Site Scripting (XSS)](https://owasp.org/www-community/attacks/xss/) vulnerability.
 
+* Whilst far from being a foolproof solution to [Cross-Site Request Forgery](https://owasp.org/www-community/attacks/csrf), the `hx-headers` attribute can support backend services to provide [CSRF prevention](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html). For more information see the [CSRF Prevention](https://htmx.org/docs/#csrf-prevention) section.
+
 ## Notes
 
 * `hx-headers` is inherited and can be placed on a parent element.
 * A child declaration of a header overrides a parent declaration.
+* The value of the attribute needs to be a valid JSON string.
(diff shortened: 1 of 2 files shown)


Telroshan:
Yeah to be honest I think that's a bit "too much detail" for the docs section.

However, I think the core idea here, i.e. to explain that you can use dynamic CSRF tokens, is totally valuable, maybe just too verbose in its current form

I think we could instead expand the previous CSRF example (`CSRF Prevention`), by adding bottom notes to it, to mention that:
- hx-headers supports a syntax to evaluate a value : link to the hx-headers notes, explain that this can be used to make a dynamic CSRF token
- maybe quickly mention the implied usage of eval (and link to the security section of the same docs page)
- also mention that it can also be done from javascript using a `htmx:configRequest` event listener (link to the associated docs again)

Whether your example (the `js:` attribute syntax) or a vanilla JS event listener is "better", I would leave this up to the reader, i.e. either showcase the 2 ways (but that'd be too much imho for this page), either none of them and simply explain them while linking to the associated docs!

I think we would benefit from a more concise explanation here, but that links to all relevant related sections of the htmx docs

MichaelWest22:
I think for the majority of cases having a static CSRF token per user session is simpler to recommend by default.  With all modern browsers and using authentication and session cookies with the best practice security settings just having a CSRF token protection will prevent basically all currently known CSRF attacks and this means a session based CSRF has the same level of protection as a constantly rotated one right now.

I think the better way to do CSRF for htmx if you had critical forms that need additional layers of security just in case is to implement per form CSRF tokens where each important form returned by from the backend as a full page or htmx partial includes a simple hidden input with a CSRF token scoped to the single form name. This allows you to set an appropriate expiry time and it prevents all attempts to create a copy of the form maliciously as the only way to get a form A token is to ask the server for the form properly.

Also I think as Telroshan mentioned using a `htmx:configRequest` JS event listener is probably a simpler and more secure way of doing dynamic security headers like this.  Using `js:` inline attribute javascript is actually quite insecure and if you have a website where the security requirements are high enough to demand constantly rotating CSRF tokens then allowing inline javascript or the 'unsafe-eval' javascript eval feature htmx uses for this `js:` feature may not be ideal.  You often need a very tight CSP header that such things to prevent XSS attacks which are riskier than CSRF issues often.  Moving all your custom JS snippets into event listeners like `htmx:configRequest` allows you to meat strict CSP requirements easily.

TracyGJG:
Withdrawn to consider valuable feedback provided.

PR #3208, opened 2025-02-23
Remove obsolete npm command + fix links to ws/sse extensions

Telroshan:
Follow-up to #3052 , when we moved the ws/sse test suite to the extensions repo, we forgot some obsolete links/mentions behind

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 ## introduction
 
 htmx allows you to access  [AJAX](https://htmx.org/docs#ajax), [CSS Transitions](https://htmx.org/docs#css_transitions),
-[WebSockets](https://htmx.org/docs#websockets) and [Server Sent Events](https://htmx.org/docs#sse)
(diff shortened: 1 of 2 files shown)


lkollar:
There is another mention of the  /test/ws-sse folder here in the docs: https://github.com/bigskysoftware/htmx/blob/master/www/content/extensions/ws.md#testing-with-the-demo-server.

Telroshan:
@lkollar thanks, though this one already links to the extensions repo where `/test/ws-sse` is, so I'd say it is correct as is!

> It is located in the /test/ws-sse folder of the [htmx-extensions](https://github.com/bigskysoftware/htmx-extensions) repository. Look at /test/ws-sse/README.md for instructions on running and using the test server.

lkollar:
Oh don't mind me then. I probably had an old checkout as I missed the link to the extensions repo.

Issue #2372, opened 2024-03-04
Problems with validation of elements in the form


benrojas01:
Please can you tell me the correct way to validate the elements of a form, since when sending the post it makes the exchange regardless of whether the fields are empty and with the required property

thanks

alexandre-dos-reis:
@benrojas01 Can you please post your code here ?

Telroshan:
Could [hx-validate](https://htmx.org/attributes/hx-validate/) solve your usecase here?

Iamrodos:
Having this issue too. 

An example form input and submit button might be.

```
<input type="email" class="form-control" name="email" autocomplete="off" 
  hx-validate="true" value="{{person.email}}" >
<button type="button" class="btn btn-primary"
   hx-post="/person/{{person.id}}"
   hx-target="#personCard"
   hx-indicator="#working-icon"
>Save</button>
```

The above fails to validate that the input is a valid email address, which the type enforces. If you do a normal submit button validation occurs and the user is informed of the failure and the submit is not done. 

```
<button type="submit" class="btn btn-primary" >Add Person</button>
```

Iamrodos:
> Could [hx-validate](https://htmx.org/attributes/hx-validate/) solve your usecase here?

My suspicion is this only works when the htmx action is on that input, not some other button on the form which captures and sends the inputs.

Iamrodos:
Found a resolution for this from https://www.reddit.com/r/htmx/comments/182y0z3/htmx_client_validation/. The htmx needs to go on the form not on the button. When done this way the validations are enforced. Here is a working example.

```
<form 
  hx-post="/person/{{person.id}}"
  hx-target="#personCard"
  hx-indicator="#working-icon"
 >
<input type="hidden" name="csrf" value="{{user['csrf']}}">
<input type="email" class="form-control" name="email" autocomplete="off"  value="{{person.email}}" >
<div class="text-end">
   <button type="post" class="btn btn-primary" >Save</button>
</div>
</form>
```

perepicornell:
Hi!

This solution apparently works only if the post is triggered by form's submit (which is the default behavior if you don't specify `hx-trigger`):

> Found a resolution for this from https://www.reddit.com/r/htmx/comments/182y0z3/htmx_client_validation/. The htmx needs to go on the form not on the button. When done this way the validations are enforced. Here is a working example.
> 
> ```
> <form 
>   hx-post="/person/{{person.id}}"
>   hx-target="#personCard"
[...]

PR #3215, opened 2025-03-02
fix: use reportValidity instead of checkValidity in validateElement 

xforman2:
## Description
Replaced `checkValidity` with `reportValidity` as described here https://github.com/bigskysoftware/htmx/issues/2372#issuecomment-2418842309

Corresponding issue:
https://github.com/bigskysoftware/htmx/issues/2372

## Testing
Tested manually in the scratch demo

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3492,7 +3492,7 @@ var htmx = (function() {
     const element = /** @type {HTMLElement & ElementInternals} */ (elt)
     if (element.willValidate) {
       triggerEvent(element, 'htmx:validation:validate')
-      if (!element.checkValidity()) {
+      if (!element.reportValidity()) {
         errors.push({ elt: element, message: element.validationMessage, validity: element.validity })
         triggerEvent(element, 'htmx:validation:failed', { message: element.validationMessage, validity: element.validity })
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "htmx.org",
-  "version": "2.0.2",
+  "version": "2.0.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "htmx.org",
-      "version": "2.0.2",
+      "version": "2.0.4",
       "license": "0BSD",
       "devDependencies": {
         "@types/node": "20.0.0",


Telroshan (approved):

src/htmx.js
```diff
@@ -3492,7 +3492,7 @@ var htmx = (function() {
     const element = /** @type {HTMLElement & ElementInternals} */ (elt)
     if (element.willValidate) {
       triggerEvent(element, 'htmx:validation:validate')
-      if (!element.checkValidity()) {
+      if (!element.reportValidity()) {
```
yawaramin, line 3495: Maybe we should add a doc comment saying that `validateElement` will now report validation errors instead of just checking them?

1cg: Agree, this is a small but significant change.  Maybe we should gate it with a config?

MichaelWest22:
```JS
      triggerEvent(element, 'htmx:validation:validate')
      if (!element.checkValidity()) {
        if(!errors.length) {
          element.reportValidity()
        }
        errors.push({ elt: element, message: element.validationMessage, validity: element.validity })
        triggerEvent(element, 'htmx:validation:failed', { message: element.validationMessage, validity: element.validity })
      }
```
Another issue is that when i tested this change it does not behave like normal form validation as it always triggers on the last input that is missing or invalid instead of the first one like a normal form would.  This is because we reportValidity() of all the inputs in order and the last invalid input gets the warning message for the user.   The above code snipit solved this for me by only doing reportValidtity on the first error found which is a much nicer UX.  

But this does seem like a kind of breaking change to add user alerts to existing projects that were not using them.  The current setup reports these error alerts as events via `htmx.validation:failed` and so i would assume some users would have used this feature to update the UI on validation errors in their apps and adding another layer of reporting would be kind of a breaking change even though it does put back the default submit behavior.  

Also @Telroshan mentioned it could be canceled with the invalid event so there is a way for a user to disable the alert if needed.  I did get this to work with preventDefault() but this invalid event does not bubble which means you have to add hx-on or event listeners to EVERY input one by one.

```JS
      triggerEvent(element, 'htmx:validation:validate')
      if (!element.checkValidity()) {
        if(triggerEvent(element, 'htmx:validation:failed', { message: element.validationMessage, validity: element.validity }) && !errors.length) {
          element.reportValidity()
        }
        errors.push({ elt: element, message: element.validationMessage, validity: element.validity })
      }
```
Something like this could allow a hx-on `hx-on::validation:failed="event.preventDefault()"` or event listener on body could allow reverting to the old behavior by using the failed event to cancel the alert when needed which bubbles fine.

But It could be better to just add documentation to our validation docs to get users who want this validation to add
```JS
document.body.addEventListener('htmx:validation:failed', function(event) {
  event.target.reportValidity()
})
```
or ```hx-on::validation:failed="event.target.reportValidity()``` 
so the can opt into this feature.

dwasyl:
Any chance of this fix making it into a release anytime soon? I just spent a few hours trying to track down what was happening since the validate seemed to be working but there was no user displayed message and this deals with that nicely.

1cg:
obviated by https://github.com/bigskysoftware/htmx/pull/3372

PR #3219, opened 2025-03-05
Fix wording in `changed` modifier description in hx-trigger.md

### LINKED ISSUE #3218: Incorrect Description for changed Event Modifier in Documentation
https://htmx.org/attributes/hx-trigger/

In the "Standard Event Modifiers" section of the documentation, the description for the changed modifier appears to be incorrect. The current text states:  

> "changed - the event will only change if the value of the element has changed."  

This phrasing suggests that the event itself changes, which is confusing and likely not the intended meaning. Based on the context , the changed modifier should restrict the change event to fire only when the element's value has changed.

**Proposed Fix**
The description should be updated to:  

> "changed - the event will only fire if the value of the element has changed."

patricksuo:
## Description
Update the description of the `changed` event modifier from "the event will only change" to "the event will only fire" to clarify that the modifier controls when the `change` event triggers, improving accuracy and readability. Addresses a potential documentation error.

Fixes #3218

## Testing
N/A

## Checklist

* [x ] I have read the contribution guidelines
* [ x] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
  approved via an issue

### DIFF
--- a/www/content/attributes/hx-trigger.md
+++ b/www/content/attributes/hx-trigger.md
@@ -56,7 +56,7 @@ for a global symbol with the name `foo`
 Standard events can also have modifiers that change how they behave.  The modifiers are:
 
 * `once` - the event will only trigger once (e.g. the first click)
-* `changed` - the event will only change if the value of the element has changed. Please pay attention `change` is the name of the event and `changed` is the name of the modifier.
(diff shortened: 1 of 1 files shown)


Telroshan (approved):
Thanks!

PR #3262, opened 2025-04-03
fix customized confirmation example

catwell:
The target is not the element that triggered the request, it does not necessarily have `hx-confirm`.


Telroshan (approved):
Good catch!

PR #3274, opened 2025-04-16
Mention hx-include does not match disabled fields. Fixes #3053

### LINKED ISSUE #3053: [Docs] Clarify that hx-include does not include disabled inputs
## Problem Definition
Hi, I just spent 2 hours debugging why the values are not included, until I realized that the reason is there is a `disabled` tag in my element (I don't want it to be editable). I replaced it with `readonly` and it worked. 

## Spec compliance
This behavior is up to [spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#attribute_interactions).

## Suggested solution
I just want to **add this info to the docs under Notes** to save people's time when they have a similar use-case and lack of knowledge in web standards just like me.

## Similar issues
Similar issues were submitted in the past (#2158 - form data overriding `hx-include` , #2384 - `name` attribute required) but this is a **separate problem**.

I am willing to create a PR for all of them, as it is a simple docs change and is appropriate for my first commit.

kontur:
## Description
An addition to the docs mentioning that `hx-include` does not include disabled inputs.

Corresponding issue:

https://github.com/bigskysoftware/htmx/issues/3053

## Checklist

* [+] I have read the contribution guidelines
* [+] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [+] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [+] I ran the test suite locally (`npm run test`) and verified that it succeeded

### DIFF
--- a/www/content/attributes/hx-include.md
+++ b/www/content/attributes/hx-include.md
@@ -52,3 +52,4 @@ Note that if you include a non-input element, all input elements enclosed in tha
   to [document.querySelectorAll](https://developer.mozilla.org/docs/Web/API/Document/querySelectorAll) and will include
   multiple elements, while the extended selectors such as `find` or `next` only return a single element at most to
   include
+* `hx-include` will ignore disabled inputs


Telroshan (approved):

Telroshan:
Thanks!

PR #3277, opened 2025-04-17
Remove old IE support

MichaelWest22:
## Description
I've looked at all cases of IE fallbacks that can be removed now that IE 11 is no longer supported by htmx 2.x.
I've included all possible old browser compatibility checks I could find but it may be possible to roll back some of these changes

As well as just safely removing dead un-needed fallbacks I also found three bugs that the removals have also forced me to address

1st is NormalizePath() which does not function at all as calling new URL(path) without an origin domain added as a second parameter is designed to throw errors every time causing the IE11 fallback to happen for ALL calls to normalizePath and this fallback does zero normalization.  I don't think this lost functionality would have caused big issues as this function is only used to set and restore paths to history so even if it is broken it was consistently broken so broken paths going into history cache came out with the same non normalized path most of the time.  After this change it normalizes properly again

2nd I found shadow dom tests were calling process/init node on the shadow dom root node itself which are not possible to init properly because this node can't have attributes.  The IE fallback was preventing any errors but it was leaving rubbish htmx internal data on the shadow root nodes then.  only processing elements resolved the shadow dom tests failing.

3rd there is a minor issue where after removing the IE fallback for if verifyPath() I found situations where new URL() will fail if things like srcdoc iframes have invalid origins.  Found using window.origin as a fallback when these null origins from `about:` protocol situations resolved the issue.

Corresponding issue:

## Testing
Performed extensive testing with the test suite and used my work on the loc coverage reporting to track down and test the dead branches.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2879,7 +2879,7 @@ var htmx = (function() {
       return
     }
     // Ensure only valid Elements and not shadow DOM roots are inited
-    if (elt instanceof Element) return
+    if (!(elt instanceof Element)) return
     const nodeData = getInternalData(elt)
     const attrHash = attributeHash(elt)
     if (nodeData.initHash !== attrHash) {
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -828,7 +828,7 @@ var htmx = (function() {
       path = url.pathname + url.search
     }
     // remove trailing slash, unless index page
-    if (!(/^\/$/.test(path))) {
+    if (path != '/') {
       path = path.replace(/\/+$/, '')
     }
     return path
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2879,41 +2879,40 @@ var htmx = (function() {
       return
     }
     // Ensure only valid Elements and not shadow DOM roots are inited
-    if (elt instanceof Element) {
-      const nodeData = getInternalData(elt)
-      const attrHash = attributeHash(elt)
-      if (nodeData.initHash !== attrHash) {
-        // clean up any previously processed info
-        deInitNode(elt)
+    if (elt instanceof Element) return
+    const nodeData = getInternalData(elt)
+    const attrHash = attributeHash(elt)
+    if (nodeData.initHash !== attrHash) {
+      // clean up any previously processed info
+      deInitNode(elt)
 
-        nodeData.initHash = attrHash
+      nodeData.initHash = attrHash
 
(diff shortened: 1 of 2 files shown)


src/htmx.js
```diff
@@ -484,10 +484,7 @@ var htmx = (function() {
    * @returns {boolean}
    */
   function matches(elt, selector) {
-    // @ts-ignore: non-standard properties for browser compatibility
-    // noinspection JSUnresolvedVariable
-    const matchesFunction = elt instanceof Element && (elt.matches || elt.matchesSelector || elt.msMatchesSelector || elt.mozMatchesSelector || elt.webkitMatchesSelector || elt.oMatchesSelector)
```
MichaelWest22, line 489: all supported browsers have long since moved to support .matches in the last 10 years so we don't need this fallback now

src/htmx.js
```diff
@@ -1398,13 +1383,7 @@ var htmx = (function() {
...
    */
   function shouldSettleAttribute(name) {
-    const attributesToSettle = htmx.config.attributesToSettle
-    for (let i = 0; i < attributesToSettle.length; i++) {
-      if (name === attributesToSettle[i]) {
-        return true
-      }
-    }
-    return false
+    return htmx.config.attributesToSettle.includes(name)
```
MichaelWest22, line 1386: This code can now be simplified down to one .includes now that this has been standard in all browsers for 10 years

src/htmx.js
```diff
@@ -2906,6 +2878,8 @@ var htmx = (function() {
       cleanUpElement(elt)
       return
     }
+    // Ensure only valid Elements and not shadow DOM roots are inited
+    if (elt instanceof Element) return
```
MichaelWest22, line 2882: ensure only elements that can have attribute which need hashing and can have hx-attributes are processed to avoid processing shadow dom root nodes

src/htmx.js
```diff
@@ -3027,11 +2994,7 @@ var htmx = (function() {
   }
 
   function logError(msg) {
-    if (console.error) {
-      console.error(msg)
-    } else if (console.log) {
-      console.log('ERROR: ', msg)
-    }
+    console.error(msg)
```
MichaelWest22, line 2997: console.error is standard in all supported browsers now

Telroshan (approved):
Good riddance if you ask me!

PR #3281, opened 2025-04-21
Docs: Update Link

AjaniBilby:
This link is pointing to a repo I want to change the name of.
Theoretically Github should auto redirect, but I want to fix the link just in case first.

Also I'm switching to linking to npm, that way if I ever need to change it again in the future, I can just update on NPM, and this document doesn't have to update again for me.


Telroshan (approved):

AjaniBilby:
Thank you

PR #2215, opened 2024-01-19
Fix broken oob swap error handling.

Wohlie:
## Description
For missing target on oob swap the event `htmx:oobErrorNoTarget` must be triggered (see https://htmx.org/events/#htmx:oobErrorNoTarget)

Corresponding issue:

## Testing
1. Added a new test
2. via `npm run test`

## Checklist

* [x] I have read the contribution guidelines
* [x] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [x] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [x] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #3282, opened 2025-04-21
Additional Code Coverage

MichaelWest22:
## Description
This is a follow up to the new web-test-runner testing PR with more fixes to increase code coverage reporting.  

Three new tests written to cover 3 edge cases that had no coverage before

1. calling htmx.process on a non element type triggers an is element check
2. file formData ignores blank file name now tested
3. disabling a sub child element triggers a disabled check in initNode()

When fixing this third one I found the is element disabled checking is used in multiple places without the eltIsDisabled() helper function so fixed this in three places

There was a bug with webkit testing https://github.com/microsoft/playwright/issues/5894 that had a try catch inserted to handle but after much time testing old versions I reproduced the issue again but found it was fixed in 2022 with webkit 16.0 so we can now safely remove this workaround.

There was also a bug/edge-case i found that is not currently handled perfectly when it runs removeValueFromFormData to remove an input that will be included on a form it only removes the input value and not multiple values in array. To fix this issue I moved a chunk of value lookup logic into a sub function getValueFromInput() so it can be ruesed by the removeValueFromFormData as well which solved the bug and the code coverage issue. Updated my test for this to show the impact.

FormData proxy had a block of code that could not be executed to test because it seems like an invalid fallback we don't need. checked with @Telroshan to confirm it wasn't needed.

Finally there were several element checks that I found were just not needed and found several of the JSDocs had Node when they should be Element which solved this issue.  Part of the confusion here is that resolveTarget() function can return node types instead of Element when dealing with some situations like from triggers as you can have document here as an option (maybe we need two resolveTarget functions??).  But for the main target use case it will only ever return an Element and be valid.  I managed to hack around it to make it target a non element text node but this just breaks things later as it tries to add a class to the target so its just not supported right now.

 

Corresponding issue:

## Testing
mostly tested using new testing suite

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1885,16 +1885,12 @@ var htmx = (function() {
     // preserve focus and selection
     const activeElt = document.activeElement
     let selectionInfo = {}
-    try {
-      selectionInfo = {
-        elt: activeElt,
-        // @ts-ignore
-        start: activeElt ? activeElt.selectionStart : null,
-        // @ts-ignore
-        end: activeElt ? activeElt.selectionEnd : null
-      }
-    } catch (e) {
-      // safari issue - see https://github.com/microsoft/playwright/issues/5894
+    selectionInfo = {
+      elt: activeElt,
+      // @ts-ignore
+      start: activeElt ? activeElt.selectionStart : null,
+      // @ts-ignore
+      end: activeElt ? activeElt.selectionEnd : null
     }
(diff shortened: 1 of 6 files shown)


MichaelWest22:
This PR also includes a fix that was in #2215 which I found was also a code coverage issue

PR #3283, opened 2025-04-21
proxy window.location for testing and extension overrides

MichaelWest22:
## Description
To allow loc coverage testing of hard reloads and refreshes I've proxied window.location into htmx.location that is used for .reload() and setting href.  This can then be replaced if needed to allow full testing of the reload features and also this would allow future extensions to override page reloading if required.

Also found instances of document.location usage which should all be just location to minify code.

Corresponding issue:

## Testing
Added new tests for HX-Refresh and HX-Redirect and history cache miss

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/test/attributes/hx-push-url.js
+++ b/test/attributes/hx-push-url.js
@@ -136,7 +136,7 @@ describe('hx-push-url attribute', function() {
     getWorkArea().textContent.should.equal('test1')
   })
 
-  it('cache miss should refresh whe refreshOnHistoryMiss true', function() {
+  it('cache miss should refresh when refreshOnHistoryMiss true', function() {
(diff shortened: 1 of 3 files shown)


test/attributes/hx-push-url.js
```diff
@@ -136,6 +136,17 @@ describe('hx-push-url attribute', function() {
     getWorkArea().textContent.should.equal('test1')
   })
 
+  it('cache miss should refresh whe refreshOnHistoryMiss true', function() {
```
scrhartley, line 139: "whe" -> "when"

MichaelWest22: thanks! fixed

PR #3293, opened 2025-04-26
[+Extension] `hx-keep`, enables form and html content to persist across reloads.

AjaniBilby:
I've created a library to enable [forms](https://hx-keep.ajanibilby.com/#hx-keepmode) to be cached in `localStorage` to allow them to survive refreshes, accidental tab closures, etc - with the server being able to then clear the cache with http [`hx-keep-evict` header](https://hx-keep.ajanibilby.com/#response-hx-keep-evict) when the data is successfully saved to the server.

banner

It also provides client side APIs to allow frontends like react to hook into it to store their state in a form for custom input elements.

HTML attribute prefix used: [`hx-keep*`](https://hx-keep.ajanibilby.com/#html-attributes) (with support for `data-` prefix)
Respects `hx-history="false"`

### DIFF
--- a/www/content/extensions/_index.md
+++ b/www/content/extensions/_index.md
@@ -127,6 +127,10 @@ htmx extensions are split into two categories:
       <td>{% markdown() %}  This extension allows htmx requests to be sent for drag drop  {% end %}</td>
     </tr>
     <tr>
+      <td>{% markdown() %}  [hx-keep](https://www.npmjs.com/package/hx-keep)  {% end %}</td>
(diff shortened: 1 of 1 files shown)


AjaniBilby:
This does have some overlap with the behaviour of [hx-preserve](https://htmx.org/attributes/hx-preserve/), however `hx-keep` has nothing to do with element state like video - it only considers the raw html or `FormData`.

1cg:
Thank you!

PR #3304, opened 2025-05-07
Fixes issue 1537 - OOB does not escape query selector - updated version

MAW016:
## Description
This change alters how the oobSwap function builds the query selector string.  Problem is usage of raw_id value obtained from getRawAttribute(oobElement, "id") without escaping.

> Original PR by https://github.com/FraserChapman is here: https://github.com/bigskysoftware/htmx/pull/2319 but there is no activity, so I decided to take over.

This fix allows the call querySelectorAll to return the nodelist of all elements with the ids such as "foo-/bar" - without getting "Failed to execute 'querySelectorAll' on 'Document': '#foo-/bar' is not a valid selector, and whilst not limiting the selection to the first matching id in the document.

Corresponding issue:
https://github.com/bigskysoftware/htmx/issues/1537

## Testing
I ran the htmx.js test suite Version: 2.0.4 and got 0 failures
I also manually tested ids such as "foo-/bar" and it all seems to be working as expected.

Further to this I've added test cases for

'handles elements with IDs containing special characters properly'

and

'handles one swap into multiple elements with the same ID properly'

And re- ran the htmx.js test suite Version: 2.0.4 and got 0 failures

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1444,7 +1444,7 @@ var htmx = (function() {
    */
   function oobSwap(oobValue, oobElement, settleInfo, rootNode) {
     rootNode = rootNode || getDocument()
-    let selector = '#' + getRawAttribute(oobElement, 'id')
+    let selector = '#' + CSS.escape(getRawAttribute(oobElement, 'id'))
     /** @type HtmxSwapStyle */
     let swapStyle = 'outerHTML'
     if (oobValue === 'true') {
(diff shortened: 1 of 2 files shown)


Telroshan (approved):

1cg:
Wow, I had no idea `CSS.escape()` (or even `CSS`) existed...  Thank you!

MAW016:
> Wow, I had no idea `CSS.escape()` (or even `CSS`) existed... Thank you!

Me neither, but this issue forced me to search for some elegant solution :)

Issue #2865, opened 2024-09-03
Possibly buggy cache handling when using hx-select / hx-target in htmx 2.x


heikojansen:
I have two files "first.html"

```
<!doctype html>
<html>
  <head><title>First</title></head>
  <body>
    <ul hx-boost="true" hx-select="main" hx-target="main" hx-swap="outerHTML">
      <li><a href="first.html">First</a></li>
      <li><a href="second.html">Second</a></li>
    </ul>
    <main> First </main>
    <!--script src="https://unpkg.com/htmx.org@1.9.12"></script-->
    <script src="https://unpkg.com/htmx.org@2.0.2"></script>
  </body>
</html>
```

and "second.html"

```
<!doctype html>
<html>
  <head><title>Second</title></head>
  <body>
    <ul hx-boost="true" hx-select="main" hx-target="main" hx-swap="outerHTML">
      <li><a href="first.html">First</a></li>
      <li><a href="second.html">Second</a></li>
    </ul>
    <main> Second </main>
    <!--script src="https://unpkg.com/htmx.org@1.9.12"></script-->
    <script src="https://unpkg.com/htmx.org@2.0.2"></script>
  </body>
</html>
```

served by some (any) webserver from the same directory.

I open ```http://.../first.html``` in some browser (I tried Firefox, Chromium and MS Edge (Linux)) and perform the following steps:

| Action                          | URL in browser after action | content shown after action |
| ----------------------------- | ------------------------------------- | ------------------------------------ |
|                                      | first.html                               | "First"                                    |
| Click on link "Second" | second.html                         | "Second"                               |
| Browser "back"            | first.html                              | "First"                                    |
| Click on link "Second" | second.html                        | "Second"                                |
| Browser "back"            | first.html                              | **"Second"**  <-- _WTF?_      |

After the second time I stepped back through the browser history, URL and content have diverged!

Now if I **remove** ```hx-select="main" hx-target="main" hx-swap="outerHTML"``` and start again (SHIFT+F5 hard reload for "first.html") everything works.

Also, if I **keep** ```hx-select="main" hx-target="main" hx-swap="outerHTML"``` but **switch to HTMX 1.9.12** everything works, too. 

So to me this really looks like a bug in HTMX 2.x?!

Can anyone confirm if it is a bug or show me what I'm doing wrong?

Thanks!
  Heiko

MichaelWest22:
yeah I was able to reproduce this and it does look like a bug to me.

[...]

PR #3305, opened 2025-05-08
Move History storage to sessionStorage and history path to window

MichaelWest22:
## Description
moving history cache to sessionStorage and fixing storage of current history path

Currently htmx stores its history cache in localStorage. However the native browser bfcache we are trying to simulate treats its cache like sessionStorage where it is stored per browser session tab and not in permanent site storage like localStorage.  I think when the feature was first implemented localStorage was picked because it was more familiar.  When localStorage is used the cache persists between sessions and browser restarts and between multiple tabs for the same domain.  The means the cache is constantly fill with old cache state from previous sessions which can't be accessed as only url's pushed and popped in the current session will work and these will update these urls cache data.  As well as the old inaccessable cache data stored there is also the risk that a user opening multiple tabs and going back on one tab will end up restoring the page state data from another tab which is undesirable.  

Simply renaming localStorage to sessionStorage in the code resolves this issue but I've left the internalAPI.canAccessLocalStorage function with the old name to avoid breaking any use of this api. local and session storage will either both be enabled or disabled together so I don't think the naming is critical.

The other Issue I've addressed is the currentPathForHistory state variable.  This variable is used to save the current URL you are on when a back/forward action popstates and changes the url so the current page can be captured against this URL for restoration later.  This variable needs to stay intact during history restoration process and it is currently just a local to htmx let variable.  If htmx script is re-executed during history navigation when the script is placed in the body this variable state is lost and htmx can backup history with the wrong URL causing wierd history navigations.  We could move this path string to sessionStorage but I found this doable but a bit cumbersome.  I think this variable makes sense to instead hoist into the window scope with a htmx prefixed name htmxCurrentPathForHistory which resolves all these issues.  

Corresponding issue:
#2865 

## Testing
Ran test suite and also had to test history back forward actions in a test application to to validate it works as expected.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -82,7 +82,7 @@ var htmx = (function() {
        */
       historyEnabled: true,
       /**
-       * The number of pages to keep in **localStorage** for history support.
+       * The number of pages to keep in **sessionStorage** for history support.
        * @type number
        * @default 10
        */
@@ -819,10 +819,10 @@ var htmx = (function() {
    * @returns {boolean}
    */
   function canAccessLocalStorage() {
-    const test = 'htmx:localStorageTest'
+    const test = 'htmx:sessionStorageTest'
     try {
-      localStorage.setItem(test, test)
-      localStorage.removeItem(test)
+      sessionStorage.setItem(test, test)
+      sessionStorage.removeItem(test)
       return true
     } catch (e) {
       return false
@@ -3033,7 +3033,7 @@ var htmx = (function() {
   //= ===================================================================
   // History Support
   //= ===================================================================
-  let currentPathForHistory = location.pathname + location.search
+  window.htmxCurrentPathForHistory = location.pathname + location.search
 
   /**
    * @returns {Element}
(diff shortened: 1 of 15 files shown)


raysuelzer:
Was this merged? I think this might be the problem i am having. I could not figure out why my history navigation was saving the wrong content for the bath. When I clicked a link, it would put the new content in the previous link. But, maybe it's because i had multiple tabs open refreshing the page which caused multiple items to get added or something?

MichaelWest22:
@raysuelzer any chance you can do a quick test with this PR's version of htmx.js manually in your project to check if this does resolve your issue or not?

https://github.com/MichaelWest22/htmx/blob/historyCache/src/htmx.js

raysuelzer:
Hmmm. Unfortunately, when I press back I am still getting the contents of current page, even though the correct contents appear to be in the history. Basically, HTMX seems to stop working when i use the back button when navigating between these two pages. I've checked that i am not using any redirects... 

I'm going to keep debugging this... 

[Uploading screen-capture (2).webm…]()

raysuelzer:
Update, actually that did fix the issue. Apparently, 'htmx-class-tools-ext' when imported via import statement also imports htmx, that means that your version was getting overwritten.

MichaelWest22:
Just updated PR to remove the second change so I can Do this second path change as a separate PR for review

PR #3315, opened 2025-05-16
Fix missing TypeScript property

scrhartley:
## Description
Even though it's in the docs for  [`htmx:beforeSwap`](https://htmx.org/events/#htmx:beforeSwap) and the event relies on it, the property `detail.requestConfig.elt` is missing from the TypeScript definition of HtmxRequestConfig.

Looking at the code it's not obvious where it's coming from:
after `requestConfig` is defined, `triggerEvent` is immediately called which adds `elt` to the passed in object.

Fix the type definition and explicitly add `elt` to the `requestConfig` definition to lessen the confusion.

## Testing
Ran dist to make sure the type turned up in the generated .d.ts file.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4355,6 +4355,7 @@ var htmx = (function() {
       unfilteredFormData: allFormData,
       unfilteredParameters: formDataProxy(allFormData),
       headers,
+      elt,
       target,
       verb,
       errors,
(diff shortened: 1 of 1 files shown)


Telroshan (approved):
Hey, good catch.

It's indeed confusing right now, `requestConfig.elt` is set by `triggerEvent` but it looks like a side effect more than a feature, so I think it's reasonable to add an explicit property declaration here.
However, it should be noted that this is effectively "useless" in the way that `triggerEvent` will override the value no matter what.

PR #2845, opened 2024-08-23
fix(swap): apply swap delay in swap function instead of handleAjaxResponse

boonedox:
## Description

moved application of swap delay to swap function so that the JS api can also use the swap delay

Corresponding issue:

## Testing
manually tested in my local environment, both with the JS api and with normal hx requests

## Checklist

* [ ] I have read the contribution guidelines
* [ ] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [ ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [ ] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #3328, opened 2025-06-02
move delay and view transitions to inside swap function for api and history use

MichaelWest22:
## Description
This change is just moving the delay and view transition wrapping code to the inside of the swap function.  There were seperate PR's to add these features but it is much cleaner to combine these ideas together. This means that api users like the sse extension can make use of delays and view transitions now and along with my changes in #3306 will allow view transitions to function during history restorations which would now use the standard swap function. for history swaps the view transition default config would need to be set.

There is no new code introduced in this PR and the total line count is the same with the exact same code just relocated to the inside of the swap function.  The only exception is a new beforeSwapCallback that had to be added but this is just relocating existing code unchanged.  because the indentation of the existing code is updated by this change the diff looks un-readable sorry so I split this PR into two commits.  if you click into the first commit you can review the real code changes before i fixed the indentation so it is easy to read and the second commit just changes the indentation and adds tests.

Note one issue with the original view transitions on history PR was that it needed `htmx:beforeTransition` event used but this should all be working fine in this PR.

Note that #3306 should ideally be merged first and then I may have to rebase to fix a couple of merge conflicts to get this one to merge

Corresponding issue:
#1703 view transitions on history issue
#2873 view transitions on history PR this replaces
https://github.com/bigskysoftware/htmx-extensions/issues/160 view transtions on SSE
https://github.com/bigskysoftware/htmx-extensions/issues/127 swap delay on SSE
#2845 swap delay PR this replaces

## Testing
Wrote some new tests and reused a test from the previous PR and did some manual testing to confirm view transitions still function as expected.

## Checklist

  source changes)
  approved via an issue


RobMaskell:
@MichaelWest22 I saw that the other PR for fixing the swap delay issue from the API was merged into dev the other day, so thought I'd test it against my code and check it was working. It is, but I don't know what that means for your PR which seems to fix more than the other one I think

MichaelWest22:
@RobMaskell Yeah i took the changes from that now merged MR and combined them with the view transition idea plus adapted them to work with my possibly history fixes to create an improved version of the delay PR plus combine the other features while just re-using the existing code without any real changes which avoids a couple of minor issues with now merged delay version.  But I've rebased this now so it can augment the now merged changes.  It will not have any impact on the delay feature that is already merged if this one gets accepted later.

PR #3234, opened 2025-03-15
Replace jsDelivr w/ UNPKG

SukkaW:
## Description

UNPKG has not been actively maintained and has been down since `Mar 15, 2025` (https://github.com/unpkg/unpkg/issues/412).

The PR replaces UNPKG usage with jsDelivr in the following places:

- Docs
- Previous blog articles
- Demo and Templates

The unpkg.com usage inside the test cases are not replaced/modified.

## Checklist

* [x] I have read the contribution guidelines
* [x] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [ ] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [x] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #3334, opened 2025-06-10
Follow up to #3234

scrhartley:
## Description
Add missing update for migration from UNPKG to jsDelivr.

PR #3234 uses references to `https://cdn.jsdelivr.net/npm/htmx.org`, relying on the default file in `package.json`.
For UNPKG this relies on the `unpkg` property being set to the non-esm file, but this hasn't been migrated for jsDelivr.
The [docs for jsDelivr](https://www.jsdelivr.com/documentation#id-configuring-a-default-file-in-packagejson) say that there is an equivalent `jsdelivr` property, so switch to using that.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   ],
   "main": "dist/htmx.esm.js",
   "types": "dist/htmx.esm.d.ts",
-  "unpkg": "dist/htmx.min.js",
+  "jsdelivr": "dist/htmx.min.js",
   "web-types": "editors/jetbrains/htmx.web-types.json",
   "scripts": {
     "dist": "./scripts/dist.sh && npm run types-generate && npm run web-types-generate",


yawaramin:
Just to explain to anyone seeing this PR, it would allow people to use jsDelivr's short-form URL like https://cdn.jsdelivr.net/npm/htmx.org@2.0.4 in their `<script>` tags, instead of having to explicitly give the path like https://cdn.jsdelivr.net/npm/htmx.org@2.0.4/dist/htmx.min.js

scrhartley:
The previous PR already uses the short form in `www/content/examples/modal-custom.md` and `www/static/js/demo/it.js`. In fact references the version-less variant which grabs the latest one published to npm.

Looking at the [modal-custom](https://htmx.org/examples/modal-custom/) page, then unless there's something more complicated going on, there might be a mistake since there seem to be competing script tags in the head and body.
In the head:
```html
<script src="/js/htmx.js"></script>
...
<script src="/js/_hyperscript.js"></script>
```
In the body (changed to jsDelivr in the previous PR, but not yet live):
```html
<script src="https://unpkg.com/htmx.org"></script>
<script src="https://unpkg.com/hyperscript.org"></script>
```

scrhartley:
Thinking about this some more, should we include properties for both CDNs in the `package.json`, rather than one or the other?

yawaramin:
Yeah, actually, let's keep both. We shouldn't delete the one for Unpkg just because our docs don't mention it any more. People might still want to use it.

scrhartley:
Before #3234, the UNPKG URLs tended to use the short forms.
Once this is merged and the new release done then we could also do another pass and move to short form URLs for jsDelivr.
Currently we can't do that since without this change the short form URL goes to the ESM file (an auto minified version):
https://cdn.jsdelivr.net/npm/htmx.org@2.0.4 (ESM script)
https://cdn.jsdelivr.net/npm/htmx.org@2.0.4/dist/htmx.min.js (normal script)

yawaramin:
As I mentioned in that PR, I would like to use shortcodes for the script source URLs so that any change going forward (eg updating a version number) wouldn't have to touch multiple files. We can do the short URL change along with that too.

Issue #2743, opened 2024-07-19
triggerEvent relies on a non-null target existing; therefore beforeSwap won't fire if target is null (possible when hx-retarget is used)


reubenfirmin:
This is additional color on https://github.com/bigskysoftware/htmx/issues/2216

If hx-retarget is used, it's very possible that the dom has changed by the time a response arrives back to the browser (e.g. there might be a longer running request to render an image, and the user might have navigated elsewhere). This causes the error described in the above bug.

There's unfortunately no way to work around this bug, because none of the events will fire in the case of a null target. This is because `triggerEvent` uses the target element itself to dispatch the event, as can be seen here: https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L2933

The fix should likely be to make triggerEvent robust to a null target, and e.g. fall back to the body or document in the case of the provided target being null.

MichaelWest22:
Resolved in 2.0.5

Issue #2216, opened 2024-01-19
HX-Retarget with a selector that has no elements results in a null error


lsymds:
# Overview

When utilizing the `HX-Retarget` header with a selector that returns no elements, a null error when trying to access that non-existent target's `classList` occurs.

```
htmx.org@1.9.10:1 TypeError: Cannot read properties of null (reading 'classList')
    at Mr (htmx.org@1.9.10:1:43638)
    at b.onload (htmx.org@1.9.10:1:40710)
```

I believe this is because the retargeting logic does not validate or do anything if the resultant target is null. Other areas of the code such as `issueAjaxRequest` when using the DOM-based `hx-target` attribute *do* perform this check and raise an error event if it cannot be found:

```
htmx.org@1.9.10:1 htmx:targetError
```

I think it's worth doing similar when retargeting.

# Example

```go
package main

import "net/http"

func main() {
	s := http.NewServeMux()

	s.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte(`
			<!doctype html>
			<html lang="en">
				<head></head>
				<body>
					<div class="target-container"></div>
					<button hx-get="/component">Do something</button>
					<script src="https://unpkg.com/htmx.org@1.9.10"></script>
				</body>
			</html>
		`))
	})

	s.HandleFunc("/component", func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("HX-Retarget", "closest .anything-goes-here")
	})

	http.ListenAndServe("localhost:9389", s)
}
```

lsymds:
Happy to PR this if a maintainer is happy with the approach I suggested. It's already a broken path in the code, so adding that error logic shouldn't cause any regressions.

webdevcody:
I just ran into this today.  I more useful error message saying it couldn't find an element with selector of header('HX-Retarget') would be very nice

tschuehly:
I often use Hx-Retarget, so this would be really helpful if the error would be more descriptive

mleonhard:
I think I'm hitting this via the `target` option in the `responseHandling` config:
```jsx
<meta name="htmx-config"
      content={JSON.stringify({
          // https://htmx.org/quirks/#by-default-4xx-5xx-responses-do-not-swap
          responseHandling: [
              {code: "204", swap: false},
              {code: "2..", swap: true},
              {code: "422", swap: true},
              {
                  code: "[45]..",
                  swap: true,
                  error: true,
[...]

PR #2222, opened 2024-01-20
Handle non-existent HX-Retarget element

lsymds:
## Description

As per the issue, HX-Retarget performed no handling of whether the resultant target element actually existed or not. This PR adds error handling seen in other places and gracefully returns from the function. 

Corresponding issue: Resolves #2216 

## Testing

Tested via an automated test and manually with the reproduction steps in the original issue.

## Checklist

* [x] I have read the contribution guidelines
* [x] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [x] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [x] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #3335, opened 2025-06-10
throw targetError correctly when target invalid during retarget

MichaelWest22:
## Description
Found this old issue tagged as a bug that has yet to be resolved.  when you use Re-Target header or the responseHandling to set a new target and the target can't be found instead of throwing a valid targetError to alert the web designer of the issue and allow them to trap it with the htmx:targetError event it just breaks trying to add swapping class to a null target element.  

Created a new resolveRetarget() function to do the checking and replaced the just added handleRetargetHeader() function.  Also found the handleRetargetHeader() function was being called twice by mistake which seems like an old oversight.  Checked the code paths carefully and confirmed the initial call is redundant as it only side effect of running it the first time is updating responseInfo.target but this field is not accessed or read before the second time the hx-retarget header is checked so removing the first call will have zero impact.

Corresponding issue:
#2743
#2216 

## Testing
Added tests for all the failed cases this covers and ran the test suite

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4716,20 +4716,20 @@ var htmx = (function() {
   }
 
   /**
-   * Updates the responseInfo's target property if an HX-Retarget header is present
-   *
-   * @param {XMLHttpRequest} xhr
-   * @param {HtmxResponseInfo} responseInfo
+   * Resove the Retarget selector and throw if not found
    * @param {Element} elt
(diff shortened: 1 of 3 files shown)


MichaelWest22:
Looks like this PR actually replaces a stalled one #2222

Issue #3316, opened 2025-05-17
my temp fix for Troublesome Tables and lists


dev-hosamhamdy:
hello am using WS extension
in this example i sumbit form with adding new group name and on response i reset the form and add new added group to table body

the request form and table :
```
<form id="add_group" ws-send hx-swap="innerHTML">
  <input type="hidden" name="message_type" value="addGroup">
  <input type="text" id="group" name="name" placeholder="Group Name" value="" required="" autocomplete="off">
  <button type="submit">Add Group</button>
</form>
<table id="test_table">
  <thead>
    <tr>
      <th>Group</th>
      <th>Members</th>
    </tr>
  </thead>
  <tbody id="group_table_body">
  </tbody>
</table>
```

the response :
```
<form id="add_group" ws-send hx-swap="innerHTML">
  <input type="hidden" name="message_type" value="addGroup">
  <input type="text" id="group" name="name" placeholder="Group Name" value="" required="" autocomplete="off">
  <button type="submit">Add Group</button>
</form>

<template hx-swap-oob="afterbegin:#group_table_body">
  <tr>
    <td id="g174_1" hx-swap="innerHTML">new_group</td>
  </tr>
</template>
```

this example isn't working the form tag get swapped but the tr tag or similar tags not getting swapped so :
1- am asking if someone can explain why it's not working and how to fix it
2- i had to do the standalone tags manual update by my self so created manual update function that work with this example to help me until i got more robust solution
here's js file contain the function
1 - let htmx do the swap normally complete his swap flow
2 - after that i extract the template tag from response
3 - swap it manually

```
document.body.addEventListener('htmx:wsAfterMessage', function (evt) {
    const html = evt.detail.message;
    if (html.includes('<template')) { handleTemplateFragments(html); }
});

function handleTemplateFragments(html) {
    const parser = new DOMParser();
    const doc = parser.parseFromString(html, 'text/html');
    const templates = doc.querySelectorAll('template');
    templates.forEach(template_tag => {
        const extSel = template_tag.getAttribute('data-manual-update') || template_tag.getAttribute('hx-swap-oob');
        if (!extSel) return;
        const nodes = Array.from(template_tag.content.childNodes).filter(n => n.nodeType === Node.ELEMENT_NODE);
        nodes.forEach(node => manualUpdate(node, extSel));
    });
}

function manualUpdate(node, extSel) {
    const [position, selector] = extSel.split(/:(.+)/);
    const container = document.querySelector(selector);
[...]

PR #3343, opened 2025-06-19
Add outerHTMLStrip and handle template tag encapsulation

MichaelWest22:
## Description
This is a possible enhancement Jeremy Howard suggested.  When doing `outerHTML` oob swaps there are situations where you may want to replace an element on the page with multiple nodes and this is currently not possible with `outerHTML` oob swaps as they can only replace it with a single node.  This proposed change would add a new swap type `outerHTMLStrip` that strips the encapsulating tag during swap like all the inner swap styles do but performs a `outerHTML` swap to replace the target with all the inner contents of the oob tagged element.

```HTML
<div id="foo" hx-swap-oob="outerHTMLStrip">
  <div id="foo2">
    One thing
  </div>
  <div>
    Another thing
  </div>
</div>
```

While I found you can do this sometimes with the following response currently:

```HTML
<div id="foo2" hx-swap-oob="outerHTML:#foo">
    One thing
</div>
</div hx-swap-oob="afterend:#foo2:>
  <div>
    Another thing
  </div>
</div>
```
Which will have the same results but it is much harder to discover or understand than just being able to control the tag stripping of the outerHTML swap

To implement this feature I just add this new `outerHTMLStrip` swap style support by letting this swap style be detected as a non inline swap style which means  the parent of the swap data becomes the content itself instead of a wrapping fragment so it uses inner content for the swap and then we just remove the word "Strip" from the swap style to allow it to be processed as outerHTML.  

Another related fix I've included is to get the none inline swaps inner content assignment has a bug where template tags are broken as they have .content and not normal inner content nodes.  So just adding support for template.content and falling back to the old way if its not a template solves this issue

This allows more generic use of oob swaps in templating libraries as you can just use template tags like this now:

```HTML
<template id="foo" hx-swap-oob="outerHTMLStrip">
  <div id="foo2">
    One thing
  </div>
  <div>
    Another thing
  </div>
</template>
```

This allows you to ignore the troublesome tables issues that are a real pain to debug as template tags can contain any element type and you don't have to match the encapsulating tag to be the perfect <table> tag type to match the content type to avoid issues.

Existing <template></template> wrapping use cases work the same as before after this change it just adds more template tag support.

Possible Changes:

The swap name `outerHTMLStrip` could be renamed to something else if a better alternative name can be found
Could split this into two separate PR's

Also I found 
Corresponding issue:
#3316 

## Testing
Added new tests and ran full test suite

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1478,8 +1478,11 @@ var htmx = (function() {
           fragment = getDocument().createDocumentFragment()
           fragment.appendChild(oobElementClone)
           if (!isInlineSwap(swapStyle, target)) {
-            fragment = asParentNode(oobElementClone) // if this is not an inline swap, we use the content of the node, not the node itself
+            // @ts-ignore if elt is template content will be valid so use as inner content
+            fragment = oobElementClone.content || asParentNode(oobElementClone) // if this is not an inline swap, we use the content of the node, not the node itself
           }
+          // outerHTMLStrip swapStyle will strip wrapping tag above but do outerHTML swap of inner contents
+          swapStyle = swapStyle.replace('Strip', '')
 
           const beforeSwapDetails = { shouldSwap: true, target, fragment }
           if (!triggerEvent(target, 'htmx:oobBeforeSwap', beforeSwapDetails)) return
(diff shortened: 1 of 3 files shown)


MichaelWest22:
closing this old idea in favor of #3352

PR #3383, opened 2025-07-21
Update CSRF recommendation

### LINKED ISSUE #3379: CSRF prevention documentation might need clarification when used with hx-boost
The currently [recommended approach](https://htmx.org/docs/#csrf-prevention) of adding `hx-headers` to `body` (or `html`) don't seem to be working well when used together with `hx-boost`, since these attributes are not picked when swapping (related to this issue: https://github.com/bigskysoftware/htmx/issues/1384).
The client keeps using the same token and is getting errors.

Since it feels like a common use case, it would be nice to call this out in the docs, as well as a recommended fix?

For now I've managed to workaround it by using a dummy div inside `body`:

```html
<body>
   <div hx-headers={"{\"X-CSRF-TOKEN\": \"#{get_csrf_token()}\"}"}>
       ...
   </div>
</body>
```

Another workaround based on this [suggestion](https://github.com/bigskysoftware/htmx/issues/1384#issuecomment-2225116919) also works for me, but seems more complex:

```js
htmx.on("htmx:afterSwap", ({ detail }) => {
  if (detail.target.tagName != "BODY") return;

  const parsedResponse = new DOMParser().parseFromString(
    detail.xhr.response,
    "text/html"
  );
  const hxHeaders = parsedResponse.body.getAttribute("hx-headers");
  document.body.setAttribute("hx-headers", hxHeaders);
});
```

yawaramin:
Recommend using a form with a hidden input or an adjacent CSRF token in the header. Stop recommending a global token in the `html` or `body` tags as those may not work with `hx-boost`.

Fix #3379

### DIFF
--- a/www/content/docs.md
+++ b/www/content/docs.md
@@ -1697,6 +1697,8 @@ for exploring this topic.
 
 The assignment and checking of CSRF tokens are typically backend responsibilities, but `htmx` can support returning the CSRF token automatically with every request using the `hx-headers` attribute. The attribute needs to be added to the element issuing the request or one of its ancestor elements. This makes the `html` and `body` elements effective global vehicles for adding the CSRF token to the `HTTP` request header, as illustrated below. 
 
+Note: `hx-boost` does not not update the `<html>` or `<body>` tags; if using this feature with `hx-boost`, make sure to include the CSRF token on an element that _will_ get replaced. Many web frameworks support automatically inserting the CSRF token as a hidden input in HTML forms. This is encouraged whenever possible.
+
 ```html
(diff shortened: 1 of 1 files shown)


alexpetros:
Would you mind not doing the rewrite and instead just adding a note to the effect of "Note that hx-boost does not not update the `<html>` or `<body>` tags; if using this feature with `hx-boost`, make sure to include the CSRF token on an element that will get replaced."

alexpetros:
Awesome, thanks @yawaramin :)

sabiwara:
Thank you @yawaramin @alexpetros ! 💜

Issue #2329, opened 2024-02-17
HTMX doesn't work if included with `async`


MonstraG:
Basically, https://github.com/bigskysoftware/htmx/issues/142, except again. And there are no errors in the console.

![Recording 2024-02-17 at 10 22 48](https://github.com/bigskysoftware/htmx/assets/17963791/33942e99-7a21-4b29-b258-f37c1696ddc2)

aburd:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

> async- For classic scripts, if the async attribute is present, then the classic script will be fetched in parallel to parsing and evaluated as soon as it is available.

> [defer](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer)
This Boolean attribute is set to indicate to a browser that the script is meant to be executed after the document has been parsed, but before firing [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event).

I could see the async tag being problematic for htmx, because it needs to attach event handlers to all the elements, but if it’s doing it in parallel to document parsing, couldn’t this cause a race condition? Maybe what you want is the defer attribute.

MonstraG:
I would want async to get the fastest page load. I would also expect htmx to wait for DOMContentLoaded already.

aburd:
Well given your issue, using the defer attribute would probably solve your problem and do it even faster since it is “executed after the document has been parsed, but before firing [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event)”. 

But you’re right, in the absence of documentation it should just work. It should be fixed to work regardless of those attributes, or at least warn you in the console that the script is not meant to work with async if it is by design.

MonstraG:
I can confirm that defer doesn't break it

MonstraG:
Just tried on `2.0.0-beta1`, same problem.

f-liva:
Me too

morganwahl:
I'm also running into this. Script tag without `async` or `defer`, script tag with `defer` and even loading it as an ESM all work fine. Loading it specifically with `async` as a classic script (not an ESM) results in none of HTMX's event listeners being added.

Note that this behavior is not consistent, since it will be dependent on how long it took the browser to fetch HTMX over the network. For me, doing a hard-refresh of a page hits this problem about half the time.

The problem seems to be in the `ready` function:

```javascript
        //====================================================================
[...]

PR #3398, opened 2025-07-30
Fix ready() to work even when HTMX is loaded async

morganwahl:
## Description

This updates the `ready` function to work correctly even if executing after `DOMContentLoaded` has been fired, but before `readyState` is `"complete"`. This is apparently possible when loading the script async (using the `async` attribute on the script tag.)

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/2329

## Testing

I have a page in another project where, in about half of page-loads, HTMX fails to call the anonymous initialization function passed to `ready`. Whether it happens or not depends no the exact network timing. By putting in breakpoints, I can see that in the cases where it works correctly, it's because `readyState` is `"completed"`. In other cases, that conditional is `false` and it adds the event listener for `DOMContentLoaded`. However the listener is never called. (In all the runs I've looked at, `isReady` is always `false`.)

I verified this fix by making the same change in the copy of HTMX in that project. The page loaded correctly 10 times in a row.

I wasn't able to add new tests since I'm not really sure how to test this.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -5056,25 +5056,31 @@ var htmx = (function() {
   //= ===================================================================
   // Initialization
   //= ===================================================================
-  var isReady = false
-  getDocument().addEventListener('DOMContentLoaded', function() {
-    isReady = true
-  })
-
   /**
-   * Execute a function now if DOMContentLoaded has fired, otherwise listen for it.
+   * Execute a function once the DOM has loaded.
    *
-   * This function uses isReady because there is no reliable way to ask the browser whether
-   * the DOMContentLoaded event has already been fired; there's a gap between DOMContentLoaded
-   * firing and readystate=complete.
+   * If the DOM is still loading, the function will be run as a listener for
+   * the DOMContentLoaded event. Otherwise, the function will be synchronously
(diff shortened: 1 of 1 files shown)


src/htmx.js
```diff
@@ -5056,25 +5056,31 @@ var htmx = (function() {
...
   /**
-   * Execute a function now if DOMContentLoaded has fired, otherwise listen for it.
+   * Execute a function once the DOM has loaded.
+   *
+   * If the DOM is still loading, the function will be run as a listener for
+   * the DOMContentLoaded event. Otherwise, the function will be synchronously
+   * executed now.
+   *
+   * This is tricky to get right since the DOMContentLoaded event only fires
+   * ones. We need to ensure we always wait for it if the DOM is still loading,
```
aburd, line 5067:
```suggestion
   * once. We need to ensure we always wait for it if the DOM is still loading,
```

1cg:
Closing b/c I believe this breaks `defer` and this is some of the touchiest (and most untestable) code in htmx.  Thank you for looking at it!

Issue #3115, opened 2025-01-09
Response from HX-Location should respect HX-Push-Url header


devashishdxt:
I have a scenario where I set the `HX-Location` header to redirect the client to other page (assume `/x`). In some cases, the response (of `/x`) could contain `HX-Push-Url` header (assume `/y`) which is not being respected, i.e., the page url is still same as the url in `HX-Location` header (`/x`).

Is there a way to push the url in `HX-Push-Url` (`/y`) header of response in browser history after the redirect using `HX-Location` header?

Telroshan:
From what I see in the sources:

https://github.com/bigskysoftware/htmx/blob/e0905ff94a3623f6383278bf2ea740c65b616105/src/htmx.js#L4692-L4694
https://github.com/bigskysoftware/htmx/blob/e0905ff94a3623f6383278bf2ea740c65b616105/src/htmx.js#L4703-L4705

I don't see any reason `HX-Push-Url` would not be applied from the second endpoint's response, however it looks like htmx ignores whether the redirected endpoint submitted any instruction about pushing a URL or not, and uses `pushUrlIntoHistory` _after_ the request is made to push the redirected URL anyway

So, I would expect your history to actually contain the URL sent back by your second endpoint's `HX-Push-Url` header as you would expect, but likely one step backwards in the history (which would be an issue of course).

Could you confirm if this is actually what is happening? Do you see the desired URL when going back in the history by 1 step? If so, we can likely add a fix for that

MichaelWest22:
also see #3045 for more detail on the cause

MichaelWest22:
The issue is the ajaxHelper function is waits to complete fully as a promise which will push the hx-push-url to history and 'then' it will push the hx-location URL every time.  We could add 

```javascript
      ajaxHelper('get', redirectPath, redirectSwapSpec).then(function() {
        if (!redirectSwapSpec?.noPushUrl) pushUrlIntoHistory(redirectPath)
      })
```
to add a new optional key to hx-location data spec that if set true would block the post request push if required.  

Maybe a better way would be this

```javascript
      const oldPath = currentPathForHistory
      ajaxHelper('get', redirectPath, redirectSwapSpec).then(function() {
        if (oldPath === currentPathForHistory) pushUrlIntoHistory(redirectPath)
      })
```

devashishdxt:
> Could you confirm if this is actually what is happening? Do you see the desired URL when going back in the history by 1 step? If so, we can likely add a fix for that

[...]

Issue #3114, opened 2025-01-09
Proposal: `htmx.boost()` Javascript API


AjaniBilby:
Currently there is no way for client side code to easily trigger a `hx-boost` like navigation, and since functions like `saveCurrentPageHistory`, and `restoreHistory` are not exposed it means an external implementation either needs a lot of work arounds, or has faulty history navigation.

**Example External Implementation**
```ts
export async function navigate(href: string, pushUrl = true) {
  if (typeof window !== "object") return;

  const url = new URL(href, window.location.href);
  if (url.host !== window.location.host) {
    window.location.assign(href);
    return;
  }
  
  await htmx().ajax("GET", href, {
    target: 'body',
    swap:   'outerHTML'
  });

  // this has a shallow history, and back navigation will not change the page's content correctly
  if (pushUrl) history.pushState({ htmx: true }, document.title, url.href);
}
```

### Scenario
If the majority of a site is server-side rendered, and you use lazy loading of dynamic content (i.e. a user's profile which would be `Cache-Control: private`) allowing the resest of the page to be publicly statically cached - and the lazy loads are protected with a `hx-preserve` so it's not reloaded on a boosted `<a>`.

If a client mounted component which needs to be able to trigger page navigation it will currently destroy all of your `hx-preserve`s because the only reliable way to navigate with history is a full page navigation.
Depending on theming and browser specific behaviour this can also cause flicker which is not present when doing a boosted load.

### Proposal
There should be an extra function added to the htmx js api (i.e. `boostTo`, `navigate`), which takes a `href: string` and `pushUrl?: boolean`, then performs behaviour similar to clicking on a `<a hx-boost="true" href="...">...</a>` tag in the top level of the body (so context-less with the exception of any attributes currently on the `document.body`).

If the htmx team agrees that this should be a feature, I would be willing to put in the time to create a PR with this feature added, however as I'm not familia with the code base possibly someone else would be more efficient at implementing this. But I certainly don't want to put in the effort of making a PR just for the idea to be possibly rejected.

MichaelWest22:
Could you get this to work right now with a hidden
```html
<a hidden hx-boost="true" href="...">...</a>
```
and then a .click() in JS would work as you expect.

[...]

Issue #2744, opened 2024-07-20
Add the `pushUrl` option for `htmx.ajax()` calls


wout:
Currently, there's no way to configure `hx-push-url` programatically using `htmx.ajax()`. It would be useful to have a `pushUrl` option there as well.

As far as I know, the only solutions are sending a `'HX-Push-Url'` header in the response or using `{ source: element }` where the referenced element has a `hx-push-url` attribute. The first is hard to maintain, and the second is quite hackish.

ALameLlama:
I've needed to do this to "fake" hx-boost on elements that aren't forms/a tags, would be nice to have support for this directly using htmx.ajax()

MichaelWest22:
added a possible PR for this feature

Issue #2706, opened 2024-07-11
HX-Location feature request


Wookiee81:
Currently HX-Location auto pushes the url to history (expected _default_ behaviour).
```
3499 | if (hasHeader(xhr, /HX-Location:/i)) {
3500 | saveCurrentPageToHistory();
3501 | var redirectPath = xhr.getResponseHeader("HX-Location");
3502 | var swapSpec;
3503 | if (redirectPath.indexOf("{") === 0) {
3504 | swapSpec = parseJSON(redirectPath);
3505 | // what's the best way to throw an error if the user didn't include this
3506 | redirectPath = swapSpec['path'];
3507 | delete swapSpec['path'];
3508 | }
3509 | ajaxHelper('GET', redirectPath, swapSpec).then(function(){
3510 | pushUrlIntoHistory(redirectPath);
3511 | });
3512 | return;
3513 | }
```
If possible could we gain some control over `pushUrlIntoHistory(redirectPath);` specifically gain the ability to set to 'false' or a custom location? 
An alternative would be to let HX-Replace-Url response header work with it; currently it does not seem to fire after this swap so does not work in this instance.
HX-Trigger works with HX-Location (which is why I was using it) as I want to sent custom trigger events with the redirect but I want to specify the url pushed or have none at all, this might _not_ be intended behaviour (a bug). HX-Trigger does not work with standard Location response header or I would have just used that.

Sorry if this is a dumb request (there was a simple solution to my issue in Django). But it strikes me that this functionality could be useful to more people than just my edge case here. The HX-Location is described as doing a client side redirect _**without**_ a full page reload. As such the push url seems to be something we might want more control over.

Disregard if I am just an Eedjit.

Wook.

Telroshan:
I wonder if we should maybe simply reuse the existing history update events here as well?
https://github.com/bigskysoftware/htmx/blob/89b9e2afa5fbc62ab7a22292e84f41d7da25447e/src/htmx.js#L4699C28-L4702

Wookiee81:
I wrote my solution and tested it. Fairly simple change. I can put in a pull request if anyone would be interested.

```
if (hasHeader(xhr, /HX-Location:/i)) {
      saveCurrentPageToHistory()
      let redirectPath = xhr.getResponseHeader('HX-Location')
      var pushPath
      /** @type {HtmxAjaxHelperContext&{path:string}} */
      var redirectSwapSpec
      if (redirectPath.indexOf('{') === 0) {
[...]

PR #3404, opened 2025-08-04
Add pushUrl option to ajax api and improve hx-location push url handling

MichaelWest22:
## Description
Currently the ajax api can not perform hx-push-url style history url updates easily as the only way to do it is by creating a dummy element with hx-push-url attribute and then using this as the source element of the ajax api request.  Just adding a push option to the context you can pass in solves this by allowing this variable to be used to pass in the url to push or true.  Also hx-location can make use of this same option to allow it pass an optional push value or 'false' if required and it default to 'true' which avoids manual history handling code in hx-location. 

There is a minor bug that if you use a hx-location to redirect to a page the return hx-push-url the old way of manually updating history after hx-location finishes overwrites the hx-push-url request in the response leading to wierd history state.  see #3115.  By moving the push url option into hx-location but allowing it to be updated in the response resolves this issue

Also added a replace option As I found in #3114 that they wanted to also do replace sometimes and this is easy to add as well.  Also found the determineHistoryUpdates function I had to update was confusing and repeated a lot of logic so re-wrote this function as well to make it simpler and easier to debug and understand.

Corresponding issue:
#3115
#2744 
#2456
#2706

## Testing
Added tests for the ajax api and hx-location to test out all the new options available.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4671,8 +4671,8 @@ var htmx = (function() {
     const requestPath = responseInfo.pathInfo.finalRequestPath
     const responsePath = responseInfo.pathInfo.responsePath
 
-    const pushUrl = getClosestAttributeValue(elt, 'hx-push-url') || responseInfo.etc.push
-    const replaceUrl = getClosestAttributeValue(elt, 'hx-replace-url') || responseInfo.etc.replace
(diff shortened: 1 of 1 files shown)


wooooodward:
@1cg just wondering if this can be merged - this will fix an issue I've had for a while.

PR #3410, opened 2025-08-16
Mention swag on the landing page

yawaramin:
Mention the swag store on the landing page above the sponsors section. Seems like the appropriate place to put it.


1cg:
thank you but want to think about this more

PR #180, opened 2020-09-01
afterOnLoad and afterRequest tests for outerHTML swapping

jreviews:
These events are not triggering at the moment.

PR #3425, opened 2025-09-04
Enable extension access to history functions

scriptogre:
## Description

This PR adds history management functions from htmx core:
1. `saveCurrentPageToHistory`
2. `setCurrentPathForHistory`
3. `pushUrlIntoHistory`
4. `replaceUrlInHistory`
 
to the `internalAPI` const, making them accessible via extensions using `api.pushUrlIntoHistory`, etc.

I'm currently working on a new extension that needs to reuse this history logic. I'm currently duplicating 200+ lines of code from core library just to be able to ensure consistent history behavior 😢 I would love to see this merged. 

## Testing
All tests pass.

## Checklist
  source changes)


scriptogre:
@1cg hey big boss, can you merge when available? thanks 🫡

MichaelWest22:
Hi, i have a PR up #3404 that adds history control to the htmx.ajax() api.  this change would add push and replace optional context inputs for the api so it can be used by javascript calls to perform the push and replace as this is a commonly asked for feature.  Just wondering if this might also help resolve your limitation with not being able to access history functions?  The only issue is this does need to do a full ajax request that could do a cached or full network request before it will trigger the htmx history update.  

If you needed to never do real ajax network requests but still push the url you could maybe use

```JS
htmx.ajax('GET', 'data:text/html,', {
  swap: 'none',
  push: '/your-new-url'  // or replace: '/your-new-url'
});
```
data url's are a bit different in that they don't do real network requests but you need to allow data: urls past the default selfRequestOnly config block by using htmx:validateUrl event listener i think.

Or you can just do a rubbish or likely to be cached real network request to allow the full htmx ajax flow.

What does your extension need to do?

scriptogre:
Hey @MichaelWest22. 

Thanks for your answer, and I'm really sorry for getting back so late!

I finally created a [PR #180](https://github.com/bigskysoftware/htmx-extensions/pull/180) in the htmx-extensions repo, with the new extension I was talking about. 

Here are the places where i'm using the internal API:
- https://github.com/bigskysoftware/htmx-extensions/blob/db110bb0c8bc3ede3c66da9e614fb800bc976729/src/server-commands/server-commands.js#L129
- https://github.com/bigskysoftware/htmx-extensions/blob/db110bb0c8bc3ede3c66da9e614fb800bc976729/src/server-commands/server-commands.js#L133
- https://github.com/bigskysoftware/htmx-extensions/blob/db110bb0c8bc3ede3c66da9e614fb800bc976729/src/server-commands/server-commands.js#L130
- https://github.com/bigskysoftware/htmx-extensions/blob/db110bb0c8bc3ede3c66da9e614fb800bc976729/src/server-commands/server-commands.js#L134

PR #3441, opened 2025-09-23
Wrap the bootstrap process in a function and expose it in the public api

maximebf:
## Description

I'm working on implementing streaming responses in my app similarly to React Suspense. The main body of the page is returned first and then blocks of content are streamed and replaced in the page.

This method works great in practice but HTMX is not compatible with it because it waits on the DOMContentLoaded event to initialize itself. However, using this technique, the event only occurs once the request is finished. This is a problem as the user has a fully rendered page in front of his eyes, than can already be interacted with but HTMX is still not loaded.

My propose change exposes an htmx.bootstrap() function that allows to manually bootstrap htmx at a choosen time. The function can be safely called multiple times but it will only execute once. It will automatically trigger on DOMContentLoaded like it does today.

The change is minimal but allows greater control, needed for the streaming use case.

## Testing

Tested manually + ran "npm run test"
Minimal change.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -320,6 +320,7 @@ var htmx = (function() {
   htmx.logNone = logNone
   htmx.parseInterval = parseInterval
   htmx._ = internalEval
+  htmx.bootstrap = bootstrap
 
   const internalAPI = {
     addTriggerHandler,
@@ -5108,8 +5109,13 @@ var htmx = (function() {
     }
   }
 
-  // initialize the document
-  ready(function() {
+  let isBootstrapped = false
+  function bootstrap() {
+    if (isBootstrapped) {
+      return
+    }
+    isBootstrapped = true
+    // initialize the document
     mergeMetaConfig()
     insertIndicatorStyles()
     let body = getDocument().body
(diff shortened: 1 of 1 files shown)


MichaelWest22:
I think this could be a great change to resolve the loading issue htmx has always had with edge cases.  see #2264 and the many linked issues and PR's we have tried in the past!  The last proposal put forward was to implement a initialize() function so it could be good to rename your implementation to that maybe if it makes sense.  don't think .init() would be good as it is already used in extensions etc.

We have to make sure we don't cause any regressions or issues for existing users but I think your proposed change should be fully backwards compatible. 

The biggest issue is how to test and validate this solution is working?  are you able to produce the most minimal html example that can show it is working as expected we can add to our manual tests?

https://github.com/bigskysoftware/htmx/compare/master...MichaelWest22:initialize

Here is a quick stab at what other things might be needed you can borrow the good bits from.

1cg:
@maximebf I love this change!   However, I would prefer the method be called `initialize()` (like in @MichaelWest22's change) and I would also like to see a docs update as well as a manual test if at all possible to ensure that the functionality works.  I know this is almost impossible to test in our automated suite.  Thank you!

MichaelWest22:
have implemented this feature now in htmx4 with a new htmx.initialize() public function that can be called to process the current body and init the history functions just like DomContentLoaded event does.  I use this in my hx-streaming extension to allow full htmx streaming from first page load so you can activate the page early while still streaming in partial response chunks later.  So closing this old v2 PR for now

Issue #3437, opened 2025-09-21
`hx-disabled-elt` on already disabled element


StabbarN:
`hx-disabled-elt` won't keep the `disabled` attribute after server has responded. It's problematic when the attribute hasn't been swapped.

Failing test:
```javascript
  it('preserves pre-disabled elements with hx-disabled-elt on other element', function() {
    this.server.respondWith('GET', '/test', 'ok')
    const btn = make('<button id="b" hx-get="/test" hx-disabled-elt="#a1">Click Me!</button>')
    const a1 = make('<a disabled id="a1"></a>')
    a1.hasAttribute('disabled').should.equal(true)
    btn.click()
    a1.hasAttribute('disabled').should.equal(true)
    this.server.respond()
    a1.hasAttribute('disabled').should.equal(true) // Fails. The `disabled` attribute has been removed.
  })
```
Copy the test into `test/attributes/hx-disabled-elt.js`.

A workaround for now is to add `:not(:disabled)` to `hx-disabled-elt`'s selector.

MichaelWest22:
Yeah currently it is designed to add and then remove the disabled elt from the target.  But it looks like it has not been designed to be used on elements that are already disabled.  Why would you need to disable an element if it is already disabled is the obvious question to ask?  The intended use is for elements that are enabled that need to be disabled just for the duration of the request so they can not be activated accidently during the request.  And as you have found in the rare edge case that you have something outside of htmx dynamically disabling things it should be possible to work around this limitation by either using advanced css selectors or by implementing the disable elt function yourself with a little manual JS on and htmx event listener.

StabbarN:
>  Why would you need to disable an element if it is already disabled is the obvious question to ask?

I don't, but it's easy to forget to add `:not(:disabled)`. I used `hx-disabled-elt` on all buttons and the already disabled ones slid along.

A side behaviour of `hx-disabled-elt` is that the already disabled elements will also become non-disabled after the request. IMO we need to clarify that.

Two suggestions:
1) Add a note to the documentation of `hx-disabled-elt` that says something like `Pre-disabled elements will not continue to be disabled after request finish.`

2) Exclude all elements that already are disabled. As you said, why would you want to apply `hx-disabled-elt` on an already disabled element?

PR #3443, opened 2025-09-24
Preserves pre-disabled elements with hx-disabled-elt on other element

StabbarN:
## Description
Elements with `hx-disabled-elt` that were already disabled before sending request will now keep the disabled attribute after server has responded or timed out. 

A good chunk of the users are probably thinking that this doesn't matter because their html will always swap after request has completed but if they have set `htmx.config.timeout` the request can time out and elements that are supposed to be disabled will not be disabled anymore.

The attribute `data-disabled-by-htmx` isn't set if the element already was disabled.

### Questions
Should I add a note to the documentation https://htmx.org/attributes/hx-disabled-elt/? Something like "Already disabled elements will not be affected"?

Corresponding issue: https://github.com/bigskysoftware/htmx/issues/3437

## Testing
Test added: `preserve pre-disabled elements`.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/test/attributes/hx-disabled-elt.js
+++ b/test/attributes/hx-disabled-elt.js
@@ -39,6 +39,18 @@ describe('hx-disabled-elt attribute', function() {
     fieldset.hasAttribute('disabled').should.equal(false)
   })
 
+  it('preserve pre-disabled elements', function() {
+    this.server.respondWith('GET', '/test', 'ok')
+    const b1 = make('<button hx-get="/test" hx-disabled-elt="#b2">Click Me!</button>')
+    const b2 = make('<button id="b2" disabled></button>')
+    b2.hasAttribute('disabled').should.equal(true)
+    b2.hasAttribute('data-disabled-by-htmx').should.equal(false)
+    b1.click()
+    b2.hasAttribute('disabled').should.equal(true)
+    this.server.respond()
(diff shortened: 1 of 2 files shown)


test/attributes/hx-disabled-elt.js
```diff
@@ -39,6 +39,18 @@ describe('hx-disabled-elt attribute', function() {
...
   })
 
+  it('preserve pre-disabled elements', function() {
+    this.server.respondWith('GET', '/test', 'ok')
+    const b1 = make('<button hx-get="/test" hx-disabled-elt="#b2">Click Me!</button>')
+    const b2 = make('<button id="b2" disabled></button>')
+    b2.hasAttribute('disabled').should.equal(true)
+    b2.hasAttribute('data-disabled-by-htmx').should.equal(false)
+    b1.click()
+    b2.hasAttribute('disabled').should.equal(true)
```
StabbarN, line 49: Suggestion: verify that `data-disabled-by-htmx` isn't set?  `b2.hasAttribute('data-disabled-by-htmx').should.equal(false)`. However, `data-disabled-by-htmx` isn't tested anywhere in this file.

1cg:
Great fix, thank you for contributing to htmx!

PR #3488, opened 2025-11-04
Allow build commands to run under Windows

scrhartley:
## Description
Allow running the build commands on Windows.

On Windows, even under WSL, it can't find `sed` from within the `package.json` scripts.
In addition, Windows doesn't have glob expansion, so instead it needs to get `brotli-cli` to do it.

Due to line ending differences, the brotli versions of the non-minified JS files don't match exactly when generated on Windows. (Minification seems to auto-convert Windows line endings to Linux ones, but the brotlification doesn't.) I tried playing about with Git auto eol conversation config, but I didn't manage to win that battle.

## Testing
I wiped out the contents of the dist folder, ran `npm run build` and then looked at the git diffs to see what changed.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,10 +14,10 @@
   ],
   "scripts": {
     "build": "npm run build:iife && npm run build:esm && npm run build:minify && npm run build:compress",
-    "build:iife": "sed 's/__/#/g' src/htmx.js > dist/htmx.js",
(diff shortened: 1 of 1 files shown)


scrhartley:
**Update:** it seems that the reason that it wasn't working in WSL is due to Windows interop, meaning it was picking up the Windows version of npm. After running `sudo apt install npm` the correct npm was used.

scrhartley:
With WSL, for a Windows checkout, in addition to the issue with line endings for brotli non-minified JS files, there is also a problem where the esm js will use Linux line endings for `export default htmx` even though rest of the file is using Windows line endings.

PR #3515, opened 2025-11-11
add relaxed JSON syntax support

scriptogre:
Introduces `htmx.parseJSON()` method with optional relaxed syntax that accepts unquoted object keys and auto-wraps bare properties in braces, making inline attribute configurations significantly more readable.

Features:
- Unquoted keys: `{timeout: 5000}` instead of `{"timeout": 5000}`
- Auto-wrapping: `timeout: 5000` instead of `{timeout: 5000}`
- Works in `hx-config` attributes, `<meta>` tags, and `data-*` attributes (useful for `hx-live` state later)
- Global `config.relaxedJSON` option (defaults to true)
- Fully backwards compatible with strict JSON (whether `config.relaxedJSON` is `true` or `false`)
- 3-line regex-based implementation using `JSON.parse` (no `eval`)

Examples:
```html
  <button hx-config='timeout: 5000, cache: "no-cache"'>
  <meta name="htmx:config" content='defaultSwap: "innerHTML", timeout: 30000'>
```

Users can disable globally with `config.relaxedJSON = false` for strict JSON-only mode.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -112,11 +112,12 @@ var htmx = (() => {
                 },
                 morphIgnore: ["data-htmx-powered"],
                 noSwap: [204, 304],
-                implicitInheritance: false
+                implicitInheritance: false,
+                relaxedJSON: true
             }
             let metaConfig = document.querySelector('meta[name="htmx:config"]');
             if (metaConfig) {
-                let overrides = JSON.parse(metaConfig.content);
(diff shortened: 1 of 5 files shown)


MichaelWest22:
I think this can be closed now that we have parseConfig() function merged that I emplemented based on this original great idea.

1cg:
closing per @MichaelWest22 comment

Issue #2892, opened 2024-09-11
Rethink the trigger delay modifier semantic for multiple event targets


jsonn:
Similar to #2891, the delay modifier behaves suboptimal when multiple nodes can provide data. I think the desired behavior here is also to have one shared delay per trigger specification. I.e. going with the active search example again, if I have two form elements `name` and `city`, as long as I change either value within the delay, the trigger shouldn't be activated.

MichaelWest22:
This should now work in htmx 4 when using from: to multiple elements with a delay.  The delay etc all link back to the same handler now so it just works fine.  also note that you can normally get the same result by listening on a parent form or fieldset so that your multiple inputs all trigger the same elements delayed listener.

PR #3526, opened 2025-11-15
Rework multiple spec and multiple targets for delayed modifier

jsonn:
## Description

If multiple nodes are targetted by one trigger specification, they share a delay. This is useful e.g. for active searches with multiple participating fields, so that the search triggers only when search modifiers have been processed.

If a node has multiple triggers with the delay modifier, they act independently. This corresponds to the logic from #2026.

Corresponding issue: #2892 

This PR is the successor for #2893.

## Testing
New tests have been added as appropriate. 

## Checklist

  source changes)
  approved via an issue


1cg:
Hi @jsonn sorry this sat so long.  We reviewed this and don't think this functionality is worth the additional code.

PR #3589, opened 2025-12-14
Trims only the first leading space in a SSE field value (#3588)

### LINKED ISSUE #3588: HTMX4 SSE parser trims all leading whitespace instead of a single space
The SSE parser in htmx 4 currently removes all leading whitespace from field values. According to the SSE spec, only one leading space may be removed, and only if it exists.

Relevant code:
https://github.com/bigskysoftware/htmx/blob/4403f8eff6361ce3f6a7fc746a6a87c78143f52c/src/htmx.js#L719

Spec reference:
https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

> Collect the characters on the line after the first U+003A COLON character (:), and let value be that string. **If value starts with a U+0020 SPACE character, remove it from value**.

## Example
**Given this SSE payload:**

```
id: 45
event:  message
data:  Hello, World!
```

**Current behavior:**

```
id = "45"
event = "message"
data = "Hello, World!"
```

**Expected behavior (per spec):**

```
id = "45"
event = " message"
data = " Hello, World!"
```

antonlorani:
## Description
Fixes #3588 

## Testing
Added a new test case to `test/tests/end2end/sse.js`

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -716,7 +716,9 @@ var htmx = (() => {
                         if (colonIndex <= 0) continue;
 
                         let field = line.slice(0, colonIndex);
-                        let value = line.slice(colonIndex + 1).trimStart();
+
+                        let value = line.slice(colonIndex + 1)
+                        if (value[0] === ' ') value = value.slice(1);
 
                         if (field === 'data') {
                             message.data += (message.data ? '\n' : '') + value;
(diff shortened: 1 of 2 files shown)


1cg:
hi @antonlorani we are reviewing this change now, thank you!

PR #3604, opened 2025-12-26
Docs: HTML Dialog Example

klnusbaum:
This adds a Dialog example using the standard HTML dialog.

## Description
I noticed that while there were several examples around how to make modal dialogs, none of them actually used the `<dialog>` tag. I was curious if I could actually get htmx to work with the `<dialog>` tag and after some playing around, I was able to get something that works. I figured it might be good to share what I learned.

I'm not entirely sure this is a good idea. The more I learn about html dialogs, the more I'm skeptical about how useful they are. I'm curious what others think.

## Testing
Ran the docs locally and verified the intended behavior.

## Checklist

  source changes)
  approved via an issue


marcoklein:
Would be interesting to have that in the docs. Were you able to solve that client side ESC closing mechanism? As [the docs state](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) a modal can be closed via Escape and then HTMX looses control over the state as it's closed client side but not yet replaced.

In your example I could produce that behaviour by opening the modal and pressing Escape. I got it working for another use-case by returning `<dialog open ...></dialog>` because then it doesn't open in the modal state and does not have that default closing mechanism.

klnusbaum:
> Would be interesting to have that in the docs. Were you able to solve that client side ESC closing mechanism? As [the docs state](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) a modal can be closed via Escape and then HTMX looses control over the state as it's closed client side but not yet replaced.
> 
> In your example I could produce that behaviour by opening the modal and pressing Escape. I got it working for another use-case by returning `<dialog open ...></dialog>` because then it doesn't open in the modal state and does not have that default closing mechanism.

The HTML spec seems to strongly discourage the use of the `open` attribute, so I tried to stay away from it. I got something working by setting `closedby="none"`. This makes it so the __only__ way to close the dialog is by clicking out button.

It doesn't feel great.

The other thing I tried doing was setting an `hx-trigger` to listen for the escape key. But I couldn't get that to work.

Overall, given the number of caveats we seem to need for this example, I'm not sure I like this. Feels like I'm trying to ice-skate uphill.

klnusbaum:
After reading the HTML spec a little more closely, the objections it has to using the `open` attribute seemed like the didn't apply in this case. I gave it a shot and it produced some weirdness where the dialog wasn't centered in the page. It was buried down in the footer and I had to scroll to find it. I'm not sure what that's all about.

1cg:
we have this example in four, thank you!

PR #3663, opened 2026-02-20
Fix v4 - ctx.isSSE is already set from the response headers at line 522

roycamp:
The local `isSSE` variable at line 532 re-reads the same Content-Type header already captured in `ctx.isSSE` at line 522.

## Checklist

* [X] I have read the contribution guidelines
* [X] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [X] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [X] I ran the test suite locally (`npm run test`) and verified that it succeeded

PR #3665, opened 2026-02-20
Extract SSE into extension, align WS extension

scriptogre:
Extract SSE from core into a standalone extension. Align WS extension config and events to match.

## SSE

- Move SSE out of core into `hx-sse.js`
- Use Fetch + ReadableStream instead of EventSource (supports POST, custom headers, cookies)
- Two modes: one-off streams via `hx-get` returning `text/event-stream`, persistent connections via `hx-sse:connect`
- Supersedes #3663 — that code no longer exists in core

## WS events

- Make `before:ws:message` / `after:ws:message` fire for all messages, not just JSON
- Previously, raw (non-JSON) messages skipped these events and went through a separate `rawMessage` hook
- Now one pipeline handles everything, so `htmx:ws:rawMessage` and `htmx:wsMessage` become redundant
- Merge `before:ws:connect` + `ws:reconnect` into `before:ws:connection` (reconnects have `attempt > 0`)
- Rename `htmx:wsSendError` → `htmx:ws:error` (camelCase breaks `hx-on:`)

## WS config

- Align reconnection defaults with SSE: 500ms delay, 60s max, 0.3 jitter, infinite attempts
- Add `pauseOnBackground: true` disconnect on tab hide, reconnect on show
- Use `api.onTrigger()` for full trigger modifier support (`delay`, `throttle`, `once`, etc.)

## Docs

- Rename `closeOnHide` → `pauseOnBackground` everywhere
- Fix wrong defaults in WS docs and migration guide
- Remove `htmx.config.sse` from core config tables (extension-only now)


scriptogre:
Splitting this into two focused PRs: one for the SSE extraction, one for the WS alignment. Will link them here once they're up.

PR #3694, opened 2026-03-11
Submitting a form using GET: discrepancy between htmx and browser

ergor:
## Description

Given this form:

```
<form action="/test" method="get">
  <input type="hidden" name="action" value="A">
  <button id="btnB" type="submit" name="action" value="B">B</button>
</form>
```

When clicking the button,
The browser's behavior is to _append_ both values, resulting in a query string as "action=A&action=B".

But, when using HTMX, given this form:
```
<form hx-get="/test">
  <input type="hidden" name="action" value="A">
  <button id="btnB" type="submit" name="action" value="B">B</button>
</form>
```
When clicking the button,
HTMX issues a GET request, where the button's value _overrides_ any other form data with the same `name`; the resulting query string is just "action=B". This is _unexpected_ behavior.

This patch tries to bring HTMX in line with default browser behavior.

Now, I think this solution is rather hacky. Why is there differentiation between GET and non-GET requests in `getInputValues()` in the first place? I don't know the intricacies of HTMX, explanations are appreciated. I am open to iterate on this PR if the solution is below quality expectation.

Corresponding issue: 
#1541 This issue brought it up, but was then closed due to #1559. However, that PR didn't fix it for the case of `GET`.

## Testing
I tested it using the test suite. I added relevant test-cases that triggered the bug, and then verified the fix by running the tests with `npm run test`

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -3630,7 +3630,8 @@ var htmx = (function() {
     (elt.tagName === 'INPUT' && getRawAttribute(elt, 'type') === 'submit')) {
       const button = internalData.lastButtonClicked || (/** @type HTMLInputElement|HTMLButtonElement */(elt))
       const name = getRawAttribute(button, 'name')
-      addValueToFormData(name, button.value, priorityFormData)
+      const selectedFormData = verb === 'get' ? formData : priorityFormData
+      addValueToFormData(name, button.value, selectedFormData)
     }
 
     // include any explicit includes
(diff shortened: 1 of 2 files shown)


1cg:
Hey @ergor this is a good fix, but I'm a little worried it will break things for people.  Would you be OK porting the fix to 4.x (or at least the tests, it may already be fixed)?

ergor:
Hi, thanks for reviewing.

I tested with v4 beta3 (pure, no modifications), and the issue is indeed fixed there. Closing the PR, as it is obsolete. 

Screenshot_2026-05-16_10-43-10

PR #3699, opened 2026-03-13
redo etag support to just prevent swap on etag match and leave caching to browser

MichaelWest22:
# Refactor ETag Implementation to Passive Comparison

## Summary
Refactored htmx's ETag support from manual `If-None-Match` header injection to **passive ETag comparison** for DOM optimization.

## What Changed
- **Removed**: Manual `If-None-Match` header injection (lines 370-377)
- **Removed**: 304 from `config.noSwap` array (now only `[204]`)
- **Added**: `__handleETagComparison()` method for passive ETag checking
- **Added**: Initial ETag support in `__initializeElement()` via `hx-config`
- **Added**: `htmx:etag:match` event (preventable to override abort behavior)
- **Updated**: Documentation in `docs.md` and `events.md`

## Why This Design?
**Problem with manual `If-None-Match`**: When application code sends `If-None-Match`, browsers pass through raw 304 responses instead of reconstructing 200 responses from cache. This breaks htmx's swap logic.

**Passive comparison benefits**:
1. **Browser handles network optimization** - Automatically sends `If-None-Match` and handles 304s per HTTP spec
2. **htmx handles DOM optimization** - Compares response ETags with stored values to prevent unnecessary swaps
3. **Layered approach** - Network caching (browser/CDN) + DOM optimization (htmx)

## Usage
```html
<!-- Opt-in via hx-config -->
<div hx-get="/news" hx-trigger="every 3s" hx-config='etag:true'>

<!-- With initial ETag to prevent first swap -->
<div hx-get="/news" hx-config='etag:"abc123"'>
```

## Behavior
- ETags stored on source element (`sourceElement._htmx.etag`)
- When response ETag matches stored ETag → abort all swaps (main + OOB + partials)
- Fires `htmx:etag:match` event (can be prevented to force swap)
- Opt-in only (no performance impact on elements without `etag` config)

## Testing
- Complete unit test suite in `test/tests/unit/_htmx.etag.js`
- Tests cover: opt-in behavior, ETag matching/changing, initial ETag support, event firing, and event prevention

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -118,7 +118,7 @@ var htmx = (() => {
                 extensions: '',
                 morphIgnore: ["data-htmx-powered"],
                 morphScanLimit: 10,
-                noSwap: [204, 304],
+                noSwap: [204],
                 implicitInheritance: false,
                 defaultSettleDelay: 1
             }
@@ -375,9 +375,6 @@ var htmx = (() => {
                     this.__htmxProp(sourceElement).etag ||= ctx.request.etag
                 }
             }
-            if (sourceElement._htmx?.etag) {
-                ctx.request.headers["If-none-match"] = sourceElement._htmx.etag
-            }
             return ctx;
         }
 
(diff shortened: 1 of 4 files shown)


1cg:
Will consider etags more in next release, closing for now. Thanks for the work on this!

PR #3739, opened 2026-04-11
simplify outerHTML target fallback

MichaelWest22:
## Description
I think we can simplify the fallback code for when outerHTML removes the target to just newContent[0].  I was worried this could be a text element which could break things so was thinking of doing a filter to return first Element in newContent but turns out if it is a text node it just works fine and triggers events fine and class changes on target are guarded already so just no-op .

The only thing we lose with this simplification is possibly scroll To will possibly go to the parent node when you use outerHTML to blank the target with "".  But I can't see a realistic situation where you would want to delete with outerHTML swap and also want to scroll to this exact point.

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1329,11 +1329,10 @@ var htmx = (() => {
                     target.textContent = fragment.textContent;
                 } else if (swapStyle === 'outerHTML') {
                     if (parentNode) {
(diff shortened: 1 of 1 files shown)


1cg (approved):
Thank you!

PR #3740, opened 2026-04-11
handle disabling hx-location push better so you can use replace easier.

MichaelWest22:
## Description
* If you need to use hx-location with the optional replace option then you also need to set push to false to remove the default push=true that hx-location applies.  But right now it checks if push is set to something falsey before setting push to 'true' and this means the only way to block push and allow replace option to work is to use javascript format and set push to "false" so it is a truthy string.  you can't use boolean false or the non JSON parseConfig format which always uses boolean false here.  To fix this we can simplify the code with ??= 'true' so it will set push to true only when it is null or undefined.

Corresponding issue:

## Testing
no automatic testing for hx-location yet as it breaks the automated tests

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -603,7 +603,7 @@ var htmx = (() => {
                     path = opts.path;
                     delete opts.path;
                 }
-                opts.push = opts.push || 'true';
+                opts.push ??= 'true';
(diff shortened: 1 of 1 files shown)


1cg (approved):
Thank you!

PR #3741, opened 2026-04-11
fix hx-location to handle replace

MichaelWest22:
## Description
With HX-Location you can disable the default push = 'true' it does with false but there is currently no way to enable the replace option I added documentation for earlier.  While you can set push to 'false' this is still truthy and this blocks it picking up replace option.  You can try using the boolean false but this just gets overwritten by 'true' right now.  

So to fix this just need to use ?? instead of || so boolean false will prevent the 'true'.  And also improve handling of 'false' by setting it to null so it won't block replace applying.

Corresponding issue:

## Testing
Added a couple of tests for setting push to false and 'false' and setting replace.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4696,8 +4696,10 @@ var htmx = (function() {
     const requestPath = responseInfo.pathInfo.finalRequestPath
     const responsePath = responseInfo.pathInfo.responsePath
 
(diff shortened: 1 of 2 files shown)


1cg:
Thank you!

Issue #3745, opened 2026-04-13
Is the addition of a dependency a bug or a feature in v.4


FatherShawn:
I'm guessing this is a bug given the previous "no other dependencies" philosophy and that `playwright` should be in the devDependencies?

https://github.com/bigskysoftware/htmx/blob/b9b368ed4552a864394c1528d04d5788812f4565/package.json#L75

Or is there an architectural change and this is a feature?

CRSylar:
Hi, i've opened the PR that moves it into the devDependecies, fell free to close it if you'll find that the dependecies was indeed right

FatherShawn:
Thanks @CRSylar !  I was coming to do that!  I had an exchange in Drupal slack with @MichaelWest22 overnight in which he affirmed  that this was unintentional.

MichaelWest22:
removed this dependency from the next release as it is not a direct dependency and only used by web test runner indirectly as a dev dependency.  was likely included by mistake when trying to fix package lock versions or something.  package.json already has a exports fix for extensions we are shipping in the next release shortly

PR #3746, opened 2026-04-14
fix: moved playwright from dependencies to devDependencies

CRSylar:
[Issue #3745](https://github.com/bigskysoftware/htmx/issues/3745)

## Description
moved playwright from dependencies to devDependencies

Corresponding issue:

## Checklist

  source changes)
  approved via an issue


MichaelWest22:
Yeah thanks this looks like just a mistake and we don't use it in dev dependency so just need to remove it so web test runner can import it if needed.  I've removed it properly now

PR #3769, opened 2026-04-24
docs: clarify next/previous selectors scan whole DOM, not just siblings (#3430)

### LINKED ISSUE #3430: `previous <CSSselector>` and `next <CSSselector>` find more than just siblings. Improve docs?
Hi,

`hx-target="previous"` works like `Element.previousElementSibling()`. If the element is its parent’s sole child, no target will be found.

However, `hx-target="previous <CSS selector>"` will apparently find anything on any hierarchy level, as long as it begins and ends before the source element.

Here is a fiddle showing what I mean: https://jsfiddle.net/cuokm8gz/2/

I like this, but the docs state “`previous .error` will target the closest previous *sibling* with `error` class”.

I propose that `previous <CSS selector>` should either be restricted to siblings (😩) or the docs should be changed to “element” instead of “sibling”. If the latter is chosen, additional [tests for next and previous](/bigskysoftware/htmx/blob/master/test/attributes/hx-target.js) should be added to prevent a regression on this. I will gladly start a PR for this.

Thanks!

manwithacat:
Fixes #3430.

The parenthetical examples for the `next <CSS selector>` and `previous <CSS selector>` extended-target syntax described the result as the closest *sibling* with the matching class. The surrounding sentence already correctly describes the actual behavior (a forward/backward DOM scan that can land on elements at any hierarchy level), but the examples in parentheses contradicted it. As @MichaelWest22 noted in the issue, this is "a copy paste error in the documentation from when that feature was added."

This PR replaces "sibling" with "element" in those examples and adds a brief clarifying clause, in 10 places across 5 docs:

- `www/content/attributes/hx-target.md` (next + previous)
- `www/content/attributes/hx-trigger.md` (next + previous)
- `www/content/attributes/hx-include.md` (next + previous)
- `www/content/attributes/hx-disabled-elt.md` (next + previous)
- `www/content/extensions/response-targets.md` (next + previous)

Docs-only — targeting `master` per CONTRIBUTING.md.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/attributes/hx-disabled-elt.md
+++ b/www/content/attributes/hx-disabled-elt.md
@@ -16,10 +16,10 @@ added to them for the duration of the request. The value of this attribute can b
 * `find <CSS selector>` which will find the first child descendant element that matches the given CSS selector
 * `next` which resolves to [element.nextElementSibling](https://developer.mozilla.org/docs/Web/API/Element/nextElementSibling)
 * `next <CSS selector>` which will scan the DOM forward for the first element that matches the given CSS selector
-  (e.g. `next button` will disable the closest following sibling `button` element)
+  (e.g. `next button` will disable the closest following `button` element — note this scans the whole DOM forward, not just siblings)
 * `previous` which resolves to [element.previousElementSibling](https://developer.mozilla.org/docs/Web/API/Element/previousElementSibling)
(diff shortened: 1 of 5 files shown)


1cg:
already fixed, thank you!

PR #3770, opened 2026-04-24
docs: clarify afterRequest successful/failed reflect responseHandling rule (#3440)

### LINKED ISSUE #3440: event.detail.successful always true with hx-on:after-request when allowing all responses status to swap
### The Problem:
allowing all responses status to swap:
`<meta name="htmx-config" content='{"responseHandling": [{"code":"...", "swap": true}]}'>`
Listen to hx-on:after-request on an element:
`<form hx-on::after-request='if(event.detail.successful) { doSome() }>`
event.detail.successful is always true, even for error (not 2xx status) responses

### From the [::afterRequest Doc](https://htmx.org/events/#htmx:afterRequest):

> detail.successful - true if the response has a 20x status code _or is marked detail.isError = false in the htmx:beforeSwap event_, else false

Doesn't mention that allowing all responses status to swap will also mark the response as successful?

manwithacat:
Fixes #3440.

The `htmx:afterRequest` event docs described `detail.successful` and `detail.failed` purely as "true if 20x status code", but the actual code (at `src/htmx.js:4849, 4893–4894`) derives them from the matched `htmx.config.responseHandling` rule's `error` flag:

```js
const shouldSwap = responseHandling.swap
let isError = !!responseHandling.error
// ... beforeSwap can override isError ...
responseInfo.failed = isError
responseInfo.successful = !isError
```

When users customize `responseHandling` to allow non-2xx responses to swap — for example with `{"code":"...", "swap": true}` and no `error: true` — those responses are *also* marked `successful` in `htmx:afterRequest`. The issue reporter was surprised by this. As @MichaelWest22 confirmed in the issue, the behavior is intentional ("If you override the default of what is defined as successful to swap then you would expect this to flow though"); the docs were just incomplete.

This PR rewords the descriptions of both `detail.successful` and `detail.failed` to:

1. Reference `htmx.config.responseHandling` as the source of truth.
2. Keep the "20x = successful" framing as a parenthetical "by default" note so existing readers' mental model is preserved.
3. Add an explicit note about the custom `responseHandling` case so future users with the same setup don't hit the same surprise.

Docs-only — targeting `master` per CONTRIBUTING.md. No code change.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/www/content/events.md
+++ b/www/content/events.md
@@ -48,10 +48,14 @@ can be paired with [`htmx:beforeRequest`](#htmx:beforeRequest) to wrap behavior 
 * `detail.xhr` - the `XMLHttpRequest`
 * `detail.target` - the target of the request
 * `detail.requestConfig` - the configuration of the AJAX request
-* `detail.successful` - true if the response has a 20x status code or is marked `detail.isError = false` in the
-  `htmx:beforeSwap` event, else false
-* `detail.failed` - true if the response does not have a 20x status code or is marked `detail.isError = true` in the
-  `htmx:beforeSwap` event, else false
+* `detail.successful` - true if the response is not marked as an error by the matched
+  [`htmx.config.responseHandling`](@/docs.md#response-handling) rule (by default, a 20x status code) or is marked
+  `detail.isError = false` in the `htmx:beforeSwap` event, else false. Note that if you customize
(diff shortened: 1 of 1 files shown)


1cg:
Thank you!

PR #3772, opened 2026-04-24
fix: prevent partial response-header match TypeError (#527)

### LINKED ISSUE #527: partial response header match can cause a TypeError
The regular expressions passed into `hasHeader()` match headers that may have additional characters in front (e.g. `X-HX-Trigger`). When the header value is later accessed (e.g. in `handleTrigger()`) it will be null.

Example: given a response with header `X-HX-Trigger: foo` the following will throw a `TypeError`.

```
if (hasHeader(xhr, /HX-Trigger:/i)) {
    handleTrigger(xhr, "HX-Trigger", elt);
}

function handleTrigger(xhr, header, elt) {
    var triggerBody = xhr.getResponseHeader(header);
    if (triggerBody.indexOf("{") === 0) {  // TypeError: triggerBody is null
```

I don't know if there are any implications in switching to match the header names exactly, I think I saw in an old issue that the headers used to be in the `X-HX-*` format which is how I stumbled across this.

I have an example fix on a fork here: https://github.com/frontenac/htmx/commit/8c81f11d8fcb50e3126a20d4396cd5fa321bfa23. I've changed the implementation of `hasHeader()` to do a null check on the return value from `xhr.getResponseHeader()` instead of regexp matching. Or, a regexp implementation pattern could see the pattern turned into a multi-line one to match the header that way, e.g. `/^HX-Trigger/im`

If my fix is acceptable (and it's something that needs fixing) I think I can open a pull request for this with the fork above, I'm new to GitHub so I'm still learning the process.

Thanks!

manwithacat:
Fixes #527 — a four-year-old bug that @1cg (thread comment) explicitly invited a fix for.

## The bug

`hasHeader(xhr, regex)` ran `regex.test(xhr.getAllResponseHeaders())` — testing against the concatenated header string. When a response carried a header whose name *contained* an htmx header name as a substring — the issue reporter's example was `X-HX-Trigger: foo` — the regex `/HX-Trigger:/i` matched the substring and returned `true`, but `xhr.getResponseHeader('HX-Trigger')` returned `null` (because the actual header name is `X-HX-Trigger`). `handleTriggerHeader` then crashed at `src/htmx.js:2068` on `triggerBody.indexOf('{')`.

## The fix

Refactor `hasHeader` to use `getResponseHeader(name) !== null` directly:

```diff
- function hasHeader(xhr, regexp) {
-   return regexp.test(xhr.getAllResponseHeaders())
- }
+ function hasHeader(xhr, name) {
+   return xhr.getResponseHeader(name) !== null
+ }
```

Every one of the 12 call sites already knew the exact header name (they used it in both the regex *and* the paired `getResponseHeader` call), so the update is mechanical:

```diff
- if (hasHeader(xhr, /HX-Trigger:/i)) {
+ if (hasHeader(xhr, 'HX-Trigger')) {
```

The substring-match class of bug is eliminated — `hasHeader` and `getResponseHeader` now use identical header lookup semantics.

## Tests

New regression test in `test/core/headers.js`:

- `does not crash or fire trigger when X-HX-Trigger header is present without HX-Trigger` — sends `X-HX-Trigger: foo`, asserts the `foo` listener is not fired. Before the fix this crashed with `TypeError: Cannot read properties of null (reading 'indexOf')`.

Full suite: **847 passed, 0 failed, 3 skipped**, 99.9% coverage. ESLint clean.

## A note on @1cg's original concern

The thread mentions Chrome "issuing pretty gnarly looking error messages when you ask for a header that doesn't exist." That was the original reason for the regex approach. Modern Chrome/Firefox/Safari all silently return `null` from `getResponseHeader()` for missing headers with no console noise, so that justification no longer applies.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -4039,11 +4039,11 @@ var htmx = (function() {
 
   /**
    * @param {XMLHttpRequest} xhr
-   * @param {RegExp} regexp
+   * @param {string} name
    * @return {boolean}
    */
-  function hasHeader(xhr, regexp) {
-    return regexp.test(xhr.getAllResponseHeaders())
+  function hasHeader(xhr, name) {
+    return xhr.getResponseHeader(name) !== null
   }
 
   /**
@@ -4668,13 +4668,13 @@ var htmx = (function() {
     //= ==========================================
     let pathFromHeaders = null
     let typeFromHeaders = null
-    if (hasHeader(xhr, /HX-Push:/i)) {
+    if (hasHeader(xhr, 'HX-Push')) {
       pathFromHeaders = xhr.getResponseHeader('HX-Push')
       typeFromHeaders = 'push'
-    } else if (hasHeader(xhr, /HX-Push-Url:/i)) {
+    } else if (hasHeader(xhr, 'HX-Push-Url')) {
       pathFromHeaders = xhr.getResponseHeader('HX-Push-Url')
       typeFromHeaders = 'push'
-    } else if (hasHeader(xhr, /HX-Replace-Url:/i)) {
+    } else if (hasHeader(xhr, 'HX-Replace-Url')) {
       pathFromHeaders = xhr.getResponseHeader('HX-Replace-Url')
       typeFromHeaders = 'replace'
     }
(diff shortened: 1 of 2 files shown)


1cg (approved):
Thank you!

PR #3776, opened 2026-04-28
Real outerHTML swap on body; use it for history restore

1cg:
## Summary
- `__makeFragment` now preserves `<body>` (with attributes) for full-page responses, alongside the existing fragment.
- `outerHTML` swap into `document.body` does a real outer swap: body attributes sync from the response via `__copyAttributes`, children are replaced. Previously this was coerced to `innerHTML` semantics, dropping body attrs.
- `__restoreHistory` now uses `outerHTML` for both the `[hx-history-elt]` partial and the body fallback. Drops the `strip:true` workaround, unifies on a single swap style, and `<body>` class/data-attrs now persist across history restore.

Follow-up to #3773.

## Test plan
- [x] `__makeFragment` unit tests: body element captured (with attributes) for full-page responses; undefined for partial responses.
- [x] Integration test: `htmx.swap` `outerHTML` into `document.body` syncs class + data-attrs and removes attrs not in the response (body state saved/restored around the test).
- [x] Integration test: history restore with `[hx-history-elt]` swaps only the marked element, leaves siblings intact, and does not leak response content outside `[hx-history-elt]` into the live document.
- [x] Full suite: 1144 passed, 0 failed, 5 skipped.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1027,13 +1027,14 @@ var htmx = (() => {
             response = response.replace(/<head(\s[^>]*)?>[\s\S]*?<\/head>/i, m => (title = this.__parseHTML(m).title, ''));
             let startTag = response.match(/<([a-z][^\/>\x20\t\r\n\f]*)/i)?.[1]?.toLowerCase();
 
-            let doc, fragment;
+            let doc, fragment, body;
             if (startTag === 'html' || startTag === 'body') {
                 doc = this.__parseHTML(response);
                 fragment = document.createDocumentFragment();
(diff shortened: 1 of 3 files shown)


MichaelWest22:
yeah body replacement has allways been a bit of a gap.  fixing this would be good.  created an alternative PR #3778 where we could simplify this into a new custom swap style just for our body/history use case.

1cg:
Superceded by #3778, which generalises this into a proper `outerSync` swap style instead of overloading `outerHTML` on body. Thanks Michael!

PR #3783, opened 2026-05-05
optimize and minimize hx-on code

MichaelWest22:
## Description
The new hx-on modifier code has been added but it could do with some code optimization to minimize the final bundle size.

metadata character code can be simplified
using map to create a mod object with true properties per mods string allows us to remove the has helper and use it as a object access much simpler.  Also this means we can safely just pass this mod object to the event listener to set all capture, passive, once and it will ignore the rest.
Also removed the removeEventListener since it is not needed once we pass once option.

Finally put back 
```
                ctx.vals = obj; // make available for json extensions
```
          
Which was recently removed accidently which is still used by json extensions to access the raw hx-vals object data with json types preserved.      

Corresponding issue:

## Testing
No real functional changes and tests still all pass

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1690,6 +1690,7 @@ var htmx = (() => {
                         if (typeof e !== 'symbol') console.error(e);
                     }
                 };
+                // mod object passes capture, passive, and once to event listener
                 target.addEventListener(evtName, handler, mod);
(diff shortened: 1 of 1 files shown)


1cg:
Thanks Michael!

Issue #3781, opened 2026-04-30
[4.0.0-alpha8] BUG: hx-target not working on Unicode id (e.g. id="foo-中文")


YamiOdymel:
An HTML `id` attribute can contain Unicode characters, but HTMX **does nothing** if `id` contains Unicode characters (or 😭emoji).

```
<div id="container-中文">
    <form hx-post="/action/follow_topic" hx-target="#container-中文" hx-swap="outerHTML">
        <button type="submit">Follow Topic 🎉</button>
    </form>
</div>
```

MichaelWest22:
I thought it could be that since this is not an id but a standard CSS selector that maybe it would need some css escaping to make it valid.  But actually as long as it is just standard utf-8 unicode it just works as a css selector like this fine.  Testing your posted html snipit on alpha8 build and it works fine for me.  You might have a unicode encoding issue on your backend solution or template files maybe?  Try generating a straight html file to test this.  

Also note that most of the time you use an `id` for htmx target selection you can actually use any of the great relative selectors htmx has built in like closest:

```html
<div id="container-中文">
    <form hx-post="/action/follow_topic" hx-target="closest div" hx-swap="outerHTML">
        <button type="submit">Follow Topic 🎉</button>
    </form>
</div>
```

YamiOdymel:
@MichaelWest22 Hmm, weird. I tried using a brand-new test `.html` file with the Alpha 8 build, but the problem still exists (even _with_ the `closest` selector). HTMX does nothing when there is Unicode in the `id`, and it also doesn’t output any error message.

Demo video:

https://github.com/user-attachments/assets/1cd6637e-540c-49b3-9d23-3ca323f5a3a3

Thanks for the note, by the way. I love seeing extra information every time you reply.

MichaelWest22:
Strange I can't reproduce this at all using the same html you have there.  The error causing the problem will likely be showing in your console where any exceptions may be posted.  Can you try also enabling full htm loggin and then click the button.  in console just paste `htmx.config.logAll=true` and then click the button to view all the htmx event logs which may give more error details

YamiOdymel:
@MichaelWest22 Great, I saw the error message now. Chinese characters and Emoji causes the same error. I'm using Traditional Chinese Windows 11 / Chrome 147

```
$ htmx.config.logAll=true
> true

> htmx:config:request
> htmx:before:request
> htmx:error 
[...]

PR #3784, opened 2026-05-05
encode unicode id

MichaelWest22:
## Description
We have to encode id's containing unicode before we send in target/source headers as otherwise these will be rejected as invalid

Also added error reporting to make it much easier to debug htmx errors in future.  It outputs a warning to console logs with all the error detail and ctx object data.

Corresponding issue:
#3781

## Testing
Added a basic test for unicode but fetch mock does not throw the error making it hard to fully test e2e

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -394,7 +394,7 @@ var htmx = (() => {
         }
 
         __buildIdentifier(elt) {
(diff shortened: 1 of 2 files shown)


1cg:
Thanks Michael!

PR #3785, opened 2026-05-05
add nonce reuse protection to hx-nonce

MichaelWest22:
## Description
The extension rewrites the response nonce to the page nonce so swapped-in elements pass subsequent nonce checks. Before doing that rewrite, we can scrub any element that already carries the page nonce value from the raw response text.

The server cannot know the page nonce - it only knows its own per-response nonce. So if the page nonce appears in a response, it was put there by an attacker, not the server. Scrubbing it first means the rewrite pass cannot accidentally promote attacker-controlled elements to trusted status.

The risk: unlike `<script nonce>`, `hx-nonce` attributes are not blanked by browsers after parse, so they are a possible additional nonce exposure surface. The scrub step is a defense-in-depth measure to ensure a stolen nonce cannot be pre-stamped into injected content to pass nonce checks. 

Corresponding issue:

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/hx-nonce.js
+++ b/src/ext/hx-nonce.js
@@ -38,8 +38,7 @@
     function checkNonce(elt) {
         if (!pageNonce) return false;
         let eltNonce = getNonce(elt);
-        if (eltNonce === pageNonce) return;
-        if (stripHxAttributes(elt, eltNonce)) return false;
+        if (eltNonce !== pageNonce && stripHxAttributes(elt, eltNonce)) return false;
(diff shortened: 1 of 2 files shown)


1cg:
Thanks!

1cg:
Landed manually on four-dev (commit 094c83fd) — the doc file path changed under you in #3786 so the auto-merge couldn't resolve, but the doc rename was clean and your authorship is preserved.

PR #3788, opened 2026-05-06
restrict fetch mode changes from hx-config

MichaelWest22:
## Description
Overriding the fetch same-site mode to cors via hx-config attribute has a potential to be a possible security issue so we should probably block overriding this with htmx attributes and only use the global htmx config value we can trust instead.  This way users have to opt in the whole page to allow cors requests and understand the security implications of doing this.  Add documentation on how to correctly configure CSP to block or allow other api domains as this should be the prefered way to implement security protections rather than managing allowed origins manually inside htmx.

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue


1cg:
Thanks Michael!

PR #3792, opened 2026-05-09
fix: encode Unicode IDs in HX-Trigger and HX-Target headers

### LINKED ISSUE #3781: [4.0.0-alpha8] BUG: hx-target not working on Unicode id (e.g. id="foo-中文")

xxiaoxiong:
## Summary

Fixes #3781

This PR fixes a bug where htmx 4.0 throws a `TypeError` when an element has an ID containing Unicode characters (e.g., Chinese characters, emoji).

## Problem

When using Unicode characters in element IDs:

```html
<div id="container-中文">
    <form hx-post="/action/follow_topic" hx-target="#container-中文" hx-swap="outerHTML">
        <button type="submit">Follow Topic 🎉</button>
    </form>
</div>
```

Clicking the button throws:

```
TypeError: Failed to execute 'fetch' on 'Window': Failed to read the 'headers' 
property from 'RequestInit': String contains non ISO-8859-1 code point.
```

### Root Cause

In htmx 4.0, the switch from `XMLHttpRequest` to `fetch` API introduced this regression. The `fetch` API strictly requires HTTP header values to be ISO-8859-1 encoded, but htmx was sending raw Unicode IDs in the `HX-Trigger` and `HX-Target` headers.

`XMLHttpRequest` had a fallback for encoding non-ASCII characters, but `fetch` does not and throws an error instead.

## Solution

Encode the ID values using `encodeURI()` before setting them as header values. This matches the behavior that htmx 2.0 had with `XMLHttpRequest`.

### Changes

- `src/htmx.js` (`getHeaders` function, lines 3700 and 3702)
  - Wrap `getRawAttribute(elt, 'id')` with `encodeURI()` for `HX-Trigger` header
  - Wrap `getAttributeValue(target, 'id')` with `encodeURI()` for `HX-Target` header
  - Add null checks to handle cases where ID is not present

## Impact

- ✅ Fixes Unicode ID support in htmx 4.0
- ✅ Maintains backward compatibility (`encodeURI` only encodes non-ASCII characters)
- ✅ Backend servers can decode the header values using `decodeURI()`
- ✅ No breaking changes for existing code with ASCII IDs

## Example

**Before:**
- Element: `<div id="container-中文">`
- Result: `fetch` throws `TypeError`

**After:**
- Element: `<div id="container-中文">`
- Header: `HX-Target: container-%E4%B8%AD%E6%96%87`
- Backend can decode: `decodeURI("container-%E4%B8%AD%E6%96%87")` → `"container-中文"`

## Testing

Tested with:
- Chinese characters: `id="container-中文"` ✅
- Emoji: `id="container-😭"` ✅
- Mixed: `id="foo-中文-bar-😭"` ✅
- ASCII only: `id="container"` ✅ (no change in behavior)

## Credit

Diagnosis by @MichaelWest22 and @1cg in issue comments.


MichaelWest22:
Sorry this issue is only for htmx 4 as htmx 2 you have targeted here already has encoding of all headers being sent.  We have already merged and released the fix required for htmx4.

PR #3797, opened 2026-05-14
Improve Non form input and checkbox handling

MichaelWest22:
## Description

# collectFormData / addInputValues Fixes

## Root Cause

`__collectFormData` had a separate code path for non-form trigger elements:

```js
if (!form && elt.name) {
    formData.append(elt.name, elt.value)  // unconditional, ignores checked state
}
```

This bypassed `__addInputValues` entirely, which meant checkbox/radio checked state, select-multiple, and disabled state were all ignored.

### 1. Checkbox/radio trigger sends correct value
Unchecked checkboxes and radios no longer send their value. Previously always sent `name=on` regardless of checked state.

### 2. Select-multiple trigger sends all selected values
Previously only sent the first selected option via `elt.value`. Now correctly sends all selected options.

### 3. Button trigger value preserved
Named buttons continue to send their value when used as a direct trigger element. The refactored `__addInputValues` handles buttons as a special case (self-only, no child collection) to maintain this existing behavior.

### 4. Disabled inputs excluded via `:disabled`
Changed `input.disabled` check to `input.matches(':disabled')`. The `disabled` property does not reflect `fieldset[disabled]` ancestry — the `:disabled` pseudo-class does. Both directly disabled inputs and inputs inside a disabled fieldset are now correctly excluded.

### 5. Fieldset and container elements collect child inputs (method-aware)

The old `elt.name` guard meant that any nameless trigger element sent nothing at all. Now `__addInputValues` handles collection with GET/POST awareness via an `isGet` parameter:

- **GET/DELETE on `input, select, textarea, fieldset`** — collects self and descendants
- **GET/DELETE on other elements (div, span, etc.)** — sends nothing (prevents junky URLs on div-as-anchor patterns)
- **POST/PUT/PATCH on any element** — always collects descendants (body isn't visible/limited)
- **Button (any method)** — only sends own value, never collects children

This enables the fieldset validation pattern while preserving the existing GET scoping rule — `<input hx-get="/validate/email">` inside a 20-field form still only sends itself (useful for per-field server-side validation on blur).

The GET restriction on containers (div, span) prevents the "div as anchor" problem where `<div hx-get="/page">` would accidentally vacuum up child inputs into the URL.  I think we could change this to always gather child inputs for div's and anchors etc if we wanted to as this would not impact elements inside forms not submitting the whole from in a get/delete.

When called from `hx-include` without `isGet`, it defaults to falsy — always collecting descendants. This is correct since `hx-include` is explicit inclusion.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -451,7 +451,7 @@ var htmx = (() => {
                 : (elt.form || elt.closest("form"))
 
             // Build request body
-            let body = this.__collectFormData(elt, form, evt.submitter, ctx.request.validate)
+            let body = this.__collectFormData(elt, form, evt.submitter, ctx.request.validate, usesQueryParams)
             if (!body) return  // Validation failed
             let valsResult = this.__getAttributeObject(elt, "hx-vals", obj => {
                 ctx.vals = obj; // make available for json extensions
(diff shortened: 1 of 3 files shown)


1cg:
Thank you Latent, another great change

PR #3798, opened 2026-05-14
unify hx-on and hx-trigger modifier grammar

scriptogre:
- enable `hx-on="eventSpec => code"` syntax that supports all existing `hx-trigger` modifiers (e.g. `[filter]`, `once`, `delay`, `changed`, `from`, etc.).

- add new modifiers to both hx-on and hx-trigger: `prevent`, `stop`, `halt`, `capture`, `passive`, `from:self`, `from:outside`
- remove undocumented dot modifiers (.prevent, .stop, .halt, .once, .self, .outside, .capture, .passive, .cc) from hx-on:event syntax
- rename `__onTrigger` -> `__onEvent`
- rename `__parseTriggerSpecs` -> `__parseEventSpecs`
- remove `hx-trigger` queue modifier from docs (no-op in 4.0)
- update tests, docs, and extension references

### DIFF
--- a/src/ext/hx-preload.js
+++ b/src/ext/hx-preload.js
@@ -8,7 +8,7 @@
         let preloadEvents = []
         let timeout = 5000;
         if (preloadSpec) {
-            let specs = api.parseTriggerSpecs(preloadSpec);
+            let specs = api.parseEventSpecs(preloadSpec);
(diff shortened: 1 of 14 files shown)


1cg:
Merged @MichaelWest22 version, thank you @scriptogre!

Issue #3802, opened 2026-05-16
[Feature Request] "rootMargin" for revealed/intersect hx-trigger


YamiOdymel:
Currently, HTMX's `revealed` and `intersect` triggers require an element to enter the viewport **precisely**, which is inefficient for infinite scrolling, because you want things to load _before_ user scrolled to the end of the list.

However, the built-in [`IntersectionObserver: rootMargin`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin) property works like a CSS margin. Setting it to `100px` expands the trigger area, so the request is sent 100px *before* the element actually enters the viewport.

So instead of just `root:<selector>` and `threshold:<float>` for `intersect` hx-trigger, I hope a `rootMargin` can also be added.

YamiOdymel:
nvm I found its duplicated of #1349 and #2539, but [last comment](https://github.com/bigskysoftware/htmx/pull/2593#issuecomment-4409488486) was added last week, I still hope it can be added 🫠

MichaelWest22:
yeah should be able to add this in easily in 4

MichaelWest22:
out in the latest Beta now

PR #3805, opened 2026-05-19
unify hx-on and hx-trigger modifier grammar2

MichaelWest22:
## Description

Built some changes on top of the great work in #3798

Changes @scriptogre made:

enable hx-on="triggerSpec -> code" syntax that supports all existing hx-trigger modifiers (e.g. [filter], once, delay, changed, from, etc.).

add new modifiers to both hx-on and hx-trigger: prevent, stop, halt, capture, passive, from:self, from:outside

remove undocumented dot modifiers (.prevent, .stop, .halt, .once, .self, .outside, .capture, .passive, .cc) from hx-on:event syntax

remove hx-trigger queue modifier from docs (no-op in 4.0)

update tests, docs, and extension references

Changes I've made on top:

reverted the triggerSpec -> eventSpec and onTrigger -> onEvent renames.  I feel that the old trigger name is clearer here and more understood.  these specs and handling can be non event based like load, intersect etc and can have modifiers in how they trigger over just a pure event based action so I think Trigger is still the right name for them both.

improved the JS handling and splitting of the triggerSpec -> code using a simple regex.  with -> as the new seperator it does not conflict with JS syntax and we can use the ; at the end to identify chained triggerSpec's cleanly now.

Simplified the onTrigger() function to use less deeply nested chained callbacks.  Found that the reworked hx-on . modifiers had a simpler handler design and we can reuse this in onTrigger to cut out much of the code complexity and just leave the things like delay etc that have to stay as handler wrappers.

Also added in rootMargin support to the IntersectionObserver which is only one line to add now which is easy. #1349 #2593 #3802

### DIFF
--- a/src/ext/hx-preload.js
+++ b/src/ext/hx-preload.js
@@ -8,7 +8,7 @@
         let preloadEvents = []
         let timeout = 5000;
         if (preloadSpec) {
-            let specs = api.parseEventSpecs(preloadSpec);
+            let specs = api.parseTriggerSpecs(preloadSpec);
             if (specs.length === 0) return;
             for (const spec of specs) {
                 preloadEvents.push(spec.name)
--- a/src/ext/hx-preload.js
+++ b/src/ext/hx-preload.js
@@ -8,7 +8,7 @@
         let preloadEvents = []
         let timeout = 5000;
         if (preloadSpec) {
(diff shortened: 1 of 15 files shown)


1cg:
Thanks @MichaelWest22

PR #3809, opened 2026-05-21
handle outerSync re-processing of the target element

MichaelWest22:
## Description
The outerSync swap style had an issue with newContent processing we just fixed but looks like I missed one case.

The target element itself was never cleaned up and re-processed after __copyAttributes synced its attributes. If hx-get, hx-trigger, or other hx-attributes changed on the root node, old listeners remained active and new ones were never bound.

The fix cleans up the target (removing old listeners from it and its descendants), deletes its _htmx state, and sets newContent = [target] so the existing processing loop re-initializes the target and all its new children from scratch.

Corresponding issue:

## Testing
Added new swap tests for the outerSync swap style handling

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1351,11 +1351,10 @@ var htmx = (() => {
                     }
                 } else if (swapStyle === 'outerSync') {
                     this.__copyAttributes(target, fragment.firstElementChild);
-                    for (const child of target.children) {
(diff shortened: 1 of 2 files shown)


1cg:
thank you latent!

PR #3811, opened 2026-05-24
improve hx-csp nonce gating on all internal eval calls to cover extension use

MichaelWest22:
## Description
Add additional eval time protection to check hx-nonce is valid to also cover extension use of api.executeJavaScript.
We strip hx- attributes on missing/invalid nonces on htmx active element and during safe-eval replacement of internal eval we already re-check the hx-nonce attribute is in place at eval time as a defense in depth.  But we can also cover the non safe-eval use cases as well.

Corresponding issue:

## Testing
Re-ran manual hx-csp tests

## Checklist

  source changes)
  approved via an issue


1cg:
Merged, thanks @MichaelWest22! (Resolved a trivial conflict in test/tests/ext/hx-live.js — the comment-typo line was already gone from four-dev.)

PR #3830, opened 2026-06-08
prevent re-initilization of hx-on from process or morph

MichaelWest22:
## Description
This prevents hx-on listeners being re applied when you morph or process existing nodes.  Without this the events can fire multiple times causing confusion.  

This does not resolve changing of existing hx-on attributes only preventing the duplication and change issues.  This matches how htmx 4 currently handles core hx-trigger actions as well which has a similar guard already.

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue


MichaelWest22:
moved changes into #3834

PR #3832, opened 2026-06-09
add ctx context to hx-confirm, hx-vals and hx-headers

MichaelWest22:
## Description
to support proper hx-confirm js syntax you need to be able to access the ctx object to do things like set request headers based on confirmed data like prompts etc.  The initial intention of the hx-confirm js support was to allow implementaiton of hx-params like patterns but without ctx access this is just not easy.  

Also found We should have ctx in hx-vals and hx-headers so it can be used here when needed to access or check the action url or method etc.  ctx is already available in many hx-on situations where it makes sense so this just extends this pattern.

Corresponding issue:

## Testing
Tests still pass and no real change other than extra context in JS expressions

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -409,10 +409,10 @@ var htmx = (() => {
             return headers;
         }
 
-        __handleHxHeaders(elt, headers) {
+        __handleHxHeaders(elt, ctx) {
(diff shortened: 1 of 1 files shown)


1cg:
Thank you!

PR #3834, opened 2026-06-09
cleanup htmx attributes for reinit during morph

MichaelWest22:
## Description
Handle morph changing or removing hx- attributes so that it can cleanup and reinitialize them clean.  outerSync was already doing this but just moving this logic inside copyAttributes means we now cover morph as well

Corresponding issue:

## Testing
Already tests for this in outerSync swap tests but added tests to morph as well

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/dist/htmx.d.ts
+++ b/dist/htmx.d.ts
@@ -76,7 +76,7 @@ export interface Htmx {
   on(event: string, handler: (evt: Event) => void): void;
   on(target: string | Element, event: string, handler: (evt: Event) => void): void;
   onLoad(callback: (elt: Element) => void): void;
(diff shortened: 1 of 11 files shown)


MichaelWest22:
Now also handle hx-live cleanup properly so it responds to the cleanup events when morph detects hx-attributes changing and re-inits all hx attributes.  This prevents duplicat live functions being registered on morph etc.  Also brought in additional tests from @scriptogre branch and my other PR changes to block hx-on from duplicating listeners on re process.

Note also had to extend the hx attribute lookup in morph to check :* attributes for changes as well

PR #3835, opened 2026-06-09
fix stale hx-* handlers on re-process

scriptogre:
Fixes stale hx-on/hx-trigger handlers after morph or re-process.

Previously:
```js
btn.setAttribute('hx-on:click', 'doV2()')
htmx.process(btn)
btn.click() // fired doV1 AND doV2
```

Now only the current handler fires. Cleanup is scoped to elements whose wiring attrs (hx-trigger, hx-boost, hx-on*) actually changed. `hx-vals` changes, `hx-trigger` with `once` modifiers, and child elements are unaffected.

Adds 12 tests. All of #3830 / #3834's tests pass.

Related: #3830, #3834

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -795,9 +795,12 @@ var htmx = (() => {
                     }, this.parseInterval(interval));
                 }
 
-                // Load: fire immediately, no listener needed
+                // Load: fire once per element, never on re-init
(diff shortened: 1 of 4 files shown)


1cg:
Decided on explicit refresh rather than hash-based tracking, thank you christian!

Issue #3821, opened 2026-06-05
htmx 4: Problem morphing `<select>` field in Firefox


lukaskleinschmidt:
I found a small bug (maybe) when using `innerMorph` or `outerMorph`.
It looks like the option texts are getting merged.

It works fine in Firefox `151` for me.
Chrome also seem to work fine.

But in Firefox `152` and `153` it looks like this when option texts change:

<img width="308" height="92" alt="Image" src="https://github.com/user-attachments/assets/5d353e23-6a74-4371-bfd3-ec382d8a55c3" />

----

- `151` Firefox
- `152` Firefox Developer Edition
- `153` Firefox Nightly

----

The HTML in the inspector looks ok.
Not even sure if this is something htmx can solve.

Here a small [example](https://kleinschmidt.at/htmx/index.html)
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Document</title>
    <script src="[https://cdn.jsdelivr.net/npm/htmx.org@4.0.0-beta4](view-source:https://cdn.jsdelivr.net/npm/htmx.org@4.0.0-beta4)" integrity="sha384-aWZK1NtOs/aWb/+YZdTM8q2JkWEshlMc9mgZ189numT9bwFhyAyYEoO4nO/2dTXt" crossorigin="anonymous"></script>
    <style>
        select {
            width: 100%;
        }
    </style>
</head>
<body>

    <form id="form">
        <select
            name="date"
            hx-get="test-1.html"
            hx-target="#form"
            hx-swap="innerMorph"
        >
            <option value="1">Date 1 (10/10)</option>
            <option value="2">Date 2 (10/10)</option>
        </select>
    </form>

</body>
</html>
```

The example currently loops through two different fragments:
[test-1.html](https://kleinschmidt.at/htmx/test-1.html)
```html
<select
    name="date"
    hx-get="test-2.html"
    hx-target="#form"
    hx-swap="innerMorph"
>
    <option value="1">Date 1 (8/10)</option>
    <option value="2">Date 2 (10/10)</option>
</select>
```

[test-2.html](https://kleinschmidt.at/htmx/test-2.html)
```html
<select
    name="date"
    hx-get="test-1.html"
    hx-target="#form"
    hx-swap="innerMorph"
>
    <option value="1">Date 1 (2/10)</option>
[...]

PR #3836, opened 2026-06-11
Improve text node morphing for firefox support

MichaelWest22:
## Description
There is an issue in the latest firevox 151 build where it now does not handle htmx morphing select options elements text nodes well and displays both before and after text node data in error.  When we morph text nodes we currently just block morphing so the new text node has to be inserted and the old one then removed.  This was done as a performance tweak to avoid wasting time mathing and morphing just text nodes that contain no preservable state.  But for a brief period both text nodes are in the options element which is not ideal and now triggering a bug in firefox.

To solve this I've added code to have a fast path to just match text nodes if both old and new nodes are both text nodes.  

Corresponding issue:
#3821 

## Testing
had to perform manual testing with firefox developer edition to validate the fix resolves the issue

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2003,6 +2003,8 @@ var htmx = (() => {
         }
 
         __findBestMatch(ctx, node, startPoint, endPoint) {
+            // text nodes match positionally — patch in place via __morphNode, 3 = TEXT_NODE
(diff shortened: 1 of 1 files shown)


1cg:
Thank you!

PR #3844, opened 2026-06-16
Fix `hx-swap-oob`-only responses wiping main target + document behavior

scriptogre:
`__processMainSwap` doesn't behave consistently for responses that only contain `<hx-patial>`s vs only `hx-swap-oob`:

```html
<!-- This response would wipe `hx-target` if `hx-swap="innerHTML"` -->
<div id="notifications" hx-swap-oob="true">3 new messages</div>
```

```html
<!-- But this one wouldn't -->
<hx-partial hx-target="#notifications">3 new messages</hx-partial>
```

**Fix:** Make behaviour consistent (skip main target in both scenarios), and make the code more obvious about it.

Also documents "responses without main content" behavior in the hx-swap-oob reference page, docs, and a new `<hx-partial>` reference page.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1186,7 +1186,7 @@ var htmx = (() => {
                 tasks.push(...oobTasks, ...partialTasks);
 
                 // Process main swap first
-                let mainSwap = this.__processMainSwap(ctx, fragment, partialTasks);
+                let mainSwap = this.__processMainSwap(ctx, fragment, tasks);
                 if (mainSwap) {
                     tasks.unshift(mainSwap);
                 }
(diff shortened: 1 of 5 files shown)


1cg:
Thank you for this! We went with #3862 for the empty-swap behavior, which keeps oob-only destructive by default and adds an opt-out swapEmpty modifier, so this one is superseded and now conflicts. Closing it.

One thing worth saving: the new `<hx-partial>` reference page (reference/06-tags/01-hx-partial.md) does not exist on four-dev and was not part of #3862. If you would re-submit just that doc page on its own, we would happily take it.

PR #3846, opened 2026-06-17
Fix hx-prompt Unicode header encoding

### LINKED ISSUE #3841: [4.0.0-beta4] BUG: hx-prompt not working on Unicode input [error: String contains non ISO-8859-1 code point]
https://github.com/bigskysoftware/htmx/pull/3828 adds `hx-prompt` back to HTMX 4 as an extension to fix https://github.com/bigskysoftware/htmx/issues/3825

I grabbed the extension directly from the pull request, but then I encountered an error with `4.0.0-beta4` if prompt contains Chinese character:

```
htmx: htmx:error: Failed to execute 'fetch' on 'Window': Failed to read the 'headers' property from 'RequestInit': String contains non ISO-8859-1 code point. TypeError: Failed to execute 'fetch' on 'Window': Failed to read the 'headers' property from 'RequestInit': String contains non ISO-8859-1 code point.
    at #issueRequest (htmx.js:544:42)
    at #handleTriggerEvent (htmx.js:509:37)
    at htmx.js:343:51
    at inner (htmx.js:719:21)
    at spec.handler (htmx.js:769:21)
```

```
L544                 let response = await ctx.fetch(ctx.request.action, ctx.request);
L509                 await this.#issueRequest(ctx);
L343                 await this.#handleTriggerEvent(ctx);
L719                 handler(evt);
L769                 timed(evt);
```

This appears to be the same issue as https://github.com/bigskysoftware/htmx/issues/3781, but [the comment](https://github.com/bigskysoftware/htmx/issues/3781#issuecomment-4415281815) mentioned that it was fixed in `4.0.0-beta3` 🤔?

cyphercodes:
## Description
URI-encodes the `HX-Prompt` header value set by the `hx-prompt` extension so Unicode prompt responses can be used with `fetch` without triggering the non ISO-8859-1 header-value error.

Corresponding issue: Fixes #3841

## Testing
- `npx web-test-runner --files test/tests/ext/hx-prompt.js --browsers chromium --config test/web-test-runner.config.mjs --playwright --coverage=false`
- `npm test`

## Checklist

* [x] I have targeted this PR against the correct branch (`four-dev` for this htmx 4 beta bugfix)

### DIFF
--- a/src/ext/hx-prompt.js
+++ b/src/ext/hx-prompt.js
@@ -17,7 +17,7 @@
             let answer = (window.htmxPrompt || window.prompt)(question);
             if (answer === null) return false;
             if (!htmx.trigger(ctx.sourceElement, 'htmx:prompt', { prompt: answer, target: ctx.target })) return false;
-            ctx.request.headers['HX-Prompt'] = answer;
+            ctx.request.headers['HX-Prompt'] = encodeURI(answer);
         }
     });
 })();
--- a/test/tests/ext/hx-prompt.js
+++ b/test/tests/ext/hx-prompt.js
@@ -42,6 +42,28 @@ describe('hx-prompt extension', function() {
         assert.equal(lastFetch().request.headers['HX-Prompt'], 'because');
     });
 
+    it('URI-encodes unicode prompt responses for fetch headers', async function() {
+        window.prompt = () => '中文 reason';
+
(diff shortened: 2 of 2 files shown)


MichaelWest22:
Thanks this is indeed the valid fix we already applied to the other headers and should resolve the issue.

PR #3847, opened 2026-06-17
Update JSDoc and parameter naming for `process()` and `cleanup()`

scriptogre:
.


1cg:
Ported the JSDoc and the .d.ts signature widening (process now typed Element | ShadowRoot) onto four-dev directly, adapted to the new force parameter from #3834. Dropped DocumentFragment from the type since process() is intended for live/connected content only. Thanks scriptogre!

Issue #3854, opened 2026-06-21
v4: multiple selector doesn't work on `hx-trigger`


boltlessengineer:
```html
<form
  hx-trigger="keydown[(ctrlKey || metaKey) && key=='Enter'] from:(#title,#body)"
>
```

```
Uncaught SyntaxError: Failed to execute 'querySelectorAll' on 'Document': '(#title' is not a valid selector.
```

Seems like new parser splits the entire attribute by `,` before parsing it. Same attribute was fine in v2.

boltlessengineer:
So I read the code. Seems like there isn't any logic for parentheses yet.

MichaelWest22:
Yeah this has not been added sorry.  The new HCON parser we now use only currently handles `from:'#title,#body'` `from:"#title,#body"` and `from:<#title,#body/>` formats.  and the split on comma breaks it before this as well.   Kind of annoying we have multiple old wrapper formats as we have the < /> one for CSS selectors that have space or comma in them that need to be wrapped but then for some reason in v2 we also have () wrapping just for the from: CSS selector!  Trying to support all the various formats is kind of annoying.  We already have code to we could reuse to handle comma splits in CSS selectors for the < /> format which we could reuse for the trigger parser comma split maybe.  But having to expand HCON to a 4th (...) wrapping format is kind of annoying :/

boltlessengineer:
@MichaelWest22 It's ok since I'm trying out a beta software. 🙃
I don't mind switching from `()` to other wrapping format. I just found they all break with commas yet.

I'll make a PR to fix this.

PR #3856, opened 2026-06-21
support comma in trigger spec config

boltlessengineer:
e.g. `hx-trigger='click[ctrlKey] from:"previous button, next a"'`

Close: #3854

## Description

Implemented lazy mode in `__parseConfig` to stop on comma separator. This allows parsing `hx-trigger` spec like `click from:"input, textarea"`.

```
__parseConfig(configString [, lazy])
```
When `lazy` is truthy and first character is not `{`, `__parseConfig` will return `{config, cursor}` where cursor pointing at separator `,`.

`__parseTriggerSpecs` will walk through the given string instead of aggressively separating with comma first. This allows `,` to be handled from `__parseConfig` with the knowledge of HCON format.

Corresponding issue:

## Testing
added new test cases for:
```
click[ctrlKey] from:"form input"
```
and
```
click[ctrlKey] from:"previous button, next a"
```

First one was already passing without changes, but I added it to compare with second case.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -242,18 +242,24 @@ var htmx = (() => {
             return defaultVal;
         }
 
-        __parseConfig(configString) {
+        __parseConfig(configString, lazy) {
             if (!configString) return {};
             if (configString[0] === '{') return JSON.parse(configString);
-            let configPattern = /(?:"([^"]+)"|([^\s,:]+))(?:\s*:\s*(?:"([^"]*)"|'([^']*)'|<([^>]+)\/>|([^\s,]+)))?(?=\s|,|$)/g;
-            return [...configString.matchAll(configPattern)].reduce((result, match) => {
+            let configPattern = /(?:"([^"]+)"|([^\s,:]+))(?:\s*:\s*(?:"([^"]*)"|'([^']*)'|<([^>]+)\/>|([^\s,]+)))?(?=\s*(,)|\s|$)/g;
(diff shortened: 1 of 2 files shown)


scriptogre:
Thanks for your PR! 

It is very eerie though, I started fixing this yesterday, independently from you 🫨.

I will try to converge our two approaches into a (hopefully) minimally invasive one.

MichaelWest22:
Yeah i already worked on a quick version of this here https://github.com/bigskysoftware/htmx/compare/four-dev...MichaelWest22:htmx:comma-split

we have to be very careful how we expand or abuse the HCON parseConfig parser as we don't want it getting too complex to manage and getting too far away from its core simplicity.  

For this issue I think we may need to think carefully about the api format we want for this parsing and not rush into the first simple fix as any parsing decision we make we will have to live with long term. 

Do we want the old from:(a,b) format to work as it did?  do we support just the < /> hyperscript format for this feature so we have just one recommended way of wrapping CSS selectors with space and comma?  do we go with detecting comma in just "" and '' or do we do all 4 formats '' "" () < /> and make it universal with more code?

1cg:
Superseded by #3859, which formalizes this into an HCON.parse/split/merge namespace and fixes the same comma-in-config case (#3854). Thank you for the fix and the clear repro, @boltlessengineer - it directly informed the direction here.

PR #3859, opened 2026-06-22
Formalize HCON parsing and comma-aware splitting

scriptogre:
This formalizes HCON as a small internal parsing namespace, mirroring the shape of JavaScript's native JSON API:

- `HCON.parse(...)`
- `HCON.merge(...)`
- `HCON.split(...)`

The main goal is to make the HCON behavior easier to reason about, test, and evolve before we lock in more parsing behavior.
  
Related: #3856, #3854

This would enable doing:
```html
<!-- _hyperscript wrapper -->
<form hx-trigger="keydown[(ctrlKey || metaKey) && key=='Enter'] from:<#title, #body/>">
<!-- or double quotes -->
<form hx-trigger='keydown[(ctrlKey || metaKey) && key=="Enter"] from:"#title, #body"'>
<!-- or single quotes -->
<form hx-trigger="keydown[(ctrlKey || metaKey) && key=='Enter'] from:'#title, #body'">
```


scriptogre:
Rebased.

codypotter:
Maybe the docs should be updated to reflect this breaking change?

[This example](https://four.htmx.org/reference/attributes/hx-trigger#fromselector) no longer works since this was introduced.

Specifically this is broken: `<div hx-trigger="submit from:closest form" hx-get="...">...</div>`

@scriptogre

scriptogre:
@codypotter Thanks a lot for finding that!

PR #3862, opened 2026-06-22
add empty swap handling override and make sse default to not swap empty responses

MichaelWest22:
## Description
When using sse with hx-swap-oob swaps it can easily blank out the target if only hx-swap-oob are passed in a response.  We already have new behavior where if only hx-partial commands are included in a swapped response we now special case this and block the main swap if it is empty as sending only partial commands is a new advanced action and this new behavior is not a breaking change from before.  But for hx-swap-oob to retain support for valid legacy users who may have hx-swap-oob responses that also blank out the requesting form as well as swapping in the form response into a oob target we should keep the old behavior by default.  

So to allow users to control the swap on empty behavior and allow sse to set a safer default i've added a new empty swap modifier and set it up to have a defaultSwapEmpty config or empty:true/false modifier.

Corresponding issue:

## Testing
Added sse and swap tests

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/hx-sse.js
+++ b/src/ext/hx-sse.js
@@ -258,7 +258,7 @@
 
                         // Swap content using the ctx from core (target/swap already resolved)
                         ctx.text = detail.message.data;
-                        if (!ctx.swap.includes('empty')) ctx.swap += ' empty:false';
+                        if (!ctx.swap.includes('swapEmpty')) ctx.swap += ' swapEmpty:false';
                         await htmx.swap(ctx);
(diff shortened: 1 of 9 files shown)


1cg:
Hey Latent,

PR looks good, my only feedback is that `empty` is a little ambiguous in my mind.  What do you think about swapEmpty instead?  That lines up w/defaultSwapEmpty, in the config too.  Does that sound reasonable?

MichaelWest22:
@1cg updated

PR #3851, opened 2026-06-19
Unify morph preservation around `hx-preserve`

scriptogre:
## Changes

- Rename `morphSkip`               -> `morphPreserve`
- Rename `morphSkipChildren` -> `morphPreserveChildrenOf`
- Rename `morphIgnore`            -> `morphPreserveAttributes`
- Make [`hx-preserve`](https://four.htmx.org/reference/attributes/hx-preserve) work with morph
- Add `hx-preserve:children` (`innerMorph`/`outerMorph`)
- Add `hx-preserve:attributes` (`innerMorph`/`outerMorph`/`outerSync`)
- Enable `*` & `(a|b)` patterns for attribute preservation (e.g. `data-*`)
- _console.warn_ when `hx-preserve` is used without `id` (outside morph)

## Examples

**`hx-preserve` in morph**
```html
<!-- This element is updated via JS -->                                                                                                                                       
<span id="visitors" hx-preserve>1247</span>
```

**`hx-preserve:children`**
```html
<!-- Children managed by Select2 -->
<select hx-preserve:children>
  <option>Loaded by Select2, won't be replaced</option>
</select>
```

**`hx-preserve:attributes`**
```html
<!-- web component stores scroll/cursor in data-* attrs -->                                                                                                                       
<code-editor data-scroll="120" data-cursor="42" hx-preserve:attributes="data-*">                                                                                                                                 
</code-editor>

<!-- ARIA state set by JS dropdown -->
<div aria-expanded="true" aria-controls="menu" hx-preserve:attributes="aria-(expanded|controls)"></div>
```

**Patterns**
```javascript
htmx.config.morphPreserveAttributes = "data-state-*, aria-(label|hidden), style";
```

PR #3864, opened 2026-06-23
add morph skip default configs

MichaelWest22:
## Description
Unify morph skip around server-driven attributes + fix outerSync cleanup

Changes
Add morphSkip default '[hx-morph-skip]' and morphSkipChildren default '[hx-morph-skip-children]' so server templates can opt elements in/out of morph freezing declaratively without config changes

Fix outerSync leaking htmx-powered child listeners/intervals when replacing children (mirrors how innerHTML already does cleanup)

Add hx-morph-skip-attrs extension for wildcard/alternation attribute name preservation during morph (additive to morphIgnore, implemented via htmx:before:morph:attr hook)

Motivation
PR #3851 proposed renaming morphSkip/morphSkipChildren/morphIgnore and adding per-element hx-preserve:children / hx-preserve:attributes modifiers. The rename breaks backwards compat, the per-element modifiers override server intent using live DOM state which conflicts with htmx's server-driven design, and the hx-preserve overloading conflates two different mechanisms.

Just trying to take the useful parts and implement them in a minimal way.

The useful attribute regex solution is a lot of extra code for a small edge cases that I think should probably live in an extension since we now have a great extension point for this use case we added for hx-live cleanup

Possible extension:
```JS
(() => {
    let api;
    let globalPatterns = [];

    function compilePatterns(patterns) {
        if (!patterns) return [];
        let list = Array.isArray(patterns) ? patterns : String(patterns).split(/[\s,]+/);
        return list.filter(Boolean).map(p =>
            p instanceof RegExp ? p :
            new RegExp('^' + p.replace(/[.+?^${}[\]\\]/g, '\\$&').replace(/\*/g, '.*') + '$')
        );
    }

    htmx.registerExtension('morph-skip-attrs', {
        init: (internalAPI) => {
            api = internalAPI;
            globalPatterns = compilePatterns(htmx.config.morphSkipAttributes);
        },

        htmx_before_morph_attr: (elt, detail) => {
            if (globalPatterns.some(r => r.test(detail.attrName))) return false;
            let perElement = api.attributeValue(elt, 'hx-morph-skip-attrs');
            if (perElement && compilePatterns(perElement).some(r => r.test(detail.attrName))) return false;
        }
    });
})();
```

Corresponding issue:

## Testing
Added tests

## Checklist

  source changes)
  approved via an issue


1cg:
Thank you!

PR #3866, opened 2026-06-24
Do not boost download links

scriptogre:
## Question

Should `hx-boost` boost `<a>` links that have the `download` attribute?

This PR assumes it should not. Was `download` support mistakenly left out of the anchor boost checks, or was boosting these links a deliberate choice?

## Changes

- Skip boosting anchors with `download`
- Add a unit test for `<a href="/report.pdf" download>`
- Document anchor boost behavior with examples


1cg:
Thank you!

PR #3868, opened 2026-06-26
improve morph preservation with focused buttons

MichaelWest22:
## Description
Fix morph scan blocking on focused non-text elements and add focused input protection test

When a morph response arrives while a button has transient focus (e.g. just clicked to trigger the request), the __findBestMatch scan was stopping at any element containing document.activeElement — including buttons. This caused sibling inputs after the button to be inserted fresh rather than morphed in place, losing any live values the user had typed.

The fix restricts the focus stop-guard to only halt the scan when the focused element is a typeable input (selectionStart != null). This correctly covers the case where a new element (e.g. a validation error div) is inserted before a focused text input in the response. Without the guard, the scan would advance past the focused input to soft-match a later sibling, and the skip-over loop would then displace and remove the input the user is actively typing into.  But for the non text case we do not need or want this guard to apply as we are doing a worse preservation match to save pointless focus.

Corresponding issue:

## Testing
Added a couple of tests

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2104,8 +2104,8 @@ var htmx = (() => {
                 // Stop if too many ID elements would be displaced
                 displaceMatchCount += oldSet?.size || 0;
                 if (displaceMatchCount > nodeMatchCount) break;
-                // Don't move elements containing focus
-                if (cursor.contains(document.activeElement)) break;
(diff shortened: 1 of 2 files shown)


1cg:
Thank you!

PR #3870, opened 2026-06-27
update docs that SSE + idiomorph requires htmx-ext-sse 2.2.4+

scriptogre:
I noticed LLMs tend to use older versions of the sse & idiomorph extensions, which makes this a common issue. It might be worth adding a short note in the docs.

bigskysoftware/htmx-extensions#164

(Re-opened from #3827, which got auto-closed when I renamed the head branch on my fork.)


1cg:
Thank you!

Issue #3871, opened 2026-06-28
Regression (4.0.0-beta4): form-associated custom element value not sent when it is a standalone (non-form) trigger


gregjotau:
### Summary

In htmx 4, a **form-associated custom element (FACE)** that is used as a **standalone trigger** (i.e. it has a `name`, is *not* inside a `<form>`, and fires its own request) no longer has its value included in the request. This worked in `4.0.0-beta3` and broke in `4.0.0-beta4` (changelog: *"Improved non-form input and checkbox handling"*), and is still broken in `4.0.0-beta5`.

This silently drops a parameter that used to be sent, which is hard to diagnose (no console error — the server just sees a missing param).

### htmx version

- ✅ Works: `4.0.0-beta3`
- ❌ Broken: `4.0.0-beta4`, `4.0.0-beta5`

### Minimal reproducible example

A standards-compliant form-associated custom element (no framework needed) that exposes `.value` and calls `ElementInternals.setFormValue`, used directly as the htmx trigger:

```html
<!doctype html>
<script src="https://cdn.jsdelivr.net/npm/htmx.org@4.0.0-beta5"></script>
<script>
  class MyControl extends HTMLElement {
    static formAssociated = true;
    #internals = this.attachInternals();
    connectedCallback() { this.value = 'hello'; }
    get value() { return this.#value; }
    set value(v) { this.#value = v; this.#internals.setFormValue(v); }
  }
  customElements.define('my-control', MyControl);
</script>

<!-- standalone trigger: has a name, NOT inside a <form> -->
<my-control name="foo" hx-post="/echo" hx-trigger="click">click me</my-control>
```

Click the element and inspect the outgoing request body.

- **beta3:** body contains `foo=hello` ✅
- **beta4 / beta5:** body is empty — `foo` is missing ❌

(Our real-world case is a Web Awesome `<wa-combobox name="tenantId">` company switcher in a page header, outside any form. It stopped sending `tenantId` after upgrading.)

### Expected behavior

The triggering element's own value is included, as documented ("an element that causes a request will include its value if it has one") and as in beta3.

### Actual behavior

The parameter is omitted. No error is raised client-side.

### Root cause

In `4.0.0-beta3`, the non-form branch of `#collectFormData` appended the triggering element's own `name`/`value` directly:

```js
// beta3
if (!form && elt.name) {
    if (validate && elt.reportValidity && !elt.reportValidity()) return
[...]

PR #3872, opened 2026-06-28
handle custom element manual input value collection

MichaelWest22:
## Description
Fix: FACE value not sent when used as standalone trigger (no form)

Form-associated custom elements (e.g. <wa-combobox name="tenantId">) used as standalone htmx triggers outside a <form> stopped sending their value in beta4. This was a regression introduced by f02a86b ("Improve Non form input and checkbox handling").

The fix adds a single else if branch in __addInputValues for elements whose tag name contains a hyphen (the W3C-mandated marker for all custom elements). These are routed through the existing loop the same way BUTTON is — inputs = [elt] — so they pick up the name/disabled/included guards and fall through to formData.append(input.name, input.value) without any duplicated logic.

The form path (new FormData(form)) is unaffected and continues to collect FACE values via setFormValue as before.

Corresponding issue:
#3871

## Testing
Added tests for the new behaviour

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1816,7 +1816,7 @@ var htmx = (() => {
             if (tag === 'BUTTON' || tag.includes('-')) {
                 inputs = [elt]; // send own value only, never collect children
             } else if (['INPUT', 'SELECT', 'TEXTAREA', 'FIELDSET'].includes(tag) || !isGet) {
-                inputs = this.__queryEltAndDescendants(elt, 'input, select, textarea');
+                inputs = this.__queryEltAndDescendants(elt, '[name]:not(button)');
             }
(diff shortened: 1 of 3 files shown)


1cg:
Thank you!

AndrewGit25:
Update: i added a PR to fix this.

Thank you for this. Its seems to be working in beta6 with brief testing of .value and .checked in custom elements. 

It doesn't work as expected when the .value is an array. it returns a flattened string in formdata outside or form.  Rather than mulitple entries. Behavior should be the same as a <select multiple> for example.

```
// getting 
fruit=banana,apple

// expecting
fruit=banana
fruit=apple
```

it works as expected when in a inside a form. i'm appending the multiple entries in formdata then adding to internals
internals.setFormData(formData)

// form submits with 
fruit=banana
fruit=apple

PR #3873, opened 2026-06-28
Rename `htmx:swap:finally` to `htmx:finally:swap`

scriptogre:
We currently also have `htmx:finally:request`, and the pattern is `htmx:timing:noun`.


1cg:
Thank you!

PR #3874, opened 2026-06-28
Forward `protocols` to WebSocket constructor in hx-ws extension

scriptogre:
The extension now forwards `connection.config.protocols` as the second argument to `new WebSocket()`, enabling `Sec-WebSocket-Protocol` negotiation.

**Before:** The WebSocket constructor was always called with just a URL. Subprotocols (MQTT, STOMP, GraphQL-WS, etc.) were impossible to set. 

 **After:**
                                                                                                                                                                                                   
 ```html                                                                                                                                                                                                           
   <meta name="htmx-config" content="ws.protocols:mqtt">                                                                                                                                                 
                                                                                                                                                                                                                   
   <div hx-ws:connect="/ws" hx-config="ws.protocols:mqtt">                                                                                                                                                         
 ```

### DIFF
--- a/src/ext/hx-ws.js
+++ b/src/ext/hx-ws.js
@@ -178,7 +178,7 @@
         }
 
         try {
-            connection.socket = new WebSocket(url);
+            connection.socket = new WebSocket(url, connection.config?.protocols);
             let ac = new AbortController();
             connection.abortController = ac;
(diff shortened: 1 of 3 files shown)


1cg:
Thank you!

PR #3880, opened 2026-07-01
Guard show target scroll method

scriptogre:
## Problem

`show` scrolling only checked whether the target exists:

```js
showTarget?.scrollIntoView(...)
```

If an `outerHTML` swap replaces the target with plain text, the follow-up `show:top` or `show:bottom` handling can receive a text node as the show target. Text nodes are truthy, but do not have `scrollIntoView`, so the request fires `htmx:error`.

## Fix

Guard the method call too:

```js
showTarget?.scrollIntoView?.(...)
```

Adds a regression test for an `outerHTML show:top` swap whose response is plain text.


1cg:
Thank you!

PR #3883, opened 2026-07-06
Unify method/action resolution and improve hx-action doco

MichaelWest22:
## Description
Found we can improve hx-action to fall back to formmethod and method native attributes just like hx-boost already does.  This allows some better progressive enhancement options when moving to use the more natural action based attribute.  You can place a hx-action on a form and have it use native method attributes. 

By combining hx-action and hx-boost action and method calculation logic it also simplifies the code required as well.

Updated doco to highlight the progressive enhancment options and method fallback.

Corresponding issue:

## Testing
Added new hx-action tests to cover the new behavior.  should be no problem breaking changes for valid existing use cases as we are just adding a method fallback that was not usable before and it would have been confusing before if native method did not override the method used.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -343,34 +343,23 @@ var htmx = (() => {
         }
 
         __determineMethodAndAction(elt, evt) {
-            if (this.__isBoosted(elt)) {
-                return this.__boostedMethodAndAction(elt, evt)
-            } else {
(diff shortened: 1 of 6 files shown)


1cg:
Great change, thank you!

Issue #3882, opened 2026-07-05
[4.0.0-beta5] outerMorph can skip processing nested hx-* after root tag change with persistent child id


qFlensT:
I hit this with htmx 4.0.0-beta5.

Short version: after an `outerMorph`, the DOM contains nested `hx-get` buttons, but clicking them does nothing. No request, no console error.

It happens with this shape:
```html
<!-- before -->
<div id="target">
  <div id="loading"></div>
</div>

<!-- response -->
<form id="target">
  <button hx-get="/next" hx-target="#target" hx-swap="outerMorph">
    Next
  </button>
  <div id="loading"></div>
</form>
```

If the response keeps the root tag stable:
```html
<div id="target">
  <form>...</form>
  <div id="loading"></div>
</div>
```
then the nested `hx-get` works.

I attached a single-file repro that mocks `fetch`, so it can be opened directly in the browser without a server:
<details>
<summary>Repro HTML</summary>

```html
<!doctype html>
<html>
  <head>
    <script>
      const html = {
        "/broken": `
          <form id="broken" style="border: 2px solid red; padding: 12px">
            <div id="same-child-id">same child id</div>
            <button
              type="button"
              hx-get="/broken-next"
              hx-target="#broken"
              hx-swap="outerMorph"
            >
              broken nested hx-get
            </button>
          </form>
        `,
        "/broken-next": `
          <form id="broken" style="border: 2px solid green; padding: 12px">
            broken nested request worked
          </form>
        `,
        "/working": `
          <div id="working" style="border: 2px solid red; padding: 12px">
            <form>
              <div id="same-child-id-working">same child id</div>
              <button
                type="button"
                hx-get="/working-next"
                hx-target="#working"
                hx-swap="outerMorph"
              >
                working nested hx-get
              </button>
            </form>
          </div>
        `,
        "/working-next": `
          <div id="working" style="border: 2px solid green; padding: 12px">
            working nested request worked
          </div>
        `,
      };

      window.fetch = async (input) => {
        const url = typeof input === "string" ? input : input.url;
        const path = new URL(url, location.href).pathname;

        document.querySelector("#log").textContent += `fetch ${path}\n`;

        return new Response(html[path], {
          headers: { "content-type": "text/html" },
        });
      };
    </script>
    <script src="https://cdn.jsdelivr.net/npm/htmx.org@4.0.0-beta5/dist/htmx.min.js"></script>
  </head>
  <body>
    <h1>outerMorph repro</h1>

[...]

PR #3884, opened 2026-07-06
handle processing morph placeholder nodes

MichaelWest22:
## Description
There is an edge case where the newContent processing of outerMorph does not catch one situation well.  We currently do not fully resolve all outerMorph inserted nodes due to the added complexity required so instead we track the newContent nodes plus the target node and process these as Nodes are either inserted new or the target node is morphed in place.  But if you break the morph by changing root tags and you still have persistent id children  contained in the now completely different parent node (this is a wierd edge case that would never happen in practice but is possible if you replace the nodes completely and include a hx-preserve node you want to move to the new different structure).   When the moprh detects that there are persistent id's that need retaining in a node even though the node is not morphable it creates a new placeholder node and morphs the new node in as just placing the new node in instead can cause problems for the id nodes inside moving levels etc.  But this placeholder node is now not the same node as the newContent nodes we started with and it could replace the main target node and in this rare edge case this placeholder node is never processed.  

To patch this edge case i've just added an extra process command that just processes this placeholder node after it is inserted to ensure it will always be processed correctly.  

Also found some dead code in morphing isSoftMatch that is meant to only be in the alpine compat extension and got left in core by mistake so removing this as well.

Corresponding issue:
#3882 

## Testing
Added test for this edge case

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2060,6 +2060,7 @@ var htmx = (() => {
                     let placeholder = document.createElement(newChild.tagName);
                     oldParent.insertBefore(placeholder, insertionPoint);
                     this.__morphNode(placeholder, newChild, ctx);
+                    this.process(placeholder);
                     insertionPoint = placeholder.nextSibling;
                 } else {
                     oldParent.insertBefore(newChild, insertionPoint);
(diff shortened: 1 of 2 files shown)


1cg:
Thank you!

PR #3885, opened 2026-07-06
Adopt Fetch API's `Headers` interface

scriptogre:
Replace the request header plain object with a native `Headers` instance.

Updates request building, `hx-config`, `hx-headers`, `htmx.ajax()`, and bundled extensions to use the standard `Headers` API.

`HCON.merge(...)` now knows how to merge into `Headers`, so existing object-style config values still work:

```html
<div hx-config="headers.X-Custom: value"></div>
```

Added an `hx-compat-htmx-4-beta` auto-registered extension that keeps `ctx.request.headers[name]` working and warns with the new syntax:

```text
[htmx] Deprecated: replace ctx.request.headers[name] with Headers methods 
  old: ctx.request.headers[name]
  new: ctx.request.headers.get(name), .set(name, value), or .delete(name)
  docs: https://developer.mozilla.org/en-US/docs/Web/API/Headers
```

The extension might seem a bit overengineered for what it does now, but it was done to cleanly accommodate some other stuff I'll propose in a PR soon.

### DIFF
--- a/src/editors/jetbrains/htmx.web-types.json
+++ b/src/editors/jetbrains/htmx.web-types.json
@@ -452,7 +452,7 @@
         },
         {
           "name": "before:response",
-          "description": "Fires after response headers are received but before the body is consumed. `detail.ctx.response.raw` is the unconsumed `Response`.",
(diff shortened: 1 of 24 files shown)


1cg:
Let's keep talking about this, closing for now

PR #3888, opened 2026-07-09
cache compiled eval tags in safeEval feature for better hx-live perf

MichaelWest22:
## Description
Found that hx-csp creates temporary script tags via its safeEval feature but when using things like hx-live this means it re-creates the same eval script each dom update.  We can easily cache the functions and reuse them so they do not need to be be re-created every time.

Corresponding issue:

## Testing
Tested this manually in a test project to confirm faster performance during events.  It went from ~2ms to fire a small batch of expressions down to ~1.2ms.  before safe-eval was a bit slower by design than normal eval but now it is just a little bit faster but the perf difference is hard to notice comparing the two.  

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/hx-csp.js
+++ b/src/ext/hx-csp.js
@@ -122,6 +122,10 @@
             let counter = 0;
             let NativeFunction = Function;
             let NativeAsyncFunction = Object.getPrototypeOf(async function(){}).constructor;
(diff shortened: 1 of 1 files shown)


1cg:
Thank you!

PR #3890, opened 2026-07-12
Document htmx event response headers type

yawaramin:
Document and link to type of the response headers object for clarity.


1cg:
Thank you!

PR #3891, opened 2026-07-13
improve hx-live perf by skipping no change updates

MichaelWest22:
## Description
We can improve hx-live performance so we can support maybe 1000 live bindings on a page by guarding writes that don't change anything so it avoids wasted time updating the dom with no changes.  This gives a 2-3x speed increase

Also added a possible 100ms debounce for input events so typing into an input does not overload the live tasks at the cost of slowing down the updates from typing.  change and dom mutations fire instantly still.

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/hx-live.js
+++ b/src/ext/hx-live.js
@@ -120,7 +120,6 @@
 
         let value = rest[0];
         for (let e of elts) {
-            if (applyAttr([e], name) === value) continue;
             if (isClass) {
                 e.classList.toggle(name.slice(1), !!value);
                 if (e.classList.length === 0) e.removeAttribute('class');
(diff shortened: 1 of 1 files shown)


src/ext/hx-live.js
```diff
@@ -18,17 +18,25 @@
...
+    const INPUT_DEBOUNCE_MS = htmx.config.live?.inputDebounceMs ?? 100;
+
     function ensureActive() {
         if (observer) return;
         recomputeBound = () => schedule();
-        document.addEventListener('input', recomputeBound, true);
+        document.addEventListener('input', () => {
+            clearTimeout(inputDebounceId);
+            inputDebounceId = setTimeout(schedule, INPUT_DEBOUNCE_MS);
+        }, true);
```
1cg: should this listener be captured and removed in deactivate() like the recomputeBound listener is?

1cg:
Thank you!

PR #3892, opened 2026-07-13
Allow custom methods like query

MichaelWest22:
## Description
We can allow custom methods with hx-action and h-method like the new Query method which is being implemented.  before we had a guard blocking all non standard methods but we only need to block the special case of 'dialog' which we have to ignore to let the browser handle it.

Corresponding issue:

## Testing
Added a test for hx-method of query

## Checklist

  source changes)
  approved via an issue


1cg:
Thank you!

PR #3898, opened 2026-07-17
add history cache to htmax but make it opt in via config

MichaelWest22:
## Description
htmax can include the history cache extension if we set it to default to disabled like this.  We can add -htmax to the version data so that the cache extension can detect this version and disable the cache extension by default then.  just need to add documentation on how to enable it via a meta tag and then users can opt in if they want

We should consider hx-ptag, hx-upsert and hx-alpine-compat as safe extensions we can easily add to htmax as well.
Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue


1cg:
Thank you!

PR #3899, opened 2026-07-18
Fix boosted history scroll restoration

scriptogre:
## Bug

`hx-boost`ed back/forward navigation loses scroll position.

The browser restores scroll against the old DOM. The position clamps before htmx swaps the restored page.

## Reproduce

1. Open https://four.htmx.org/reference.
2. Scroll to the bottom.
3. Press e.g. `<hx-partial>`.
4. Go back.

## Fix

Modern browsers:

1. Intercept traversal with the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API).
2. Swap the restored content.
3. Let the browser restore scroll.

Older browsers:

1. Save scroll when pushing the URL.
2. Swap the restored content.
3. Restore the saved position.

The fallback covers common back and forward cases.

Full native behavior needs more code (track scroll, update history entries, handle page lifecycle, restore after `popstate`, handle repeated visits).

## Prior work

PR #3704 only fixed the rare `history: "reload"` case and was more complex.

Analysis: https://gist.github.com/scriptogre/40c5c4af3ba971b7cf9be10603a80e7c

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1595,21 +1595,37 @@ var htmx = (() => {
         // History Support
         //============================================================================================
 
+        #supportsNavigationApi = typeof Navigation === 'function'
+            && window.navigation instanceof Navigation && !!NavigateEvent.prototype.intercept;
+
         __initHistoryHandling() {
             if (!this.config.history) return;
             if (!history.state) {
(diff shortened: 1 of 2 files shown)


1cg:
Superceded by 3905

PR #3907, opened 2026-07-21
Improve hx-live updates, debounce, and warnings

scriptogre:
## Description

Follow-up to #3891.

Make sure that in even more scenarios, we skip unnecessary work.

| Binding result | #3891 | This PR |
|---|:---:|:---:|
| Text and HTML | ✅ | ✅ |
| String and boolean attributes | ✅ | ✅ |
| Boolean and string class bindings | ✅ | ✅ |
| Normalized attribute values | | ✅ |
| ARIA attributes | | ✅ |
| `checked`, `selected`, and `value` | | ✅ |
| Object class bindings | | ✅ |
| Style bindings | | ✅ |

### Input debounce

Rename the `live.inputDebounceMs` config from #3891 to `live.inputDebounce`, and enable both numbers and HCON intervals.

```html
<meta name="htmx-config" content="live.inputDebounce:20ms"
```

### Recompute warning

Made the automatic warning appear based on main-thread cost rather than expression count.

Previously, you'd get warnings for 50x (cheap) recompute passes that took e.g. 0.1ms in total (negligible). 

Now, you get a warning when one single recompute pass exceeds 16ms (expensive), or recomputes consume over 100ms/s for 2 consecutive seconds.

### DIFF
--- a/src/ext/hx-live.js
+++ b/src/ext/hx-live.js
@@ -314,21 +314,6 @@
         });
     }
 
-    /**
-     * Toggle or cycle a class, ARIA attribute, or attribute on an element.
-     *
-     * @param {Element} elt - DOM element to mutate.
-     * @param {string} name - Class (`.foo`) or attribute name.
-     * @param {string|string[]} [values] - Cycle list (pipe-delimited string or array). Omit for binary flip.
-     *
-     * @example
-     * toggle('.active')                      // toggle class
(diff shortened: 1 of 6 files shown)


scriptogre:
This bundles too many changes.

I will bring back any useful parts as small, focused PRs.

PR #3909, opened 2026-07-21
implement navigation api and manual scroll restoration for history restore

MichaelWest22:
## Description
This is a PR to demonstrate what would be required to implement proper scroll restoration support including firefox navigation api support.  The core issue I found with firefox navigation api is that their implementation does not restore the previous page scroll position well like chomre/safari does.  In chrome/safari the browser captures the exact scroll position of the previous page and on back navigation it will bring it back after we restore the old content inside the navigation intercept handler.  

With firefox it has very inconsistent scroll behavior.  The popstate based scroll restoration was not perfect but works semi ok most of the time but it restores the scroll too early so navigating back from a short page to a long page sometimes cause scroll position errors.  But with navigation api enabled if we use the intercept like we do with chrome/safari it is even worse.  The core issue is that it seems to remember the last cached scroll position from the last refresh somehow instead of the last scroll position before you navigated forwards.  So often on loading a page it remembers position 0 and once you have scrolled and then triggered a boosted link and then hit back button it restores the 0 position back.  there is a feature where it tries to remember the last position you were on for this page from previous sessions so you get back to where you were last read up to and this feature is probably causing issues here somehow.   

After much attempts to find a simple solutuion the only way to fix this in firefox is to move to manual restore mode and then capture the scroll postion ourselves.  To reliably capture the last scroll position in all back/forward situations you need a scroll event listener and a way to store the scroll position based on the internal key id of the new navigation history element.  I found trying to push the scroll position onto the history.state just causes errors at times especially with safari and its rate limits.  So a simple in memory map solves this.

This does add a lot of code to handle this scroll edge cases.

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/hx-history-cache.js
+++ b/src/ext/hx-history-cache.js
@@ -198,7 +198,7 @@
         // Before core pushes/replaces, save the outgoing page
         htmx_before_history_update: (elt, detail) => {
             if (cfg().disable) return;
-            if (!_currentId) stampCurrentEntry();
+            stampCurrentEntry();
             saveCurrentPage();
         },
 
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1600,10 +1600,18 @@ var htmx = (() => {
             if (!history.state) {
                 history.replaceState({htmx: true}, '', location.href);
             }
-            if (window.navigation && !/firefox/i.test(navigator.userAgent)) {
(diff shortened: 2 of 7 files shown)


1cg:
Superceded by 3905

PR #3912, opened 2026-07-22
Restore form state before deferred head scripts

scriptogre:
## Description

During a cached history restore, `hx-head` appends deferred head scripts from `htmx:after:swap`. Form annotations are restored afterward, so those scripts observe default form values instead of the cached values.

Carry the existing restore detail through `htmx:history:cache:after:restore` and append cached deferred scripts there, after form state restoration. The existing `detail.item` field is preserved.

Corresponding issue: none

## Testing

Added an integration test that restores a cached input and verifies a deferred head script reads its restored value.

Chromium: 1,588 passed, 0 failed, 6 skipped.

## Checklist

* [x] I have targeted this PR against the correct htmx 4 development branch (`four-dev`)
* [x] This is a bugfix
* [x] I ran the test suite locally and verified that it succeeded

### DIFF
--- a/src/ext/hx-head.js
+++ b/src/ext/hx-head.js
@@ -182,6 +182,9 @@
                     detail._deferredHeadScripts = deferred;
                 });
             }
+        },
+        htmx_history_cache_after_restore: (elt, detail) => {
+            for (const node of detail._deferredHeadScripts || []) appendNode(node)
         }
     })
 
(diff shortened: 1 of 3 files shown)


MichaelWest22 (approved):
Yeah this fix should be fine.  I think deferred head scripts that depend on exact input values will be a very narrow edge case in practice but may as well fix it anyway.

PR #3913, opened 2026-07-22
Respect disabled history during swaps

scriptogre:
## Description

When `htmx.config.history` is `false`, swaps still emit history update events around a no-op URL update. The history cache extension handles those events and saves the outgoing page despite history being disabled.

Return before resolving or emitting a history update when history is disabled.

Corresponding issue: none

## Testing

Added an integration test that performs an `hx-push-url` request with history disabled and verifies the history cache stays empty.

Chromium: 1,588 passed, 0 failed, 6 skipped.

## Checklist

* [x] I have targeted this PR against the correct htmx 4 development branch (`four-dev`)
* [x] This is a bugfix
* [x] I ran the test suite locally and verified that it succeeded

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1697,6 +1697,7 @@ var htmx = (() => {
         }
 
         __handleHistoryUpdate(ctx) {
+            if (!this.config.history) return;
             let action = this.__resolveHistoryAction(ctx);
             if (!action) return;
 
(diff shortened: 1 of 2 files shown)


MichaelWest22 (approved):
looks good

1cg:
Thank you!

PR #3918, opened 2026-07-24
Improve view transitions extension point

MichaelWest22:
## Description
Add an async extension point for view transitions by allowing the wrapping of the task object

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2291,9 +2291,9 @@ var htmx = (() => {
 
             try {
                 if (document.startViewTransition) {
-                    this.__trigger(document, "htmx:before:viewTransition", {task})
-                    await document.startViewTransition(task).finished;
-                    this.__trigger(document, "htmx:after:viewTransition", {task})
+                    let detail = { task, transition: () => document.startViewTransition(task).finished };
+                    this.__triggerExtensions(document, "htmx:before:viewTransition", detail);
+                    await detail.transition();
                 } else {
                     await task();
                 }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -2291,9 +2291,10 @@ var htmx = (() => {
 
             try {
                 if (document.startViewTransition) {
-                    let detail = { task, transition: () => document.startViewTransition(task).finished };
-                    this.__triggerExtensions(document, "htmx:before:viewTransition", detail);
-                    await detail.transition();
+                    let detail = {task};
+                    this.__trigger(document, "htmx:before:viewTransition", detail)
+                    await document.startViewTransition(detail.task).finished;
+                    this.__trigger(document, "htmx:after:viewTransition", detail)
                 } else {
                     await task();
                 }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1294,7 +1294,7 @@ var htmx = (() => {
                             await this.__insertContent(task, false)
                         }
                     }
-                    swapPromises.push(this.__submitTransitionTask(tasksWrapper));
+                    swapPromises.push(this.__submitTransitionTask(tasksWrapper, ctx));
                 }
 
                 await Promise.all(swapPromises);
@@ -2271,10 +2271,10 @@ var htmx = (() => {
             }
         }
 
-        __submitTransitionTask(task) {
+        __submitTransitionTask(task, ctx) {
             return new Promise((resolve) => {
                 this.#transitionQueue ||= [];
-                this.#transitionQueue.push({ task, resolve });
+                this.#transitionQueue.push({ task, resolve, ctx });
                 if (!this.#processingTransition) {
                     this.__processTransitionQueue();
                 }
@@ -2287,14 +2287,14 @@ var htmx = (() => {
             }
 
             this.#processingTransition = true;
-            let { task, resolve } = this.#transitionQueue.shift();
+            let { task, resolve, ctx } = this.#transitionQueue.shift();
 
             try {
                 if (document.startViewTransition) {
-                    let detail = {task};
-                    this.__trigger(document, "htmx:before:viewTransition", detail)
+                    let detail = {task, ctx};
+                    this.__trigger(ctx.sourceElement, "htmx:before:viewTransition", detail)
                     await document.startViewTransition(detail.task).finished;
-                    this.__trigger(document, "htmx:after:viewTransition", detail)
+                    this.__trigger(ctx.sourceElement, "htmx:after:viewTransition", detail)
                 } else {
                     await task();
                 }


bkuhlmann:
Thanks for doing this. :bow: I would need to test this with [Terminus](https://alchemists.io/projects/terminus) and [htmx Slide](https://alchemists.io/projects/htmx-slide) to see if this finally allows me to dynamically swap out the desired View Transition at runtime. I'm still traveling back from Big Sky Dev Con but should be able to investigate this more at the end of this week.

MichaelWest22:
@bkuhlmann Just updated my change to include a useful sourceElement to the triggered event and the full internal ctx object which you can use to make conditional decisions on what to change in the transition.  You can perform an action before the transition starts and in the before event update the event.detail.task which is an async function to be your own async function that wraps the original task function allowing you to run some custom code before or after the swap inside the transition call.  There is then an after event where you could do another action if needed.  I think the after event is likely not needed in practice so we could probably remove this.  Let me know if you can think of any use for the after event so we know if its worth keeping or not.

bkuhlmann:
Michael, looks like Carson merged this so will switch to using htmx 4.0.0. Can log an issue if anything crops up.

MichaelWest22:
@bkuhlmann I wonder if changing the events to add the viewTransition object into the first event would help.  Then you could tap into .ready promise to customize the transition in JS or skip and create your own transition from scratch.  We would just await the event details transition object afterwords which could be reassigned if needed.  Be good to know if this was useful or not

bkuhlmann:
Could you provide an example of usage (or I guess I could use a feature branch)? I'd need to tinker with this via htmx Slide to see if what you suggest works.

MichaelWest22:
> Could you provide an example of usage (or I guess I could use a feature branch)? I'd need to tinker with this via htmx Slide to see if what you suggest works.
see the changes in #4000 

https://github.com/MichaelWest22/htmx/blob/660a3996437f9b6f2c393f8fc9cefd112c9e0c66/src/htmx.js has this modified htmx.js you could test with.

```JS
                    let detail = {task, ctx, transition: document.startViewTransition(() => detail.task())};
                    this.__trigger(ctx.sourceElement, "htmx:before:viewTransition", detail);
                    await detail.transition?.finished;
                    this.__trigger(ctx.sourceElement, "htmx:after:viewTransition", detail);
```

is the new code change.  you can still override detail.task as you did before to wrap it in a different before and after.  but you can now also use the transistion object from the event with something like:

```
htmx.on("htmx:before:viewTransition", ({ detail }) => {
    detail.transition.ready.then(() => {
        const isMobile = window.matchMedia("(max-width: 768px)").matches;
        const isForward = navigation.currentEntry.index > navigation.entries().indexOf(navigation.currentEntry) - 1;

        document.documentElement.animate(
            {
                clipPath: isMobile
                    ? ['inset(0 0 100% 0)', 'inset(0 0 0% 0)']
                    : ['inset(0 100% 0 0)', 'inset(0 0% 0 0)'],
                easing: 'cubic-bezier(0.4, 0, 0.2, 1)'
            },
            {
                duration: 350,
                pseudoElement: isForward
                    ? '::view-transition-new(root)'
                    : '::view-transition-old(root)'
            }
        );
    });
});

```

bkuhlmann:
Thanks. :bow: Give a me a couple days as I'm upgrading my entire stack to htmx 4.0.0. I'll respond back as soon as I have something for you.

bkuhlmann:
**Status Update**: I'm 2/3rds complete on 4.0.0 upgrade work. I'm tackling the above this week so hope to have a better perspective on this then.

bkuhlmann:
OK, I've been working on converting my [htmx Slide](https://alchemists.io/projects/htmx-slide) extension to htmx 4.0.0 and am able to see that my slides work properly but none of the CSS View Transitions fire. Here's my [release branch](https://github.com/bkuhlmann/htmx-slide/compare/release). Only the last commit, *Updated to htmx 4.0.0*, is of interest since it shows the diff of converting from htmx 2.0.0 to 4.0.0.

I tried the following extension hooks:

* `htmx_before_viewTransition`: This isn't documented and doesn't work.
* `htmx_before_swap`: This one fires but none of my styles are being applied properly. I'm not sure this is even the right hook.

I tried deconstructing your code snippets but the properties and functions in your example don't seem to exist on the `detail` object provided via an extension hook. I can see this being powerful in being able to manipulate the transition directly if I could get a hold of the object. 

What am I missing?

MichaelWest22:
```
<meta name="htmx-config" content='{"transitions":true}'>
```
Can you try with This config at the top of you index html?

I think the select oob's are not picking up the main swap style transition setting properly.  

Also can you please try manually downloading https://github.com/MichaelWest22/htmx/blob/viewTransitionExt2/src/htmx.js into your local project and linking locally to test the new change I made?

There is also another issue i found where the default behaviour for empty main swaps with just oob's is to not swap now and your swaps responses are empty with just select oob's which now safely prevent the probably invalid empty main swap now.  you can set a new config:

```
<meta name="htmx-config" content='{"transitions":true, "allowEmptySwapAfterOOB":true}'>
```
That will make it do the empty main swaps again for you.  And there is another small bug where if you prevent the main swap then the main swaps transition:true manual override does not apply so it won't upgrade you OOB's to be transition trues.  And you can't set selectOOB's swap style to make them transitions.  We do now support OOB's and hx-partials being manually included or excluded from transitions with swap modifiers but not for selectOOB.  Moving your server to return <hx-partial> tags is my recommendation as it makes your server code super clear and simple. and the requesting element doesn't have to have so many complex attributes just right.  but for your case just setting the config to turn transitions on makes the most sense first.

bkuhlmann:
Thanks. OK, so adding `<meta name="htmx-config" content='{"transitions":true}'>` enabled the transitions but only for *forward* transitions. The *backward* transitions almost work but then start to fail. Part of the reason is that with htmx 2.0.0, I could do this:

1. Use `onEvent` to check for `name` and `event`.
2. Then check if the `name` was `htmx:beforeTransition` or `htmx:oobBeforeSwap`.
3. `htmx:beforeTransition` allowed me to apply *forward* CSS View Transitions while `htmx:oobBeforeSwap` allowed me to apply *backward* transitions.

I also tested using your fix (my feature branch is updated to include it if you need it). :bulb: Running `bin/demo` wires up everything for quick visual verification. With your changes, I was getting a lot of these errors in the console:

``` 
Uncaught SyntaxError: Invalid or unexpected token (at VM342:3:14)
    at new Function (<anonymous>)
    at Htmx.__executeJavaScript (htmx.js:908:24)
    at HTMLBodyElement.<anonymous> (htmx.js:803:35)
```

I will test using `hx-partial` tomorrow and get back to you (your morning is my evening).

bkuhlmann:
OK, finally got this working (and my [release](https://github.com/bkuhlmann/htmx-slide/compare/release) branch is rebased). A few notes:

* I wasn't able to make your code work within an extension.
* Wrapping all response fragments in `hx-partial` made it easier to configure what should be swapped without having to rely on `selectOOB`. This also made it easier to apply which CSS View Transition style is needed for the elements. I had to use `detail.tasks` and iterate through each fragment to find the right one and then apply the correct style. This is also done via the `htmx_before_swap` event.
* The syntax error, mentioned above, is due to htmx 4.0.0 no longer understanding a key up trigger for `]`. I had to switch to using `keyup[event.code=='BracketRight']` in order to make this work.

I'm still not sure if I'm using htmx the most effectively because so much has changed with 4.0.0 but specs are passing...but I would like a better way to tap into the CSS View Transitions event via an extension, though.

PR #3919, opened 2026-07-25
Allows hx-live to cache JS eval functions

MichaelWest22:
## Description
I found that we can get around a 23x speed improvment to hx-live expression evaluations by updating executeJavaScript function to return a function instead of the value which then allows hx-live to cache this function in its run closure

Corresponding issue:

## Testing
Tested performance and found it went from around 1.15s for 20,000 live bindings to 50ms.  No real function changes so tests all pass.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "check:content": "python3 src/scripts/content/check.py",
     "upgrade-check": "python3 src/scripts/upgrade-check.py",
     "upgrade-check:test": "python3 src/scripts/upgrade-check.py test/manual/upgrade/",
+    "benchmark:hx-live": "node test/benchmarks/hx-live.mjs",
     "test": "npm run test:chrome",
     "test:chrome": "web-test-runner --browsers chromium --config test/web-test-runner.config.mjs --playwright",
     "test:firefox": "web-test-runner --browsers firefox --config test/web-test-runner.config.mjs --playwright",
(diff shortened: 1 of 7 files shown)


scriptogre:
@MichaelWest22 what do you think about making compilation opt-in?

```js
__executeJavaScript(thisArg, obj, code, expression = true, isAsync = true, compile = false)

return compile ? () => func.call(thisArg, ...values) : func.call(thisArg, ...values)
```

Then hx-live can cache it inside the existing `try`:

```js
exec ||= api.executeJavaScript(elt, { debounce }, code, true, isAsync, true)
```

This keeps the current return value for other extensions and keeps the hx-live perf win. I tried it here: https://github.com/scriptogre/htmx/commit/bea4514e

scriptogre:
I also pushed a benchmark that covers text, HTML, attributes, properties, classes, styles, and ARIA. Changed update p50:

```text
100 expressions:   4.1ms ->  0.4ms  (10.3x)
500 expressions:  21.6ms ->  4.7ms   (4.6x)
1000 expressions: 50.2ms -> 16.0ms   (3.1x)
```

Compare two or more worktrees with named cases:

```sh
bun run benchmark:hx-live --   --case before=../htmx-base   --case compiled=.
```

1cg:
Thank you Latent!

Issue #3920, opened 2026-07-26
hx-sse:close does not work when there is no data


aytekinar:
I am using/testing HTMX `v4.0.0-beta6` right now, and I have the following test file (runnable via `uv`):

```python
#! /usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "sse-starlette~=3.0",
#     "starlette~=1.0",
#     "uvicorn[standard]~=0.51.0",
# ]
# ///
import asyncio
import collections.abc

import sse_starlette
import starlette.applications
import starlette.requests
import starlette.responses
import starlette.routing
import uvicorn

CONTENT = """
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
    <script src="https://cdn.jsdelivr.net/npm/htmx.org@4.0.0-beta6/dist/htmx.min.js" defer></script>
    <script src="https://cdn.jsdelivr.net/npm/htmx.org@4.0.0-beta6/dist/ext/hx-sse.min.js" defer></script>
    <title>HTMX SSE Playground</title>
  </head>
  <body>
    <output aria-live="polite" hx-sse:connect="/sse" hx-sse:close="done"></output>
  </body>
</html>
"""


async def homepage(_: starlette.requests.Request) -> starlette.responses.HTMLResponse:
    return starlette.responses.HTMLResponse(content=CONTENT)


async def sse_endpoint(
    _: starlette.requests.Request,
) -> sse_starlette.EventSourceResponse:
    async def values() -> collections.abc.AsyncIterable[sse_starlette.ServerSentEvent]:
        for idx in range(10):
            yield sse_starlette.ServerSentEvent(data=f"data {idx}")
            await asyncio.sleep(1)

        # if data=None, `hx-sse:close` does not work
        # yield sse_starlette.ServerSentEvent(event="done")
        yield sse_starlette.ServerSentEvent(event="done", data="")

    return sse_starlette.EventSourceResponse(content=values())


app = starlette.applications.Starlette(
    routes=[
        starlette.routing.Route("/sse", sse_endpoint, methods=["GET"]),
        starlette.routing.Route("/", homepage, methods=["GET"]),
    ]
)

if __name__ == "__main__":
    uvicorn.run(app, host="127.0.0.1", port=8080)
```

[...]

PR #3923, opened 2026-07-28
handle sse close with no event data

MichaelWest22:
## Description
sse extension does not handle close events well with no data.  reworked the logic so that events even with no data still flow though as you would expect.  Also last id update logic can be improved a little to always count the last sent id properly and only set last id when there is an id.

Corresponding issue:
#3920 

## Testing
Added a test for no data close event and removed a test that showed canceling a id update would not increment the last id which could cause invalid resending of data.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/hx-sse.js
+++ b/src/ext/hx-sse.js
@@ -59,8 +59,8 @@
 
                 for (let line of lines) {
                     if (!line) {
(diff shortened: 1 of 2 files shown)


1cg:
Thank you!

PR #3924, opened 2026-07-28
support extended css selectors in hx-partial

MichaelWest22:
## Description
Found we can easily support extended selectors in hx-partial targeting.  This allows you to target elements relative to the triggering source element using all the htmx enhanced selectors like closest, previous, next etc.

Also fixed a bug in the same code where if you only have hx-partial's and they don't find valid targets it would not do the expected behaviour of blocking the empty main swap.  we just need to make sure it always emits a partial task so the auto empty detection code works.

Corresponding issue:

## Testing
added a basic test

## Checklist

  source changes)
  approved via an issue


scriptogre:
I think it'd be more sane if `<hx-partial>` and `hx-swap-oob` behaved the same.

Can we combine this with #3926 and close one of the PRs?

1cg:
Thank you Latent!

PR #3926, opened 2026-07-29
Align `<hx-partial>` and `hx-swap-oob` behavior on empty main swap

scriptogre:
## Why

Main target:

```html
<div id="main-target">Original</div>
```

`hx-swap-oob` response:

```html
<div id="oob-target" hx-swap-oob="true">Updated</div>
```

Result:

```html
<div id="main-target"></div>
```

`<hx-partial>` response:

```html
<hx-partial hx-target="#partial-target">Updated</hx-partial>
```

Result:

```html
<div id="main-target">Original</div>
```

`swapEmpty` currently has three states:

- `true` clears the main target.
- `false` keeps the main target.
- `undefined` clears it for `hx-swap-oob`, but keeps it for `<hx-partial>`.

The special `undefined` state is confusing.

## Change

Default `defaultSwapEmpty` to `true`, so both responses clear the main target (like `hx-swap-oob` does now), so they at least work consistently.

### DIFF
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -131,13 +131,10 @@ export interface HtmxConfig {
   metaCharacter?: string;
   /**
    * Whether an empty response body performs the main swap.
-   * - `true` — swap (clears target)
-   * - `false` — skip swap
-   * - `undefined` — swap unless response contained only `<hx-partial>` elements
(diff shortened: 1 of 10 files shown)


1cg:
Per discussion in chat, going to keep the asymmetry just needs docs

PR #3929, opened 2026-07-30
Standardize `hx-sse` events, config, and legacy support

scriptogre:
## Why

Before:

```text
htmx:before:sse:message
htmx:sse:error
```

After:

```text
htmx:sse:before:message
htmx:sse:error
```

## Changes

1. Rename events to `htmx:sse:<timing>:<noun>`.
2. Add support for legacy `sse-close` (previously just `sse-connect`)
3. Warn for `sse-connect`, `sse-close`, and (now removed) `sse-swap`.
3. Resolve config in this order:

   ```text
   built-in defaults → htmx.config.sse → inherited hx-config
   ```


1cg:
Thanks Christian!

PR #3930, opened 2026-07-30
Accept intervals for `hx-live` input debounce

scriptogre:
## Why

The setting accepts only a number of milliseconds:

```ts
inputDebounceMs: 20
```

This change also accepts htmx interval strings:

```html
<meta name="htmx-config" content="live.inputDebounce:20ms">
```

## Changes

1. Rename `live.inputDebounceMs` to `live.inputDebounce`.
2. Accept a number or an interval string:

   ```ts
   inputDebounce: 20
   inputDebounce: '20ms'
   ```


1cg:
THank you!

PR #3932, opened 2026-07-30
Warn when `hx-live` expressions take over `16ms`

scriptogre:
## Why

After a page change, `hx-live` runs all live expressions once.

The current warning counts how often this happens. It does not measure the time:

```text
51 runs at 0.1ms each -> warning
 1 run at 100ms       -> no warning
```

## Change

Measure each run instead. Warn once if it takes more than `16ms`:

```text
htmx: hx-live expressions took 18.4ms.
```

The warning does not stop the expressions.

Merges cleanly before or after #3919, #3928, #3930, and #3931.

### DIFF
--- a/src/ext/hx-live.js
+++ b/src/ext/hx-live.js
@@ -13,8 +13,6 @@
     let recomputeBound = null;
     let inputBound = null;
     let swaps = 0;
-    let i = 0;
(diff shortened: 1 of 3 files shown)


1cg:
Thank you!

PR #3933, opened 2026-07-30
Fix `hx-live` form reset state

scriptogre:
## Why

A `checked` binding can return `0`:

```html
<input type="checkbox" :checked="items.length">
```

The input is unchecked, but `hx-live` also writes `checked="0"`. HTML treats the attribute as true. A form reset then checks the input:

```text
Before reset: unchecked
After reset:  checked
```

A `selected` binding has the same problem.

## Change

Convert `checked` and `selected` results to booleans. Use the boolean to set the DOM property and add or remove the HTML attribute.

```text
Before reset: unchecked
After reset:  unchecked
```

### DIFF
--- a/src/ext/hx-live.js
+++ b/src/ext/hx-live.js
@@ -134,7 +134,11 @@
                     : (value ? 'true' : 'false');
                 e.setAttribute(name, attrVal);
             } else if (isPropAttr) {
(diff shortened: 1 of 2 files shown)


1cg:
Thank you!

PR #3934, opened 2026-07-31
Fix overlapping `hx-live` styles

scriptogre:
## Why

```html
<div data-all="true" :style="data.all
    ? 'border: 1px solid red'
    : 'border-left-color: green'">
</div>
```

After `data-all` becomes `false`:

```text
Expected: border-left-color: green
Actual:   no border style
```

## Change

Remove old managed styles before writing new styles.

```text
Result: border-left-color: green
```


1cg:
Thank you!

PR #3935, opened 2026-07-31
Fix hx-sse reader-cancel race that produces an unhandled rejection

marciomazza:
## Description

`parseSSE`'s `finally` released the reader's lock, but `connection.reader` was only nulled one
microtask later (via the `for await...of` loop's async-generator close protocol). Any concurrent
`connection.reader.cancel()` call (pauseOnBackground's `visibilitychange` handler, or `cleanup()`)
landing in that one-microtask window canceled an already-released reader, which per the Streams
spec rejects instead of throwing synchronously — surfacing as an unhandled promise rejection.

Fix: pass `connection` into `parseSSE` (instead of a bare `reader`) and null `connection.reader` in
the same synchronous `finally` block as `reader.releaseLock()` — no `await` in between — closing
the race window at its source instead of guarding each `.cancel()` call site with
`.catch(() => {})`. The now-redundant `connection.reader = null` after the `for await` loop in
`handleSSEResponse` is removed.

## Testing

Added a regression test, `cancelling an already-released reader does not produce an unhandled
rejection`, that deterministically lands in the race window: it uses a hand-rolled reader so the
test's concurrent `visibilitychange` trigger can be chained onto the exact same promise the
extension's internal `await reader.read()` is waiting on, instead of relying on timing coincidences.

Ran the full suite locally via `web-test-runner` (Playwright/Chromium): 1606 passed, 0 failed, 6
skipped. Also confirmed the new test fails against the pre-fix code and passes after the fix.

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/ext/hx-sse.js
+++ b/src/ext/hx-sse.js
@@ -32,7 +32,9 @@
     // SSE PARSER
     // ========================================
 
-    async function* parseSSE(reader, lastEventId = '') {
+    async function* parseSSE(connection) {
+        let reader = connection.reader;
+        let lastEventId = connection.lastEventId;
         let decoder = new TextDecoder();
         let buffer = '';
         let hasData = false;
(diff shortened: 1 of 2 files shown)


1cg:
Thank you!

PR #3938, opened 2026-08-01
Namespace `hx-head` lifecycle events

scriptogre:
Group the `hx-head` lifecycle events under the extension namespace.

- `htmx:before:head:merge` → `htmx:head:before:merge`
- `htmx:before:head:add` → `htmx:head:before:add`
- `htmx:before:head:remove` → `htmx:head:before:remove`
- `htmx:after:head:merge` → `htmx:head:after:merge`


1cg:
Looks good, we should probably document the event name changes on the extension page

scriptogre:
@1cg Already did!

PR #3939, opened 2026-08-01
Fix `HX-Location` replace option

scriptogre:
`replace` is ignored because `HX-Location` always adds the default `push` option:

```http
HX-Location: {"path":"/dashboard","replace":"/dashboard"}
```

```text
Expected: replace the current history entry
Actual:   push a new history entry
```

The same format check misclassifies URLs with commas and minimal HCON:

| Value | Expected | Actual |
| --- | --- | --- |
| `/files/a,b` | Plain URL | HCON |
| `path:/search` | HCON | Plain URL |

## Changes

1. Default to `push: 'true'` only when `push` and `replace` are absent.

2. Use parsed `path` to distinguish HCON from a plain URL:

   ```text
   /files/a,b    -> /files/a,b
   path:/search  -> { path: '/search' }
   ```

3. Rewrite the `HX-Location` reference page.

4. Add regression tests.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -685,12 +685,13 @@ var htmx = (() => {
             }
             if (ctx.hx.location) { // HX-Location
                 let path = ctx.hx.location, opts = {};
-                if (path[0] === '{' || /[\s,]/.test(path)) {
-                    opts = HCON.parse(path);
+                let parsed = HCON.parse(path);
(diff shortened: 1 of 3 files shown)


MichaelWest22 (approved):
seems good to me

1cg:
Thanks!

PR #3940, opened 2026-08-01
Rename `htmx.ajax()` context to options

scriptogre:
The third `htmx.ajax()` argument contains options used to build the request context. Calling both objects `context` makes them difficult to distinguish.

Before:

```text
htmx.ajax(method, url, context)
HtmxAjaxContext
```

After:

```text
htmx.ajax(method, url, options)
HtmxAjaxOptions
```

## Changes

1. Rename the runtime argument from `context` to `options`.
2. Rename `HtmxAjaxContext` to `HtmxAjaxOptions`.
3. Update docs and test names.
4. Link `options` to the request context shape.

JavaScript calls and runtime behavior do not change.

### DIFF
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -536,8 +536,8 @@ export interface HtmxEventMap {
 
 export type HtmxEvent<K extends keyof HtmxEventMap> = CustomEvent<HtmxEventMap[K]>;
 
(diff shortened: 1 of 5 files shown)


1cg:
Thank you!  Great catch

PR #3942, opened 2026-08-02
move to allowEmptySwapAfterOOB config instead of defaultSwapEmpty

MichaelWest22:
## Description
Change the empty swap behavior so that oob work the same as hx-partial do now.  when only oob content is in the response it will now not swap the main target with an empty response like it did before.  This could mean that a response that is trying to blank out the requesting form with an empty response and place the form's completed action into another location with hx-swap-oob it will now not remove the form as it did before.  Ther is a new config item allowEmptySwapAfterOOB  which can be configured to return oob behavior back to the old way where empty contents will swap after removing oob's.  hx-partial is set to always prevent empty swap if present and you have to use swapEmpty swap modifier to alter this.

Removed the defaultSwapEmpty global config because this config can cause unexpected behavior like blocking deliberately empty  responses from happening or breaking expected behavior with hx-partial only responses.  Its better to be more explicit and control the empty swap behavior with swap modifiers per request instead.

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue


1cg:
Thanks!

Issue #3937, opened 2026-07-31
[4.0.0-beta6] htmx.trigger('#id','htmx:abort') works only for requests made using hx-* attribute


devseppala:
I tried 4.0.0-beta6 on a small hobby project I'm developing and soon end up with issues while trying to abort a request.  Turns out that htmx.trigger('#id','htmx:abort') does not work if the reguest was made from javascript and not directly from a attribute hx-get,hx-post... attribute. 

Below is a PHP/html example that shows the problem.
* update, keep the developer tools network tab open to see the issue.

```
----- /abort/sleep30s.php-----
<?php sleep(30); ?>
<div>Hello World</div>

-----/abort/htmx.html-----
<html>
  <head>
    <!--<script src="https://cdn.jsdelivr.net/npm/htmx.org@2.0.10/dist/htmx.min.js"></script>-->
    <script src="https://cdn.jsdelivr.net/npm/htmx.org@4.0.0-beta6/dist/htmx.min.js"></script>
  </head>

  <body>
    <div>
      <button id="clickme" hx-post="/abort/sleep30s.php" hx-swap="outerHTML">Click Me</button>
      <button onclick="htmx.trigger('#clickme','htmx:abort')">abort works on v2 & v4</button>
    </div>
    <div>
      <button id="clickmeJs2" onclick="htmx.ajax('GET','/abort/sleep30s.php',{target:'#clickmeJs2',source:'#clickmeJs2'});">Click Me</button>
      <button onclick="htmx.trigger('#clickmeJs2','htmx:abort')">abort works on v2 not v4</button>
    </div>
  </body>
</html>
```

MichaelWest22:
Put up a PR to improve the abort handling

devseppala:
Thank you Michael for devoting your time to fix this issue! Hopefully your PR gets merged soon.

Just started thinking if there are any other relevant events that are not listened in v4, but I don't see any. I will just leave this comment here if something comes to your mind.

PR #3944, opened 2026-08-02
allow abort event on htmx.ajax requests

MichaelWest22:
## Description
htmx.ajax requests can not be aborted with htmx:abort event because this listener is only setup on active htmx elements and not on elements used as the source of a manual ajax api call.  To allow aborting of these we need to lazy load the abort listener during the issue request instead.

Corresponding issue:
#3937

## Testing
Added a ajax abort test

## Checklist

  source changes)
  approved via an issue


1cg:
Thanks Latent!

Issue #3948, opened 2026-08-04
[4.0.0-beta6] hx-preload fails to reuse cached GET when hx-get uses relative URLs


woutergrimme:
## Bug report

Using htmx 4.0.0-beta6 with the hx-preload extension.

When preloading a link with a relative `hx-get` URL, the preload request is made successfully, but the cached response is not reused on click.

Example:

```html
<a
    hx-get="https://domain.com/some/url"
    hx-preload="mouseover timeout:10s">
    Some link
</a>
```

Steps to reproduce:
Hover the link.
Wait for the preload request to complete.
Click the link.

Expected:
The click should reuse the preloaded response.

Actual:
A second GET request is sent.

Debugging showed:
```
cached action : https://domain.com/some/url
current action: /some/url
```
The comparison in hx-preload.js:
```
elt._htmx.preload.action === ctx.request.action
```
This fails because the URLs are semantically identical but different strings.

A possible fix (tested locally):
```js
elt._htmx.preload.action === new URL(ctx.request.action, location.href).href
```
This normalizes the URL before comparison.

Environment:

htmx: 4.0.0-beta6
hx-preload: 4.0.0-beta6
Browser: Chrome/Firefox

MichaelWest22:
Thanks for finding this.  I've pushed up a fix that reuses the core htmx logic to calculate the action url to compare with which seems to be a more reliable fix from my testing.  Would be good if you can test yourself with the four-dev branch src/htmx.js version to confirm it fixes your issue.

PR #3949, opened 2026-08-04
Fix preload action URL comparison in hx-preload.js

woutergrimme:
## Description
Fix `hx-preload` failing to reuse preloaded GET responses when the preload URL and the later htmx request URL use different URL formats.

The preload extension stores the fetched URL as an absolute URL, but the later htmx request can contain a relative URL.

Corresponding issue:
https://github.com/bigskysoftware/htmx/issues/3948

## Testing
Tested manually:
- Hovered over an `hx-preload="mouseover timeout:10s"` link.
- Waited for the preload request to complete.
- Clicked the link.
- Verified that the click reused the preload and did not send a second request.

## Checklist

* [x] I have targeted this PR against the correct branch
  approved via an issue

### DIFF
--- a/src/ext/hx-preload.js
+++ b/src/ext/hx-preload.js
@@ -78,7 +78,7 @@
         htmx_before_request: (elt, detail) => {
             let {ctx} = detail;
             if (elt._htmx?.preload &&
-                elt._htmx.preload.action === ctx.request.action &&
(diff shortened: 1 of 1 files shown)


MichaelWest22:
Thanks i've pushed another fix for this now.  Had to use the same logic as core htmx does to normalize the action before we compare it as this proposed fix was not working with edge case testing

PR #3950, opened 2026-08-05
add scanning for non ascii characters in js src files

MichaelWest22:
## Description
For maximum compatibility in case any js file is ever loaded without utf-8 support somehow we should avoid the use of non ascii characters and comments in the source code itself.  And we can add a check script to validate this going forward so we can clean up any more that slip in.

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue


1cg:
Thank you Latent!

PR #3953, opened 2026-08-08
expose initialize function for async and streaming use cases

MichaelWest22:
## Description
For streaming use cases like my hx-streaming extension we need a way to get htmx to initialize its page handling and history support before domContentLoaded event fires.  The ability to trigger htmx init has been a regular request before like in #2264 and many linked issues before. 

Here i've made it a new public function so it can be called by anyone and it ensures history support is working and reprocesses the body to make sure it is live when you call it.  We could move this function into internalAPI instead if we want to make it extension only instead.

Corresponding issue:
#2264

## Testing
Added basic tests

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.d.ts
+++ b/src/htmx.d.ts
@@ -613,6 +613,13 @@ export interface Htmx {
    */
   onLoad(callback: (elt: Element) => void): void;
   /**
+   * Sets up history handling and processes `document.body`.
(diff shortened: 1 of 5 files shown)


1cg:
OK, just for you :)

PR #3954, opened 2026-08-10
Add HX-Request header usage documentation

HarshAwasth-i:
Added documentation explaining how the HX-Request header can be used to differentiate htmx requests from normal requests and decide whether to return full pages or partial HTML fragments.


1cg:
this material is covered elsewhere

PR #3958, opened 2026-08-12
rename hx-optimistic to hx-pending

MichaelWest22:
## Description
hx-optimistic extension can be used for more than just fully optimistic UI's so we should rename it to something that shows its true purpose of allowing the user to insert pending/placeholder content in place of the real response while the request is in flight.  Still need to support the fully optimistic use case as well but make it clear that it can be used for a range of different cases and that you can use htmx indicators instead 

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue


1cg:
Thank you

PR #3960, opened 2026-08-14
Add `connection` to `hx-sse` and `hx-multipart` event details

scriptogre:
PR #3910 exposes the connection on `hx-ws` message and error events.

This aligns `hx-sse` and `hx-multipart`:

- SSE message events include `connection`.
- Multipart part events include `connection`.
- Stream and reconnect errors include `connection`.
- SSE setup errors keep `url` because no connection exists yet.

The event's connection is the same object exposed by its before/after connection events.


1cg:
Thank you @scriptogre!

PR #3962, opened 2026-08-15
add hx-query to support the new RFC method

MichaelWest22:
## Description
We can add support for the new http method QUERY that now has an official RFC.  We just need to add hx-query as a new option as we already support it via hx-action/hx-method.  while query has not yet been fully adopted by everyone yet the browsers already natively support new methods like this and the real work left implementing it fully is in the CDN and caching layers so it is safe for us to adopt it early.

Corresponding issue:

## Testing

## Checklist

  source changes)
  approved via an issue


1cg:
Thanks!

PR #3976, opened 2026-08-26
Document the shared WebSocket connection pattern

stukennedy:
Follow-up to the review discussion on #3910, delivering the docs note agreed there ("one connection element, multiple senders attach to it").

**This branch is based on #3910's head (`scriptogre:4.0/improve-ws-ext`), so it includes those commits — only the final docs commit (cf71cd4c) is new here.** If #3910 merges first, this rebases down to a 3-line docs diff.

The docs already state the ownership rule ("Separate elements open separate connections, even when they use the same URL") and show the shared pattern in **Use Shared Connections**, but nothing connected the two — a reader could learn the rule without learning how to avoid accidentally fanning out sockets to the same URL.

Three small additions, no code changes:

- **Use Shared Connections** now opens with the ownership rule, framing the section as the way to share one socket.
- The `hx-ws:connect` reference cross-links from the ownership statement to that section.
- The beta → RC1 migration notes call out that pages relying on the beta's URL pooling now open several sockets, with the same cross-link.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


stukennedy:
Closing in favor of targeting #3910's branch directly (scriptogre/htmx#3) so it merges into that PR.

PR #3977, opened 2026-08-26
Add reset trigger and swap modifiers to reset the closest form

MichaelWest22:
## Description
Add new modifiers to swap and trigger that reset the closest form input values by just using hx-trigger="submit reset" or hx-swap="innerHTML reset".  Trigger form happens at trigger event just before it sends the request and the swap form happens after a valid swap happens which is safer if the request fails but is not as instant.

Corresponding issue:

## Testing
Added basic tests.

## Checklist

  source changes)
  approved via an issue


1cg:
Thank you!

PR #3979, opened 2026-08-27
docs: fix dead sinon fake-xhr link in README

1t1sCooL:
The sinon "fake XHR and server" link in the README returns **HTTP 404** — sinonjs.org was rebuilt and no longer hosts any `/releases/vN/fake-xhr-and-server/` pages (the functionality was split into the `nise` package and the doc has no live equivalent). Replaced with the Internet Archive snapshot of the referenced page (HTTP 200, "Fake XHR and server - Sinon.JS").


1cg:
Thank you!

PR #3982, opened 2026-08-27
fix: escape IDs in restorePreservedElements, selectOOB, and anchor scroll

n0vdd:
## Description

Three remaining call sites build a CSS selector from a raw element id without `CSS.escape()`, causing failures when the id contains CSS-special characters (`.`, `/`, `:`, etc.):

1. **`restorePreservedElements`** (line ~1521) — `find('#' + preservedElt.id)` → `querySelector('#' + id)`. On browsers with `moveBefore` (Chrome), htmx moves the preserved element into the pantry div during `handlePreservedElements`, then `restorePreservedElements` looks it back up by id to put it in place. A dotted id like `payment.parcelas` becomes the selector `#payment.parcelas` (id=payment AND class=parcelas), matches nothing, and the next line throws `TypeError: Cannot read properties of null (reading 'parentNode')`, aborting the swap.

2. **`selectOOB` handling** (line ~1929) — `fragment.querySelector('#' + id)`. Same pattern: the id from `hx-select-oob` is interpolated raw.

3. **Anchor scroll** (line ~2004) — `resolveTarget('#' + swapOptions.anchor)` → `find()` → `querySelector()`. An anchor id with special chars hits the same parse error.

The sibling call sites in `oobSwap` (fixed by #3304) and `handleAttributes` (fixed by #3752) already use `CSS.escape()` correctly — these three were missed.

### Repro (restorePreservedElements — the most visible one)

```html
<div hx-get="/update">
  <input id="payment.amount" hx-preserve value="typed-value">
</div>
```

On Chrome (moveBefore path): swap replaces the content, `handlePreservedElements` stashes the input in the pantry via `getElementById` (works fine — getElementById treats id as opaque), then `restorePreservedElements` tries `querySelector('#payment.amount')` → no match → TypeError on `.parentNode`. The preserved element is stranded in the pantry div and all subsequent swaps on the page may break.

On Firefox/Safari (replaceChild path): `handlePreservedElements` takes the else branch and calls `replaceChild` directly — `restorePreservedElements` is never called, so the bug is latent.

Corresponding issues: the same bug class as #1537 (oobSwap) and the settle-lookup in #3752, applied to the three remaining unescaped sites.

## Testing

- Added `handles hx-preserve on elements with dotted IDs` to `test/attributes/hx-preserve.js`
- Added `handles elements with IDs containing dots in hx-select-oob` to `test/attributes/hx-select-oob.js`
- Full test suite: 849 passed, 0 failed, 3 skipped (Chrome)

## Checklist

* [x] I have targeted this PR against the correct branch (`dev` for source changes)
* [x] This is a bugfix
* [x] I ran the test suite locally (`npm run test:chrome`) and verified that it succeeded

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1518,7 +1518,7 @@ var htmx = (function() {
     const pantry = find('#--htmx-preserve-pantry--')
     if (pantry) {
       for (const preservedElt of [...pantry.children]) {
-        const existingElement = find('#' + preservedElt.id)
+        const existingElement = find('#' + CSS.escape(preservedElt.id))
         // @ts-ignore - use proposed moveBefore feature
         existingElement.parentNode.moveBefore(preservedElt, existingElement)
         existingElement.remove()
@@ -1926,7 +1926,7 @@ var htmx = (function() {
               id = id.substring(1)
             }
             const oobValue = oobSelectValue[1] || 'true'
-            const oobElement = fragment.querySelector('#' + id)
+            const oobElement = fragment.querySelector('#' + CSS.escape(id))
             if (oobElement) {
               oobSwap(oobValue, oobElement, settleInfo, rootNode)
             }
@@ -2001,7 +2001,7 @@ var htmx = (function() {
         })
 
         if (swapOptions.anchor) {
-          const anchorTarget = asElement(resolveTarget('#' + swapOptions.anchor))
+          const anchorTarget = asElement(resolveTarget('#' + CSS.escape(swapOptions.anchor)))
           if (anchorTarget) {
             anchorTarget.scrollIntoView({ block: 'start', behavior: 'auto' })
           }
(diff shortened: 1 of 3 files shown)


1cg:
Thank you for digging into this, and for the clear repro on each of the three call sites.

We are going to leave this one to htmx 4 rather than change 2.x.

The reason is the `hx-select-oob` site. Escaping the id there locks the value to a literal id, which is the opposite of what #2561 asks for. htmx 4 already went the other way: it uses the value as a real selector and drops the `#` prefix entirely, so the escape is not needed and the general selector support people want is there.

```js
for (let spec of selectOOB.split(',')) {
    let [selector, oobValue = 'true'] = spec.split(/:(.*)/);
    for (let elt of fragment.querySelectorAll(selector)) {
        this.__createOOBTask(tasks, elt, oobValue, sourceElement);
    }
}
```

htmx 4 also uses `CSS.escape` where an id really is an id, in `__createOOBTask`.

Your analysis was correct and useful. Thank you!

PR #3993, opened 2026-08-30
Fix minor documentation error

remyzandwijk:
There is an error in the v4 documentation. This example is given:

```
<button hx-post="/clicked"
        hx-trigger="click"
        hx-target="#output"
        hx-swap="outerHTML">
    Click Me
</button>
<output id="output"></output>
```

The documentation states that the button is replaced when the button is clicked, but that is obviously not true. The complete ```<output id="output"></output>``` block is being replaced.


1cg:
Thank you!

PR #4017, opened 2026-09-03
fix renamed afterSettle event on upgrade-check.py

jorgecs10:
Fix bug on 
Upgrade check tool

## Description 
upgrade check tool was instructing to rename htmx:afterSettle event to htmx:after:swap instead of htmx:after:settle

## Checklist

* [ ✓] I have read the contribution guidelines
* [ ✓] I have targeted this PR against the correct branch (`master` for website changes, `dev` for
  source changes)
* [ ✗] This is either a bugfix, a documentation update, or a new feature that has been explicitly
  approved via an issue
* [ ✗] I ran the test suite locally (`npm run test`) and verified that it succeeded

### DIFF
--- a/src/scripts/upgrade-check.py
+++ b/src/scripts/upgrade-check.py
@@ -52,7 +52,7 @@ EVENT_RENAMES = {
     "htmx:afterOnLoad": "htmx:after:init",
     "htmx:afterProcessNode": "htmx:after:init",
     "htmx:afterRequest": "htmx:after:request",
-    "htmx:afterSettle": "htmx:after:swap",
+    "htmx:afterSettle": "htmx:after:settle",
     "htmx:afterSwap": "htmx:after:swap",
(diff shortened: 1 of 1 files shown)


MichaelWest22:
Thanks for the contribution! However, the fix isn't quite right here.

In htmx 2, htmx:afterSettle fired after the full swap-and-settle cycle completed. In htmx 4, the event model was redesigned with clearer before/after wrapping: the swap phase now has htmx:before:swap → htmx:before:settle → htmx:after:settle → htmx:after:swap. So htmx:after:swap in htmx 4 is the final event after everything (including settling) is done — making it the closest equivalent to the old htmx:afterSettle.

The upgrade-check tool's current mapping of htmx:afterSettle → htmx:after:swap is intentional and correct.

jorgecs10:
thanks for the explanation! I'll close this pull request then.

PR #4028, opened 2026-09-06
Fix stale swaps after request replacement

### LINKED ISSUE #4027: [4.0.0] hx-sync="replace" lets an older response overwrite the latest result
With `hx-sync="#result:replace"`, an older response can overwrite the latest result.

Three requests, A, B, and C, start in that order:

1. B cancels A.
2. A finishes its cleanup.
3. C starts, but does not cancel B.
4. C updates the page.
5. B responds later and overwrites C.

I expected C to cancel B and remain the final result.

## Reproduction

The files are `main.go` and `repro.html` below, plus `htmx.js` from the official htmx v4.0.0 release.

Run `go run main.go` and open <http://127.0.0.1:8765/repro.html>. The test starts automatically.

A and B respond after 1.5 seconds. C responds after 50 milliseconds.
The page starts C in the next task after A's final request event, so A's cleanup finishes first.
The final result is **B**, but it should be **C**.

<details>
<summary>main.go</summary>

```go
package main

import (
	"fmt"
	"log"
	"net/http"
	"time"
)

func main() {
	log.SetFlags(log.Ldate | log.Ltime | log.Lmicroseconds)
	http.HandleFunc("/reply", func(w http.ResponseWriter, r *http.Request) {
		started := time.Now()
		log.Printf("received %s %s", r.Method, r.RequestURI)
		name := r.URL.Query().Get("name")
		delay := 1500 * time.Millisecond
		if name == "C" {
			delay = 50 * time.Millisecond
		}
		select {
		case <-r.Context().Done():
			log.Printf("cancelled %s after %s", r.RequestURI, time.Since(started))
			return
		case <-time.After(delay):
		}
		w.Header().Set("Content-Type", "text/html")
		_, err := fmt.Fprint(w, name)
		log.Printf("response written %s after %s; error=%v", r.RequestURI, time.Since(started), err)
	})
	http.Handle("/", http.FileServer(http.Dir(".")))
	log.Print("Open http://127.0.0.1:8765/repro.html")
	log.Fatal(http.ListenAndServe("127.0.0.1:8765", nil))
}
```

</details>

<details>
<summary>repro.html</summary>

```html
<!doctype html>
<meta charset="utf-8">
<title>htmx replace queue race</title>
<script src="htmx.js"></script>
<h1>htmx replace queue race</h1>
<p>Expected final content: C. Bug: B overwrites C. Reload to repeat.</p>
<button id="A" hx-get="/reply?name=A" hx-sync="#result:replace" hx-target="#result">A</button>
<button id="B" hx-get="/reply?name=B" hx-sync="#result:replace" hx-target="#result">B</button>
<button id="C" hx-get="/reply?name=C" hx-sync="#result:replace" hx-target="#result">C</button>
<div id="result">Waiting</div>
<pre id="log"></pre>
<script>
const log = message => document.querySelector('#log').textContent += message + '\n';
const aborted = [];
const swaps = [];
const finished = new Set();
document.addEventListener('htmx:before:request', event => {
  const {ctx} = event.detail;
  const name = ctx.sourceElement.id;
  log('start ' + name);
  ctx.request.signal.addEventListener('abort', () => {
    aborted.push(name);
    log('abort ' + name);
  });
  if (name === 'A') setTimeout(() => document.querySelector('#B').click(), 100);
});
document.addEventListener('htmx:after:swap', () => {
  swaps.push(document.querySelector('#result').textContent.trim());
  log('render ' + swaps.at(-1));
});
document.addEventListener('htmx:finally:request', event => {
  const name = event.detail.ctx.sourceElement.id;
  log('finally ' + name);
  finished.add(name);
  // Next task: A's entire finally block, including queue cleanup, has returned.
  if (name === 'A') setTimeout(() => document.querySelector('#C').click(), 0);
  if (finished.has('B') && finished.has('C')) setTimeout(() => {
    const bug = !aborted.includes('B') && swaps.join(',') === 'C,B';
    log(bug ? 'BUG REPRODUCED: B was not aborted; B overwrote C' : 'UNEXPECTED RESULT');
    document.body.dataset.done = 'true';
  }, 0);
});
window.addEventListener('load', () => document.querySelector('#A').click());
</script>
```

</details>

## Cause

`RequestQueue.admit()` cancels A and makes B the active request.
Then A's `finally` block calls `requestQueue.continue()`, which clears the active request without checking which request called it.
C sees an empty queue slot, so it does not cancel B. B can then overwrite C's result.

This also affects [`four-dev`](https://github.com/bigskysoftware/htmx/tree/four-dev); I checked commit [`4a7a447`](https://github.com/bigskysoftware/htmx/commit/4a7a4478e7b7a7527bd4dabbc2590d4af576c090).

alessioalex:
Ignore cleanup from a request that is no longer active. Add regression tests for the replace and abort strategies.

Fixes #4027

## Description

Fix a bug where `hx-sync="replace"` lets an older response overwrite the latest result.

The cleanup of a cancelled request can clear the active request from the queue. This change checks which request owns the queue before clearing it, so later requests can still cancel it.

Fixes #4027

## Testing

Added regression tests for `replace` and `abort`. Both fail without the fix and pass with it.

Ran `npm run test` in Chromium: 1,738 passed, 5 skipped.

Checked the browser repro with the patched build. B is cancelled and only C renders.

## Checklist

  - [x] I have read the contribution guidelines
  - [x] I have targeted this PR against `four-dev` for the htmx 4 fix
  - [x] This is a bugfix
  - [x] I ran the test suite locally (`npm run test`) and verified that it succeeded

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -116,7 +116,8 @@ var htmx = (() => {
             return "queued"
         }
 
-        continue() {
+        continue(abortRequest) {
+            if (this.#current?.abort !== abortRequest) return
             this.#current = null    // free current slot
             this.#queue.shift()?.() // run next request
         }
@@ -545,10 +546,12 @@ var htmx = (() => {
             let requestQueue = this.__getRequestQueue(elt);
             this.__initializeAbortListener(elt);
 
+            let abortRequest = () => ctx.request?.abort?.()
+
             if (requestQueue.admit(
                 syncStrategy,
                 () => this.__issueRequest(ctx), // run when ready
-                () => ctx.request?.abort?.()    // abort if replaced
+                abortRequest                  // abort if replaced
             ) !== "run") return
 
             ctx.status = "issuing"
@@ -626,7 +629,7 @@ var htmx = (() => {
                     this.__enableElements(disableElements);
                 }
 
-                requestQueue.continue()
+                requestQueue.continue(abortRequest)
             }
         }
 
--- a/test/tests/unit/__getRequestQueue.js
+++ b/test/tests/unit/__getRequestQueue.js
@@ -48,8 +48,8 @@ describe('__getRequestQueue / RequestQueue unit tests', function() {
 
         assert.equal(result, 'queued')
 
-        queue.continue()
-        queue.continue()
+        queue.continue(noop)
+        queue.continue(noop)
 
         assert.deepEqual(started, [3])
     })
@@ -78,8 +78,8 @@ describe('__getRequestQueue / RequestQueue unit tests', function() {
         assert.equal(second, 'queued')
         assert.equal(third, 'dropped')
 
-        queue.continue()
-        queue.continue()
+        queue.continue(noop)
+        queue.continue(noop)
 
         assert.deepEqual(started, [2])
     })
@@ -93,7 +93,7 @@ describe('__getRequestQueue / RequestQueue unit tests', function() {
         queue.admit('queue all', () => started.push(2), noop)
         queue.admit('queue all', () => started.push(3), noop)
 
-        queue.continue()
+        queue.continue(noop)
 
         assert.deepEqual(started, [2])
     })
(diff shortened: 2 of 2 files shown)


MichaelWest22:
Thanks found another related bug while testing your change so raised a new PR #4039

PR #4035, opened 2026-09-07
fix: changed modifier permanently suppressed requests on form elements

### LINKED ISSUE #4033: [htmx 4] `changed` trigger modifier doesn't seem to work in 4.0.0-beta6
The `changed` modifier appears to block the trigger event from being detected.

Example: 

```django
<div>
  <form
    hx-post="{% url "note_update" %}"
    hx-trigger="keydown changed delay:250ms"
    hx-target="next .validation"
    hx-swap="innerHTML"
  >
    {...form fields}
  </form>
  <div class="validation"></div>
</div>
```

When the `changed` modifier is omitted, the POST request is correctly sent after a 250ms delay on keydown. If `changed` is included, then the POST never fires.

Expected behavior: when a keydown event is detected, only fire the POST if the field has changed

MichaelWest22:
## Description
# Fix `changed` modifier on forms (closes #4033)

## What was broken

`hx-trigger="keydown changed"` on a `<form>` silently suppresses every request. Doesn't matter what you type — nothing fires. Works fine on a bare `<input>`, so it's easy to miss until you try the natural pattern of triggering on the form itself.

## Why

The `changed` block checked `.value` on the element carrying `hx-trigger`. For a form that's `form.value`, which is always `undefined`:

```javascript
for (let fromElt of fromElts) {
    if (values.get(fromElt) !== fromElt.value) {  // undefined !== undefined — always false
        changed = true;
    }
}
if (!changed) return;  // always hits this
```

## The fix

Track `evt.target` instead — the actual input the user typed in. That's always correct regardless of whether the listener is on the input directly, a parent form, or a `from:` element. Seed on first sight so we don't fire before anything has actually changed.

```javascript
if (spec.changed) {
    let values = spec.values ??= new WeakMap();
    let target = evt.target;
    let value = target?.value;
    if (!values.has(target)) { values.set(target, value); return; }
    if (values.get(target) === value) return;
    values.set(target, value);
}
```

Three tests added: direct input suppression, the form bubbling case from #4033, and two inputs in one form tracked independently.

Corresponding issue:
#4033

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -789,9 +789,11 @@ var htmx = (() => {
                     if (spec.changed) {
                         let values = spec.values ??= new WeakMap();
                         let target = evt.target;
-                        let value = target?.value;
-                        if (values.has(target) && values.get(target) === value) return;
-                        values.set(target, value);
+                        if (!['checkbox', 'radio'].includes(target?.type)) {
+                            let value = target?.value;
+                            if (values.has(target) && values.get(target) === value) return;
+                            values.set(target, value);
+                        }
                     }
                     if (filter) {
                         if (this.__shouldCancel(evt)) evt.preventDefault();
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -790,8 +790,7 @@ var htmx = (() => {
                         let values = spec.values ??= new WeakMap();
                         let target = evt.target;
                         let value = target?.value;
-                        if (!values.has(target)) values.set(target, target?.value);
-                        if (values.get(target) === value) return;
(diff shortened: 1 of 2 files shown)


1cg:
i think changed doesn't work w/ checkboxes & radio buttons as well.  Just stumbled on that looking at:

https://github.com/bigskysoftware/htmx/pull/2896

MichaelWest22:
Fixed checkboxes and changed support now.  best to just ignore and skip them as changed makes no useful sense for them

PR #4038, opened 2026-09-07
fix: hx-swap-oob colon form supports spaces in selectors

### LINKED ISSUE #4029: hx-swap-oob cannot parse a selector containing a space
In htmx 4, `hx-swap-oob` drops the legacy colon form when the value contains a space:

```js
// __createOOBTask
if (oobValue !== "true" && oobValue && !oobValue.includes(" ")) {
```

The value then goes to `__parseSwapSpec`, which mis-parses it:

```
__parseSwapSpec("beforeend:#table tbody")       -> {style:"innerHTML", beforeend:"#table", tbody:true}
__parseSwapSpec("innerHTML:global #oob-target") -> {style:"innerHTML", innerHTML:"global", "#oob-target":true}
__parseSwapSpec("innerHTML:#legacy")            -> {style:"innerHTML", innerHTML:"#legacy"}
```

The swap style is wrong and no target survives, so the swap is a silent no-op. Verified end to end: `hx-swap-oob="innerHTML:#outer div"` leaves the target unchanged.

`beforeend:#table tbody` is the documented example in `www/src/content/reference/01-attributes/14-hx-swap-oob.md`, under "Using alternate swap strategies".

This also blocks the `global ` selector prefix, because that prefix needs a space. Two of the three shadow root cases from #2846 cannot be ported to htmx 4 until this is fixed. The third was ported in 2db5aaca.

MichaelWest22:
## Description
fix: hx-swap-oob colon form supports spaces in selectors

Fixes #4029.

The swapStyle:selector colon form in hx-swap-oob now correctly supports selectors containing spaces (e.g. innerHTML:.outer .inner, innerHTML:closest div, innerHTML:global #foo) by checking for a space only in the swap style part (before the colon) rather than the entire value. Values with a space before the colon fall through to HCON parsing as before, enabling the target:'...' form.

recent change was already fixed and just needed a test:
OOB swap targets are now resolved via __findAllExt instead of document.querySelectorAll, giving parity with partials — shadow root scoping, global, host, and all extended selectors work correctly. Ports the htmx 2 fix from #2846.

Tests added for: space-containing selectors, closest extended selector, and global crossing shadow root boundaries.

Corresponding issue:
#4029

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -1077,9 +1077,9 @@ var htmx = (() => {
 
         __createOOBTask(tasks, elt, oobValue, sourceElement) {
             let targetSelector = elt.id ? '#' + CSS.escape(elt.id) : null;
-            if (oobValue !== 'true' && oobValue && !oobValue.includes(' ')) {
-                [oobValue, targetSelector = targetSelector] = oobValue.split(/:(.*)/);
-            }
+            // if oobValue is in "swapStyle:selector" form (no space before colon), split it; otherwise treat as HCON
+            let [, style, sel] = oobValue.match(/^(\S+):(.+)/) ?? [];
+            if (style) [oobValue, targetSelector] = [style, sel || targetSelector];
             if (oobValue === 'true' || !oobValue) oobValue = 'outerHTML';
 
             let swapSpec = this.__parseSwapSpec(oobValue);
@@ -2192,7 +2192,7 @@ var htmx = (() => {
                     if (!this.__triggerExtensions(destination, 'htmx:before:morph:attr', { attrName: attr.name, newValue: attr.value })) continue;
                     destination.setAttribute(attr.name, attr.value);
                     if (attr.name === "value" && destination instanceof HTMLInputElement && destination.type !== "file" && document.activeElement !== destination) {
-                        destination.value = attr.value;
+                            destination.value = attr.value;
                     }
                 }
             }
--- a/test/tests/attributes/hx-swap-oob.js
+++ b/test/tests/attributes/hx-swap-oob.js
@@ -73,6 +73,41 @@ describe('hx-swap-oob', function() {
         playground().querySelectorAll('.target').forEach(el => el.innerText.should.equal('Updated'))
     })
 
+    it('swaps oob with colon form selector containing a space', async function () {
+        mockResponse('GET', '/test', '<div>Main</div><div hx-swap-oob="innerHTML:.outer .inner">New Content</div>')
+        createProcessedHTML('<div hx-get="/test">Click</div><div class="outer"><div class="inner">Original</div></div>');
+        find('[hx-get]').click()
+        await forRequest()
+        assertTextContentIs('.outer .inner', 'New Content')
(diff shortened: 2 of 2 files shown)


manwithacat:
Thanks for jumping on this — the `closest` / `global ` / descendant tests are the right #2846 hangover.

One angle: this isn’t really a “space in the selector” bug. `hx-swap-oob` is hosting two grammars:

1. **Colon form** (2020, 2.x `indexOf(':')`): `swapStyle:selector`. The selector is allowed to contain spaces — `beforeend:#table tbody`, `innerHTML:global #outside`.
2. **HCON** (4.x, same as `hx-swap`): `swapStyle key:value` — `innerHTML target:'#table tbody'`.

The `!includes(' ')` guard treats any space as HCON, so colon form never splits. `/^(\S+):(.+)/` is the 2.x first-colon split in regex form, plus “no space before the colon”. That fixes the documented cases.

It still treats **any** `token:rest` as colon form, so HCON-only `hx-swap-oob="target:#foo"` (a valid `hx-swap` value, no leading style) becomes style=`target`. Existing tests use `innerHTML target:#custom` (space before the colon) so they stay green.

I’d keep HCON rather than drop it — `target:` / `settle:` on oob are already tested. The disambiguation is: first colon inside the first token, unless that token is an HCON modifier.

```js
let colon = oobValue.indexOf(':');
let head = colon > 0 ? oobValue.slice(0, colon) : '';
let hconKey = ['target', 'settle', 'swap', 'transition', 'ignoreTitle',
    'scroll', 'show', 'focusScroll', 'swapEmpty', 'scrollTarget', 'showTarget'].includes(head);
if (colon > 0 && !head.includes(' ') && !hconKey) {
    [oobValue, targetSelector] = [head, oobValue.slice(colon + 1) || targetSelector];
}
```

Branch with that plus tests (including `target:#custom` and the documented `#table tbody`), against `four-dev`, no competing PR — take any of it into this one:

https://github.com/manwithacat/htmx/tree/fix/oob-colon-vs-hcon

Related family: OOB claiming “any `hx-swap` value” but using a different parser — we hit that on `textContent` in #3563.

MichaelWest22:
Yeah there is an argument that hx-swap-oob should not have the advanced target: HCON format I added.  And somewhere along the line i stuffed up the space colon code which is all this PR is fixing to put back the intended design.  hx-partial solves all the issues so much better which is why it is not the recommended way of doing all but the most basic hx-swap-oob=true styles.  in some way we should just drop all advanced hx-swap-oob parsing but as long as the code is tight and small I think it is better to keep it for now. 

But we don't have to worry about invalid use cases like the target/modifier only use case as this just is not currently designed to work and didn't work before in v2 so we are not losing anything if it doesn't work.  This is also why i'd be fine with just dropping HCON support from hx-swap-oob.

manwithacat:
HCON does offer some interesting options if you go the other direction and lean into it. Turning HCON into a legit grammar isn't a huge exercise.

PR #4039, opened 2026-09-07
fix: prevent stale continue() from clearing queue after hx-sync replace/abort

MichaelWest22:
## Description

When `hx-sync="replace"` (or `abort`) causes an in-flight request to be cancelled, its `finally` block still runs and calls `requestQueue.continue()`. This prematurely clears the slot that now belongs to the newer request, corrupting queue state and causing the newer request's own `continue()` to advance the queue a second time — potentially running queued requests out of order or skipping them entirely.

A second latent bug: firing `htmx:abort` called `abort()` which cancelled the fetch but never cleared `#current` or advanced the queue, permanently stalling any queued requests.

## Fix

`continue(abortFn)` now takes an identity token — the same function reference passed to `admit()`. It returns early if the caller is no longer the current owner:

```js
continue(abortFn) {
    if (this.#current?.abort !== abortFn) return
    this.#current = null
    this.#queue.shift()?.()
}
```

`abort()` delegates to `continue()` passing the current abort fn, which always matches, fixing the queue-stall bug with no duplicated logic:

```js
abort() {
    this.#current?.abort?.()
    this.continue(this.#current?.abort)
}
```

Corresponding issue:
#4027 

## Checklist

  source changes)
  approved via an issue

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -116,13 +116,15 @@ var htmx = (() => {
             return "queued"
         }
 
-        continue() {
+        continue(abortFn) {
+            if (this.#current?.abort !== abortFn) return
             this.#current = null    // free current slot
             this.#queue.shift()?.() // run next request
         }
 
         abort() {
             this.#current?.abort?.()
+            this.continue(this.#current?.abort)
         }
     }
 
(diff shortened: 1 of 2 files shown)


1cg:
someone beat you to it :)

Issue #4041, opened 2026-09-07
`build:ext` does not clean `dist/ext`, so renamed extensions 404 (hx-optimistic at 4.0.0)


manwithacat:
`build:ext` copies `src/ext/*.js` into `dist/ext` but never deletes the directory first. After a rename, the old basename stays in the published tarball (or an orphan `.map` does).

## Repro (4.0.0)

```
curl -I https://cdn.jsdelivr.net/npm/htmx.org@4.0.0/dist/ext/hx-optimistic.min.js
# 404

curl -I https://cdn.jsdelivr.net/npm/htmx.org@4.0.0/dist/ext/hx-pending.min.js
# 200
```

`hx-optimistic.js` was renamed to `hx-pending.js` in #3958. The 4.0.0 tree / npm pack still contains `dist/ext/hx-optimistic.min.js.map` with **no** matching `.js`. The same tree also still has a full leftover `hx-compat.*` set (`src/ext` only has `htmx-2-compat.js`).

Cause is mechanical: `package.json` `build:ext` is `mkdir -p dist/ext && for file in src/ext/*.js …` with no wipe. CI (`htmx 4 CI`) runs tests and `src/scripts/content/check.py`, but does not build dist or compare `src/ext` to `dist/ext`. `dev/RELEASE.md` also has no completeness check and no `--prerelease` for `-beta` / `-rc` tags (`v4.0.0-beta3` through `beta5` were published with `prerelease=false`; `beta6` was marked correctly).

## Ask

1. Wipe `dist/ext` at the start of `build:ext` so a release tarball is a pure function of `src/ext`.
2. Add a cheap completeness check (every `src/ext/*.js` has `.js` + `.min.js`; no dist basename that is not in src; every `.map` has a matching `.js`) and run it from CI after `bun run build`, and from the release checklist before `npm publish` / `gh release create`.
3. In `dev/RELEASE.md`: `npm publish --tag next` while 2.x remains npm `latest`; pass `--prerelease` on GitHub when the version contains `-`.

No runtime change. A changelog line on the next 4.x release that `hx-optimistic` is now `hx-pending` (old CDN path 404s) would help consumers. I am happy to send a `four-dev` PR for (1)–(3).

Not asking to move npm `latest` off 2.x — that policy is already documented.

manwithacat:
Follow-up PR: https://github.com/bigskysoftware/htmx/pull/4042

PR #4042, opened 2026-09-07
fix: wipe dist/ext at the start of build:ext

manwithacat:
## Description

`build:ext` copied `src/ext/*.js` into `dist/ext` but never wiped the directory. After a rename, the old basename stayed in the published tarball.

On 4.0.0 that shows up as:

- `https://cdn.jsdelivr.net/npm/htmx.org@4.0.0/dist/ext/hx-optimistic.min.js` → **404** (`hx-optimistic` was renamed to `hx-pending` in #3958; only an orphan `.map` shipped)
- leftover `dist/ext/hx-compat.*` with no `src/ext/hx-compat.js`

This PR is just `rm -rf dist/ext` at the start of `build:ext`, per @1cg. No runtime change. `dist/` is not in this PR; the next release build drops the orphans from the tarball.

Corresponding issue: #4041

Not asking to move npm `latest` off 2.x.

## Testing

`npm run build:ext` now starts from an empty `dist/ext`, so only current `src/ext/*.js` names are emitted.

## Checklist

* [x] I have targeted this PR against the correct branch (`four-dev` for htmx 4 source)
* [ ] I ran the full browser test suite locally (`npm run test`) — not applicable; this is a one-line build-script change

## Optional follow-ups (not in this PR)

Dropped the checker, CI job, and `dev/RELEASE.md` notes from the first version. Recording them here in case they are useful later:

1. **Release-only completeness check** — a tag / `workflow_dispatch` job that compares *committed* `dist/ext` to `src/ext` *without* rebuilding. A PR-time `bun run build` then check always passes once the wipe is in, so it would not gate the published tarball. A tag check would catch “forgot to rebuild `dist/` before publish.” Keep any such script out of `src/scripts/` (`build:scripts` copies `*.py` into the npm package).
2. **Same wipe for other generated trees** — `build:editors`, `build:skills`, and `build:scripts` are also copy-without-clean. A single `rm -rf dist && mkdir -p dist` at the start of `npm run build` would cover them, still with no extra tooling.
3. **`dev/RELEASE.md`** — `npm build` is a typo for `npm run build`. While 2.x is npm `latest`, 4.x needs `npm publish --tag next`. `gh release create --prerelease` when the version contains `-` keeps GitHub Latest on a real GA tag (`v4.0.0-beta3` through `beta5` were not marked prerelease).
4. **Changelog on the next 4.x** — `hx-optimistic` was renamed to `hx-pending` (#3958); the old CDN path 404s.


1cg:
Hey @manwithacat I like the cleaning out the dist on build but don't think we need the rest of the infrastructure.  Can you update that?

manwithacat:
Updated — this PR is now just `rm -rf dist/ext` at the start of `build:ext`. Dropped the checker, the CI job, and the `RELEASE.md` notes.

The next release build will drop the leftover `hx-optimistic` / `hx-compat` files from the tarball. I left a couple of optional follow-ups at the bottom of the description if they ever become useful.

Issue #4020, opened 2026-09-04
hx-download ignores the UTF-8 filename under certain circumstances (includes suggested fix)


francis-chiss:
My backend is, in my scenario, returning a file with the following `Content-Disposition` response header:

`attachment; filename="R_conciliation Inv priv_s.xlsx"; filename*=UTF-8''R%C3%A9conciliation%20Inv%20priv%C3%A9s.xlsx`

In other words, I expect the saved file to be called "Réconciliation Inv privés.xlsx" (french language).

With the current code, the parseFilename() function uses this regex:

`/filename\*?=['"]?(?:UTF-8'')?([^'";]+)/i`

That pattern matches the first filename... occurrence in the header, which in my scenario is filename="R_conciliation ..." and returns it immediately. It should prioritize filename*=UTF-8 when it exists.

Here is an LLM-generated implementation that worked for my scenario:

```
function parseFilename(headers, url) {
        let cd = headers.get('Content-Disposition');
        if (cd) {
            // 1) Prefer RFC 5987 / RFC 6266 UTF-8 form: filename*=UTF-8''...
            let utf8 = cd.match(/filename\*\s*=\s*(?:"([^"]+)"|([^;]+))/i);
            if (utf8) {
                let raw = (utf8[1] || utf8[2] || '').trim();
                // charset'lang'value
                let parts = raw.split("'");
                if (parts.length >= 3) {
                    let encoded = parts.slice(2).join("'"); // keep apostrophes in value if any
                    return decodeURIComponent(encoded);
                }
                // Some servers send filename*=<percent-encoded> without charset/lang
                return decodeURIComponent(raw);
            }

            // 2) Fallback to legacy filename=
            let ascii = cd.match(/filename\s*=\s*(?:"([^"]+)"|([^;]+))/i);
            if (ascii) return (ascii[1] || ascii[2]).trim();
        }

        return url.split('/').pop().split('?')[0] || 'download';
    }
```

I love HTMX. Thanks!

manwithacat:
PR: https://github.com/bigskysoftware/htmx/pull/4043 — demo: https://manwithacat.github.io/htmx/demos/hx-download-utf8-filename/

MichaelWest22:
Fix merged

PR #4043, opened 2026-09-07
fix: prefer RFC 5987 filename* in hx-download

manwithacat:
## Description

`parseFilename()` used `/filename\*?=/`, so the first `filename=` token won. When a server sent both an ASCII fallback and `filename*=UTF-8''…`, the UTF-8 name was ignored.

Corresponding issue: #4020

Demo: https://manwithacat.github.io/htmx/demos/hx-download-utf8-filename/

## Testing

Added tests for dual `filename=` + `filename*` (the reported header) and `filename*` alone.

Local `npm test` (Chromium, same as CI `htmx_tests`): 87 files, 1740 passed, 0 failed, 5 skipped.

## Checklist

* [x] I have targeted this PR against `four-dev`
* [x] This is a bugfix for documented RFC 6266 / `filename*` behaviour
* [x] Full browser suite — CI; `parseFilename` unit-checked with the reporter's header

### DIFF
--- a/src/ext/hx-download.js
+++ b/src/ext/hx-download.js
@@ -73,8 +73,14 @@
     function parseFilename(headers, url) {
         let cd = headers.get('Content-Disposition');
         if (cd) {
-            let match = cd.match(/filename\*?=['"]?(?:UTF-8'')?([^'";]+)/i);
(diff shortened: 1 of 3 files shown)


MichaelWest22:
Looks good thanks

Issue #4021, opened 2026-09-05
Distinguish request timeouts from intentional aborts in htmx:error


slovyagin:
## Problem

With the installed htmx.org 4.0.0 distribution, a request timeout and an hx-sync replacement both produce an AbortError in htmx:error, and ctx.request.signal.reason.name is AbortError in both cases. Applications cannot use that reason to show connection-failure feedback for timeouts while keeping intentional cancellation silent.

## Reproduction

Serve an endpoint /slow that holds its response open, load htmx, and use:

```html
<button hx-get="/slow" hx-config="timeout:100ms" hx-sync="this:replace">Send</button>
<script>
document.addEventListener("htmx:error", ({detail: {ctx, error}}) => {
  console.log(error.name, ctx.request.signal.reason?.name);
});
</script>
```

1. Click once and wait for the timeout: AbortError / AbortError.
2. Set hx-config="timeout:0", then click twice while the first request is pending: AbortError / AbortError for the replaced request.

I reproduced both cases in Chromium using the installed distribution without an application error handler.

## Expected behavior

Give timeout cancellation a distinguishable reason, such as new DOMException("Request timed out", "TimeoutError"), or expose a supported timeout discriminator on the error event. Intentional aborts should remain distinguishable.

The current #initTimeout implementation calls ctx.request.abort() without a reason. An application workaround can replace ctx.requestTimeout in htmx:before:request with a timer that supplies TimeoutError, but that depends on private timer initialization and cleanup rather than a supported error boundary.

manwithacat:
PR: https://github.com/bigskysoftware/htmx/pull/4045 — demo: https://manwithacat.github.io/htmx/demos/timeout-error-reason/

PR #4045, opened 2026-09-07
fix: abort timed-out requests with TimeoutError

manwithacat:
## Description

Request timeouts called `AbortController.abort()` with no reason, so a timeout and an `hx-sync` replace were both `AbortError`. Pass `TimeoutError` (`"Request timed out"`) as the abort reason so apps can tell them apart via `ctx.request.signal.reason`.

Corresponding issue: #4021

Demo: https://manwithacat.github.io/htmx/demos/timeout-error-reason/

## Testing

Existing timeout test now asserts `signal.reason.name === 'TimeoutError'`. Added a replace-abort case that is not `TimeoutError`. Documented the discriminator on `htmx:error`.

Local `npm test` (Chromium, same as CI `htmx_tests`): 87 files, 1739 passed, 0 failed, 5 skipped.

## Checklist

* [x] I have targeted this PR against `four-dev`
* [x] This is a bugfix
* [x] Full browser suite — CI; `__issueRequest` unit tests cover timeout vs replace

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -671,7 +671,9 @@ var htmx = (() => {
                 ? this.parseInterval(ctx.request.timeout)
                 : this.config.defaultTimeout;
             if (timeout) {
-                ctx.requestTimeout = setTimeout(() => ctx.request?.abort?.(), timeout);
+                ctx.requestTimeout = setTimeout(() => {
+                    ctx.request?.abort?.(new DOMException("Request timed out", "TimeoutError"));
+                }, timeout);
             }
         }
 
--- a/test/tests/unit/__issueRequest.js
+++ b/test/tests/unit/__issueRequest.js
@@ -337,6 +337,23 @@ describe('__issueRequest unit tests', function() {
         await htmx.__issueRequest(ctx)
         assert.isTrue(errorFired)
         assert.isTrue(ctx.request.signal.aborted)
+        assert.equal(ctx.request.signal.reason?.name, 'TimeoutError')
+    })
+
+    it('timeout abort reason is distinguishable from hx-sync replace', async function () {
+        let div = createProcessedHTML('<div hx-get="/test" hx-swap="none" hx-sync="this:replace"></div>')
+        let ctx = htmx.__createRequestContext(div, new Event('click'))
+        ctx.fetch = (url, opts) => new Promise((_, reject) => {
+            opts.signal.addEventListener('abort', () => {
+                reject(opts.signal.reason || new DOMException('The operation was aborted', 'AbortError'))
+            })
+        })
+        let p = htmx.__issueRequest(ctx)
+        await new Promise(r => setTimeout(r, 10))
+        ctx.request.abort()
+        await p
+        assert.isTrue(ctx.request.signal.aborted)
+        assert.notEqual(ctx.request.signal.reason?.name, 'TimeoutError')
     })
 
     it('htmx:abort event aborts in-flight request', async function () {
--- a/www/src/content/reference/03-events/11-htmx-error.md
+++ b/www/src/content/reference/03-events/11-htmx-error.md
@@ -28,4 +28,6 @@ htmx.on('htmx:error', (evt) => {
 });
 ```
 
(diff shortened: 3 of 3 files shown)


src/htmx.js
```diff
@@ -671,7 +671,9 @@ var htmx = (() => {
                 ? this.parseInterval(ctx.request.timeout)
                 : this.config.defaultTimeout;
             if (timeout) {
-                ctx.requestTimeout = setTimeout(() => ctx.request?.abort?.(), timeout);
+                ctx.requestTimeout = setTimeout(() => {
+                    ctx.request?.abort?.(new DOMException("Request timed out", "TimeoutError"));
```
salomvary, line 675: Would it make sense to name the `DOMException` something htmx specific so that it is not going to get confused with a "native" `TimeoutError`? https://developer.mozilla.org/en-US/docs/Web/API/DOMException#timeouterror

Something like `signal.reason.name === 'htmx.TimeoutError'`?

salomvary:
This might deserve to be documented as a breaking change when upgrading htmx 2 to 4...

src/htmx.js
manwithacat: Good question. I used the standard TimeoutError name on purpose so this matches AbortSignal.timeout() and the usual reason.name === 'TimeoutError' check.

The distinction between an htmx timeout and a “native” timeout is empty: they are the same kind of event. The request ran out of time. Colliding with the platform TimeoutError is the goal, not something to avoid.

A custom htmx.TimeoutError would only help if we needed to tell htmx’s timer apart from another timeout on the same signal, which isn’t the #4021 case.

salomvary: > A custom htmx.TimeoutError would only help if we needed to tell htmx’s timer apart from another timeout on the same signal, which isn’t the #4021 case.

I'm not sure that's true. My use case, which I think is the same problem as outlined in #4021 is the following: as a lazy solution to handling all sorts of network errors and unexpected server responses, we show a generic "something went wrong" error notification in the UI whenever an `htmx:error` event is triggered. 

And the problem we are seeing with htmx 4 is the very specific `hx-sync="this:replace"` case, where htmx itself aborts "obsolete" pending requests, in which case we don't want to show an error notification, because this is "business as usual". 

But I might as well be misreading what/how this fix achieves :)

manwithacat: With the patch:  
* an hx-sync="this:replace" cancellation still has signal.reason.name === "AbortError"
* an actual request timeout has signal.reason.name === "TimeoutError"

So in your generic htmx:error handler you can treat AbortError as the expected “obsolete request was replaced” case and ignore it, while still treating TimeoutError as a genuine failure.

The reason I used the standard TimeoutError name rather than htmx.TimeoutError is that there doesn’t seem to be any additional htmx-specific distinction to encode here. A timeout produced by htmx’s timer has the same semantics as AbortSignal.timeout(): the operation exceeded its allotted time.

I do think your separate point about this being a behavioural change from htmx 2 is worth documenting. If htmx:error now includes expected cancellations such as hx-sync replacement, applications with a generic error handler need to know that they should discriminate on the abort reason.

MichaelWest22: Yeah in the ideal world we would actually use AbortSignal.timeout and not use our own custom setTimeout.  But for this to work we need to wrap it in AbortSignal.any() so we can keep the manual abort as an option as well and these two features are too newly released to be used yet in htmx without expensive fallback code.

to keep the code change minimal we should just be returning a text error reason directly like:
```
ctx.requestTimeout = setTimeout(() => ctx.request?.abort?.('timeout'), timeout);
```
or maybe 'RequestTimeout'  
This makes it clear it is just a custom timeout being reported by htmx and not trying to fake a standard abort signal error.

but aborts via hx-sync or manual htmx:abort events are not errors but are thrown as errors to be caught by the abort controller we use.  so we should probably also do:

```
            } catch (error) {
                ctx.status = "error: " + error;
                if (error?.name !== 'AbortError') this.__trigger(elt, "htmx:error", {ctx, error})
            } finally {
```
to swallow this expected thrown error ideally

MichaelWest22:
added #4098 that may be better as it prevents the non errors from triggering htmx:error

PR #4062, opened 2026-09-10
Fix default prevention for delayed triggers

### LINKED ISSUE #4057: [v4] hx-trigger: when form's `submit` is `delay`ed, it's native form submission is not cancelled
Reproduction: https://plnkr.co/edit/UEhJr2jeNzP52Z1E?open=lib%2Fscript.js (open `index.html` in the tree and click preview)

```html 
   <form action="/submit-native"
          method="get"
          hx-get="/submit-htmx"
          hx-trigger="submit delay:2s"
          hx-target="#result"
          hx-swap="innerHTML">
        <input name="message" value="Hello" aria-label="Message">
        <button type="submit">Submit</button>
    </form>
```

Click on the button.

### Expected

a request to `/submit-htmx` 2 seconds after clicking

### Actual

Native request immediately

Image

## Root cause

"Pre-timing" check at 

https://github.com/bigskysoftware/htmx/blob/d1b14b5411c49af7182e19519f127a644e171aea/src/htmx.js#L800-L804

`if (this.__shouldCancel(evt)) evt.preventDefault();` fires when event filter is specified. `hx-trigger="submit[true] delay:2s"` works as expected 

```html 
   <form action="/submit-native"
          method="get"
          hx-get="/submit-htmx"
          hx-trigger="submit[true] delay:2s" <!-- HERE -->
          hx-target="#result"
          hx-swap="innerHTML">
        <input name="message" value="Hello" aria-label="Message">
        <button type="submit">Submit</button>
    </form>
```

Image

Renerick:
Fixes #4057

Call `.preventDefault()` synchronously after trigger target checks, before delay or throttle. This keeps delayed form submissions from submitting natively.

+ a regression test verifying that delayed form submission is prevented immediately and the htmx request is still delayed.

### DIFF
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -797,8 +797,8 @@ var htmx = (() => {
                         }
                         if (!changed) return;
                     }
+                    if (this.__shouldCancel(evt)) evt.preventDefault();
                     if (filter) {
-                        if (this.__shouldCancel(evt)) evt.preventDefault();
                         let evtArgs = {}; for (let k in evt) evtArgs[k] = evt[k];
                         if (!this.__executeJavaScript(elt, evtArgs, filter, true, false)) return;
                     }
--- a/src/htmx.js
+++ b/src/htmx.js
@@ -783,6 +783,7 @@ var htmx = (() => {
 
                 // Guarded: pre-timing checks that determine if event should proceed
                 spec.handler = (evt) => {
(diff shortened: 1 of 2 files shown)


MichaelWest22:
actually this check needs to move to the very first line of the function before any early returns so that it can prevent native actions even when they are out of from or target scopes.  we still want them to be htmx overrides in these situations as well

Renerick:
Thank you!

Issue #4078, opened 2026-09-16
[htmx4] SSE unhandled AbortError when closing on the configured event


raxelo:
With htmx 4.0.0 and its bundled SSE extension, receiving the configured close event closes the connection but also produces an unhandled promise rejection.

```html
<div hx-sse:connect="/events" hx-sse:close="done"></div>
```

The endpoint responds with `Content-Type: text/event-stream`, writes `event: done\ndata: Complete\n\n`, and leaves the response open.

Expected: receiving `done` closes the connection without an unhandled error.

Actual: the connection closes, followed by:
```text
Uncaught (in promise) AbortError: BodyStreamBuffer was aborted
```

Reproduced with htmx 4.0.0, but not with htmx 2.0.10 + htmx-ext-sse 2.2.4.

[Repro on StackBlitz](https://stackblitz.com/edit/stackblitz-starters-wgke724y?file=v4.html)

MichaelWest22:
Added a PR to fix

XChikuX:
I thought the `htmx-sse` extension was completely separate from the core htmx.

Is this not the case with htmx 4 anymore? Does it come bundled?

I have several [improvements](https://github.com/bigskysoftware/htmx-extensions/pull/188) I've been trying to get merged since the month of may.

PR #4081, opened 2026-09-16
prevent abort exception on sse close event

MichaelWest22:
## Description
When hx-sse:close receives its configured event, the cleanup code aborts the fetch controller while parseSSE is still suspended on reader.read(), causing the browser to reject that pending read with an AbortError that has no handler. The fix cancels the reader first so parseSSE exits cleanly before the abort fires.

Corresponding issue:
#4078

## Testing
Added tests

## Checklist

  source changes)
  approved via an issue


liambuchanan:
Thanks for this, @MichaelWest22!

Wondering if it'd make sense to remove `connection.reader?.cancel?.()` from `cleanup()`? I think the preceding `abort()` should stop the fetch, and cancelling the errored reader produces the unhandled rejection. Is there a case where that additional cancellation is needed?

Issue #4090, opened 2026-09-21
htmx4 indicator never disappears when sending back an HX-Location to refresh only a partial page


jeffreybray:
**Summary**
htmx4 indicator never disappears when sending back an HX-Location to refresh only a partial page.

**Cause**
Reviewing 2.0.10, it looks like the .keepIndicators flag is set only for HX-Refresh and HX-Redirect.

In 4.0.0, in the __handleHeadersAndMaybeReturnEarly(ctx) function, it includes both HX-Refresh and HX-Redirect as before, but adds HX-Location. In __issueRequest() upon return from __handleHeadersAndMaybeReturnEarly it sets ctx.keepIndicators = true that doesn't call the function to remove the indicators.
I believe HX-Refresh and HX-Redirect will do a full page load, thus re-initializing the indicator as the whole page is loaded. Keeping the indicator going until the page reloads makes perfect sense. However, since HX-Location only refreshes a partial page, there needs to be something that removes the indicator request class.

**Reproduction**
The indicator spinner is in the base part of the page that doesn't get refreshed. Most interactions refresh only the #main part of the page, which leaves the "global" indicator untouched. Here's an overview of the page:

```
<body
  hx-indicator:inherited="#spinner"
  hx-target:inherited="#main"
>
...
<div id="main">
  <form
    hx-patch="/article/{{.ID}}"
    hx-push-url="false"
    hx-headers:inherited='{"X-CSRF-Token": "123"}'
  >...<button type=submit...>...</form>
  ...
</div>
<svg id="spinner" class="htmx-indicator absolute top-1/2 left-1/2">...</svg>
</body>
```
On submitting the form, the indicator appears and the server returns an HX-Location response header with a 200 HTTP code to refresh just the main panel.

The browser tools show the 200 response from the form submission and the subsequent get request to the specified page, but even after the partial page finishes updating, the spinner keeps spinning and never disappears.

MichaelWest22:
Yeah I think there is a behavior change here sorry.  We have had issues in the past like in #3390 where for the default hx-location use case of replacing the body you would want indicators and disabled elements to remain in place till the second hx-location redirect request replaces the body so indicators don't flash away in unexpected ways.  But there are these other rare edge cases where you will have issues.  You can fix this right now by using the htmx:finally:request event to set ctx.keepIndicators=false so that it would work as you expect for the partial replacement use case.

[...]

PR #4091, opened 2026-09-21
Fix HX-Location indicators not clearing

Ai-Eastern:
## Description

Only keep request indicators for `HX-Refresh` and `HX-Redirect`, which trigger full-page navigation. `HX-Location` starts a partial request, so the original request indicator must be cleared normally.

Adds regression coverage for the `HX-Location` flow and guards the existing `HX-Refresh`/`HX-Redirect` behavior.

Fixes #4090

## Testing

- RED: `npm test -- test/tests/unit/__handleHxHeadersAndMaybeReturnEarly.js` — 14 passed, 1 expected failure (`HX-Location` left the indicator visible).
- PASS: `npm test -- test/tests/unit/__handleHxHeadersAndMaybeReturnEarly.js test/tests/attributes/hx-indicator.js` — 20 passed, 0 failed.
- PASS: `git diff --check`.
- PARTIAL: `npm test` — 1782 passed, 1 failed, 4 skipped. The unrelated `hx-ws` timing test `reconnects after every default close code` expected 5 connections but observed 4; isolated rerun was 119 passed, 1 failed with the same assertion.
- UNAVAILABLE: `npm run format` — the `four-dev` branch has no `format` script or ESLint/Prettier configuration.

No generated `dist` files are included.


Ai-Eastern:
@MichaelWest22 I've updated this PR based on your suggestion: the existing behavior is preserved by default, while `HX-Location` can now set `keepIndicators:false` for partial-page replacements. I also added regression tests and documentation. Would you mind taking a look when you have a chance?

MichaelWest22:
@Ai-Eastern thanks.  i've had another look and I think we can maybe do a better solution to just get hx-location to pass the indicators and disabled elements down to the next ajax request in a chain so it can tidy them up naturally.  This avoid having to know to set keepIndicators manually when required as i had originally suggested.  Opened #4097 with this version

Ai-Eastern:
Thanks for taking another look. Passing the existing indicator and disabled-element state through the HX-Location request chain is cleaner and avoids requiring callers to opt into cleanup manually. Closing this in favor of #4097.

